PR #4, opened 2020-03-16
typo Ed25519verify -> Ed25519Verify

shaih:


stechu:
Thanks for the fix @shaih!

PR #12, opened 2020-07-15
TEAL v2 support

### LINKED ISSUE #5: Invalid TEAL program when the main program is `Cond`
When the main program is `Cond`, a label is generated to jump to the end of the program. However, this is not allowed by TEAL.

For example, using the following `simple.py` script:

```python
from pyteal import *

t = Cond(
        [Int(1) == Int(1), Int(2) == Int(2)],
        [Int(3) == Int(3), Int(4) == Int(4)]
        )

print(t.teal())
```

We have:
```bash
$ python3 simple.py > simple.teal
$ goal clerk send --from-program simple.teal --to GCBRBIIMS7IVTHC3Y7ML6WAGGPK4L3LVSHDWXPTMD5YO2E7CFWNN5X2KOE --amount 10
Couldn't broadcast tx with algod: HTTP 400 Bad Request:  32 bnz target beyond end of program
```

where `simple.teal` is:
```
int 1
int 1
==
bnz l0
int 3
int 3
==
bnz l1
err
int 1
bnz l2
pop
l1:
int 4
int 4
==
l2:
int 1
bnz l3
pop
l0:
int 2
int 2
==
l3:
```

One solution would be to add a no-op at the end. There does not seem any single-op no-op in TEAL currently, but it can be simulated using a double op:
```
dup
pop
```

### LINKED ISSUE #10: Create tests for new AST

jasonpaulos:
This PR adds support for LogicSig v2 opcodes and features, including stateful TEAL applications. In addition to supporting new v2 opcodes, this PR also adds the following:
* `TxnType` enum
* `Pop` expression
* `Not` expression
* `BitwiseNot` expression
* `BitwiseAnd` expression
* `BitwiseOr` expression
* `BitwiseXor` expression
* `Neq` (not equal) expression
* `Assert` expression
* `ScratchSlot`, `ScratchLoad`, and `ScratchStore` objects/expressions for manipulating scratch space
* Testing `.type_of()` for all expressions
* More structured IR that consists of `List[TealComponent]`

Closes #5, closes #10.

### DIFF
--- a/pyteal/ast/addr.py
+++ b/pyteal/ast/addr.py
@@ -1,12 +1,13 @@
 from ..types import TealType, valid_address
+from ..ir import TealOp, Op
 from .leafexpr import LeafExpr
 
 class Addr(LeafExpr):
     """An expression that represents an Algorand address."""
-     
+
     def __init__(self, address: str) -> None:
         """Create a new Addr expression.
-        
+
         Args:
             address: A string containing a valid base32 Algorand address
         """
@@ -14,7 +15,7 @@ class Addr(LeafExpr):
         self.address = address
 
     def __teal__(self):
-        return [["addr", self.address]]
+        return [TealOp(Op.addr, self.address)]
 
     def __str__(self):
         return "(address: {})".format(self.address)
--- a/pyteal/ast/app.py
+++ b/pyteal/ast/app.py
@@ -1,0 +1,185 @@
+from enum import Enum
+
+from ..types import TealType, require_type
+from ..ir import TealOp, Op
+from .leafexpr import LeafExpr
+from .expr import Expr
+from .maybe import MaybeValue
+from .int import EnumInt
+from .global_ import Global
+
+class OnComplete:
+    """An enum of values that Txn.on_completion() may return."""
+    NoOp = EnumInt("NoOp")
+    OptIn = EnumInt("OptIn")
+    CloseOut = EnumInt("CloseOut")
+    ClearState = EnumInt("ClearState")
+    UpdateApplication = EnumInt("UpdateApplication")
+    DeleteApplication = EnumInt("DeleteApplication")
+
+class AppField(Enum):
+    optedIn = (Op.app_opted_in, TealType.uint64)
+    localGet = (Op.app_local_get, TealType.anytype)
+    localGetEx = (Op.app_local_get_ex, TealType.none)
+    globalGet = (Op.app_global_get, TealType.anytype)
+    globalGetEx = (Op.app_global_get_ex, TealType.none)
+    localPut = (Op.app_local_put, TealType.none)
+    globalPut = (Op.app_global_put, TealType.none)
+    localDel = (Op.app_local_del, TealType.none)
+    globalDel = (Op.app_global_del, TealType.none)
+
+    def __init__(self, op: Op, type: TealType) -> None:
+        self.op = op
+        self.ret_type = type
+    
+    def get_op(self) -> Op:
+        return self.op
+    
+    def type_of(self) -> TealType:
+        return self.ret_type
+
+class App(LeafExpr):
+    """An expression related to applications."""
+
+    def __init__(self, field:AppField, args) -> None:
+        self.field = field
+        self.args = args
+
+    def __str__(self):
+        ret_str = "({}".format(self.field.get_op())
+        for a in self.args:
+            ret_str += " " + a.__str__()
+        ret_str += ")"
+        return ret_str
+
+    def __teal__(self):
+        teal = []
+        for arg in self.args:
+            teal += arg.__teal__()
+        teal += [TealOp(self.field.get_op())]
+        return teal
+
+    def type_of(self):
+        return self.field.type_of()
+
+    @classmethod
+    def id(cls):
+        """Get the ID of the current running application.
+        
+        This is the same as Global.current_application_id().
+        """
+        return Global.current_application_id()
+
+    @classmethod
+    def optedIn(cls, account: Expr, app: Expr):
+        """Check if an account has opted in for an application.
+
+        Args:
+            account: An index into Txn.Accounts that corresponds to the account to check. Must
+            evaluate to uint64.
+            app: The ID of the application being checked. Must evaluate to uint64.
+        """
+        require_type(account.type_of(), TealType.uint64)
+        require_type(app.type_of(), TealType.uint64)
+        return cls(AppField.optedIn, [account, app])
+    
+    @classmethod
+    def localGet(cls, account: Expr, key: Expr):
+        """Read from an account's local state for the current application.
+
+        Args:
+            account: An index into Txn.Accounts that corresponds to the account to read from. Must
+            evaluate to uint64.
+            key: The key to read from the account's local state. Must evaluate to bytes.
+        """
+        require_type(account.type_of(), TealType.uint64)
+        require_type(key.type_of(), TealType.bytes)
+        return cls(AppField.localGet, [account, key])
+    
+    @classmethod
+    def localGetEx(cls, account: Expr, app: Expr, key: Expr):
+        """Read from an account's local state for an application.
+
+        Args:
+            account: An index into Txn.Accounts that corresponds to the account to read from. Must
+            evaluate to uint64.
+            app: The ID of the application being checked. Must evaluate to uint64.
+            key: The key to read from the account's local state. Must evaluate to bytes.
+        """
+        require_type(account.type_of(), TealType.uint64)
+        require_type(app.type_of(), TealType.uint64)
+        require_type(key.type_of(), TealType.bytes)
+        return MaybeValue(AppField.localGetEx.get_op(), TealType.anytype, args=[account, app, key])
+
+    @classmethod
+    def globalGet(cls, key: Expr):
+        """Read from the global state of the current application.
+
+        Args:
+            key: The key to read from the global application state. Must evaluate to bytes.
+        """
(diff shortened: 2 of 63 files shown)


examples/vote.py
```diff
@@ -0,0 +1,107 @@
...
+        Global.round() > App.globalGet(Bytes("VoteEnd")),
+        Return(Int(1)),
+        Seq([
+            get_vote,
+            If(get_vote.hasValue(), if_voted, Return(Int(1)))
+        ])
+    )
+
+    on_register = Seq([
+        Assert(And(Global.round() >= App.globalGet(Bytes("RegBegin")), Global.round() <= App.globalGet(Bytes("RegEnd")))),
```
algorandskiy, line 47: seems `OptIn` is missed here

pyteal/ast/addr.py
```diff
@@ -0,0 +1,24 @@
...
+        """Create a new Addr expression.
+
+        Args:
+            address: A string containing a valid base32 Algorand address
+        """
+        valid_address(address)
+        self.address = address
+
+    def __teal__(self):
+        return [TealOp(Op.addr, self.address)]
```
algorandskiy, line 18: not sure if it matter and just fyi, `addr` is assembler-level syntax sugar

jasonpaulos:
#13 contains the project restructure in preparation for LogicSig v2 features. This PR will rebase onto that when it's merged.

jasonpaulos:
This PR will be ready for review once #15 is merged and I rebase onto its changes.

pyteal/ast/app.py
```diff
@@ -0,0 +1,185 @@
...
+            account: An index into Txn.Accounts that corresponds to the account to check. Must
+            evaluate to uint64.
+            app: The ID of the application being checked. Must evaluate to uint64.
+        """
+        require_type(account.type_of(), TealType.uint64)
+        require_type(app.type_of(), TealType.uint64)
+        return cls(AppField.optedIn, [account, app])
+    
+    @classmethod
+    def localGet(cls, account: Expr, key: Expr):
```
derbear, line 87: I don't know if this is a good idea, but I wonder if we can take advantage of `__getitem__`, `__setitem__`, and `__delitem__` overloading here.

pyteal/ast/arg.py
```diff
@@ -1,13 +1,17 @@
 from ..types import TealType
+from ..ir import TealOp, Op
 from ..errors import TealInputError
 from .leafexpr import LeafExpr
 
 class Arg(LeafExpr):
-    """An expression to get an argument."""
-    
-    def __init__(self, index:int) -> None:
+    """An expression to get an argument when running in signature verification mode."""
```
derbear, line 7: I wonder if it makes sense to add mode checks statically.

pyteal/ast/scratch.py
```diff
@@ -0,0 +1,80 @@
+from ..types import TealType
+from .expr import Expr
+
+class ScratchSlot:
```
derbear, line 4: Is it possible to store type information along with the slot?

derbear: Perhaps this would add a restriction to the set of programs that you can write in PyTEAL in that a single slot number can't be used for data of two different types statically (even though the program would run fine dynamically due to e.g., branching).

pyteal/ast/arg.py
jasonpaulos: Yes, the `verifyOpsForMode` function in `compiler.py` will raise an error if an unsupported op is used. The docstring is purely to help direct users to the correct expression to use based on the mode they want.

pyteal/ast/scratch.py
jasonpaulos: Currently the only place slots are used is in `MaybeValue`, which stores the type information. The type is then passed into `ScratchSlot.load()` so that type checking can be done on the loaded value. I think this is acceptable for now, especially because I'm not sure how slot reuse will be implemented in the future.

pyteal/ast/app.py
jasonpaulos: I have considered using `__getitem__`, but I believe `__setitem__` and `__delitem__` are not able to return anything so we wouldn't be able to use them as a drop in replacement for these AST functions. They would be great to have in a higher level approach that generates an AST though.

One alternative interface I have thought about is `Global.store(k, v)`, `Global.get(k)`, `Txn.accounts[0].store(k, v)`, `Txn.accounts[0].get(k)`. Some open questions with this are:
* Should `Txn.accounts[0]` still represent the address when used in an AST? Or should we have `Txn.accounts[0].address()`?
* How can I manipulate the sender's local state?
* The indexes in this array approach are one less than the actual runtime indexes would need to be. So should we only allow compile-time ints to be used? Or should `Txn.accounts[Int(1)]` be allowed, and what would it resolve to?

pyteal/ast/addr.py
jasonpaulos: `addr` takes in a base32 Algorand address with checksum, so it's useful to have.

pyteal/ast/app.py
derbear: > __setitem__ and __delitem__ are not able to return anything so we wouldn't be able to use them as a drop in replacement for these AST functions

I don't think `app_*_put` and `app_*_del` place anything on the stack; were you thinking that there might be a value returned by those that is of interested to PyTEAL?

derbear: > Should `Txn.accounts[0]` still represent the address when used in an AST? Or should we have `Txn.accounts[0].address()`?

> How can I manipulate the sender's local state?

I didn't think this idea through completely, but could you alternatively have `Txn.accountStorage[0]`?

derbear: I do like the idea of "calling" something to get the local storage corresponding to an account--as you point out, we do a bit of translation (subtract 1) under the covers

jasonpaulos: > I don't think `app_*_put` and `app_*_del` place anything on the stack; were you thinking that there might be a value returned by those that is of interest to PyTEAL?

With the current way we build the AST, constructing an expression must return an AST for that expression, regardless of whether it actually places something on the stack. For example,
```python
program = Seq([
    Assert(App.global["status"] == Int(0)), # this is ok
    App.global["status"] = Int(1), # doesn't work because `__setitem__` can't return a value
    del App.global["status"] # doesn't work becasue `__delitem__` can't return a value
])
```
We would have to use contexts or another front end in order to successfully use `__setitem__` and friends. I am definitely open to that, just not in this release 😅

jasonpaulos: > I didn't think this idea through completely, but could you alternatively have `Txn.accountStorage[0]`?

I will look into this today.

jasonpaulos: I've combined the `Txn.accountStorage` and `App.global[key].set/delete` proposals into this: https://gist.github.com/jasonpaulos/28844a66ec0aae10c3629a0fdc2ae512. I've implemented part of the backend for it and I think it works pretty well.

derbear: Oh, interesting. I do think I like the current approach in this PR, but this seems like it's getting pretty subjective. @algorandskiy any thoughts here?

pyteal/util.py
```diff
@@ -28,3 +28,9 @@ def execute(args):
     stdout, stderr = process.communicate()
     
     return (stdout.decode("utf-8"), stderr.decode("utf-8"))
+
+def escapeStr(s: str) -> str:
+    s = s.encode("utf-8").decode("latin-1").encode("unicode-escape").decode("latin-1")
```
derbear, line 33: Why does this work? Could you document it?

pyteal/ast/gtxn.py
```diff
@@ -113,6 +137,113 @@ def group_index(cls, index):
...
+        return cls(index, TxnField.freeze_asset_frozen)
+    
+    class ArrayAccessor:
+
+        def __init__(self, txIndex: int, accessField: TxnField, lengthField: TxnField) -> None:
+            self.txIndex = txIndex
+            self.accessField = accessField
+            self.lengthField = lengthField
+        
+        def length(self):
```
derbear, line 228: Perhaps we could use `__len__` here? `https://docs.python.org/3/reference/datamodel.html#object.__len__`

pyteal/compiler.py
```diff
@@ -11,6 +24,24 @@ def compileTeal(ast: Expr) -> str:
...
-    lines = ["#pragma version 1"]
-    lines += [" ".join(i) for i in teal]
+    verifyOpsForMode(teal, mode)
+
+    slots = set()
+    for stmt in teal:
+        for slot in stmt.getSlots():
+            slots.add(slot)
+    
+    if len(slots) > NUM_SLOTS:
```
derbear, line 34: Should this be `>=`?

derbear:
Looks fine to me, other than the remaining comments.

pyteal/ast/gtxn.py
jasonpaulos: From that page:
> Should return the length of the object, an integer >= 0.

Unfortunately Python will raise an error if `__len__` does not return an integer, so this is the next best thing.

pyteal/util.py
jasonpaulos: Added documentation

derbear: Unrelated: I'm surprised that there's no standard library implementation for doing this...

pyteal/ast/app.py
jasonpaulos: We've decided to wait on any change here for this release.

PR #13, opened 2020-07-27
Internal restructure

jasonpaulos:
A restructure of internal components in preparation for new LogicSig v2 features. The main change is separating every expression into its own file and slightly improving Cond codegen.

~~Externally, the following things have changed:~~
* ~~`TealTypeMismatchError` has been removed in favor of the existing `TealTypeError`.~~
* ~~Templated arguments now have a new interface, which simplifies internal implementation:~~
  - ~~Int template: `Tmpl.Int("TMPL_NAME")`~~
  - ~~Bytes template: `Tmpl.Bytes("TMPL_NAME")`~~
  - ~~Address template: `Tmpl.Addr("TMPL_NAME")`~~


derbear:
The restructuring changes look good. Would it be possible to split those from the ones that affect the external interface? Thanks!

jasonpaulos:
@derbear this PR now has only the internal changes. I will create another with the external changes once this has been merged.

jasonpaulos:
#14 contains the external changes that were previously here.

PR #14, opened 2020-07-29
Interface restructure

jasonpaulos:
A restructure of some external features in preparation for new LogicSig v2 features. The main changes are:
* `TealTypeMismatchError` has been removed in favor of the existing `TealTypeError`.
* Templated arguments now have a new interface, which simplifies internal implementation:
  - Int template: `Tmpl.Int("TMPL_NAME")`
  - Bytes template: `Tmpl.Bytes("TMPL_NAME")`
  - Address template: `Tmpl.Addr("TMPL_NAME")`
* The compilation function has changed from `expression.teal()` to `compileTeal(expression)`.

### DIFF
--- a/pyteal/ast/tmpl.py
+++ b/pyteal/ast/tmpl.py
@@ -5,15 +5,47 @@ from .leafexpr import LeafExpr
 class Tmpl(LeafExpr):
     """Template expression for creating placeholder values."""
 
-    def __init__(self, tmpl_v: str) -> None:
-        valid_tmpl(tmpl_v)
-        self.name = tmpl_v
+    def __init__(self, op: str, type: TealType, name: str) -> None:
+        valid_tmpl(name)
+        self.op = op
+        self.type = type
+        self.name = name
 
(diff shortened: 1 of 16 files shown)


pyteal/ast/tmpl.py
```diff
@@ -5,15 +5,47 @@
...
+        """Create a new Bytes template.
+
+        Args:
+            placeholder: The name to use for this template variable. Must start with "TMPL_" and
+            only consist of uppercase alphanumeric characters and underscores.
+        """
+        return cls("byte", TealType.bytes, placeholder)
+
+    @classmethod
+    def Addr(cls, placeholder: str):
```
derbear, line 44: Is the difference between `Bytes` and `Addr` here purely cosmetic?

jasonpaulos: Not entirely. Since `addr` takes in a base32 Algorand address with checksum, you would have to discard the checksum to import it as a normal byte string.

PR #15, opened 2020-08-03
Restructure tests

jasonpaulos:
Move AST tests from ` tests/test_ops.py` into individual testing files in the `pyteal` directory. The first commit only moves tests, while the second commit restructures the individual tests and adds checking for generated intermediate code.


derbear:
A proposal for testing intermediate code (that might be good to add to a different PR):

It may be possible to automatically generate differential tests that catch if some detail of low-level code generation changes from one version of PyTEAL to the next. There could be a test generator which exhaustively runs through the opcodes (and maybe small combinations of them) and outputs the expected values of these compiled opcodes; if these expected values are checked into the git repo, then we can have unit tests that check that we match these values. If we change how code generation works from one version to the next, then these tests will catch that.

I think we do something similar in the TEAL assembler here:
https://github.com/algorand/go-algorand/blob/master/data/transactions/logic/assembler_test.go#L217-L247

pyteal/ast/binaryexpr_test.py
```diff
@@ -0,0 +1,86 @@
...
+    with pytest.raises(TealTypeError):
+        Int(2) / Txn.receiver()
+
+def test_mod():
+    Int(10) % Int(9)
+
+    with pytest.raises(TealTypeError):
+        Txn.receiver() % Int(2)
+
+def test_arithmic():
```
derbear, line 130: Nit: typo here

pyteal/ast/bytes.py
```diff
@@ -10,7 +10,8 @@ def __init__(self, base: str, byte_str: str) -> None:
 
         Args:
             base: The base type for this byte string. Must be one of base16, base32, or base64.
-            byte_string: The content of the byte string, encoding with the passed in base.
+            byte_string: The content of the byte string, encoding with the passed in base. If base
+            is base16, this may be an empty string, otherwise an empty string is not allowed.
```
derbear, line 14: Nit: comma splice

pyteal/types.py
```diff
@@ -39,15 +39,15 @@ def valid_base32(s:str):
...
+    if len(s) == 0 or pattern.fullmatch(s) is None:
         raise TealInputError("{} is not a valid RFC 4648 base 32 string".format(s))
 
 def valid_base64(s:str):
     """ check if s is a valid base64 encoding string
     """
     pattern = re.compile(r'^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$')
 
-    if pattern.fullmatch(s) is None:
+    if len(s) == 0 or pattern.fullmatch(s) is None:
```
derbear, line 50: I believe that the TEAL assembler accepts empty strings.

Also, to be pedantic about the error message, see here: https://stackoverflow.com/questions/15212127/is-an-empty-string-valid-base64-encoded-data-of-zero-bytes-length

tests/compile_test.py
```diff
@@ -51,6 +52,7 @@ def test_atomic_swap():
 &&
 ||
 &&"""
+    reset_label_count()
```
derbear, line 55: Out of curiosity, why are these required here?

pyteal/types.py
jasonpaulos: You are correct. It looks like `byte base64` is not allowed, but `byte base64()` does create an empty string. I will remove this restriction.

derbear: It might also make sense to alter `byte base32` and just make the TEAL assembler consistent in all encodings.

tests/compile_test.py
jasonpaulos: Those are required in general to reset the global label counter in `util.py` so that the labels match the expected code string. Although it looks like this example does not actually use labels, so that call can be removed from here if you think it's confusing.

jasonpaulos:
@derbear that testing proposal sounds like a good idea. I've created an issue for it: #17

tests/compile_test.py
derbear: Ah, okay--that makes sense.

I don't think it's necessary to remove them; rather, it seems somewhat redundant that it's necessary to place it before every test for the tests to be correct.

To a larger point, it's a little unfortunate that state for these labels is global and not already implicitly reset per invocation of `compileTeal`, which is what intuitively makes sense to me. Perhaps there's a way to move this global counter to somewhere local to `compileTeal`.

PR #19, opened 2020-08-14
Allow single branch if statements

jasonpaulos:
This change allows the creation of `If` expressions with only a true predicate, provided that predicate evaluates to `TealType.none`. This is useful in situations where a program should return early or manipulate state only if a condition is true. For example,
```python
from pyteal import *

program = Seq([
    If(App.optedIn(Int(0), App.id()),
        App.localPut(Int(0), Bytes("admin"), Int(1))
    ),
    Return(Int(1))
])

print(compileTeal(program, Mode.Application))
```
This compiles to:
```
#pragma version 2
int 0
global CurrentApplicationID
app_opted_in
bz l0
int 0
byte "admin"
int 1
app_local_put
l0:
int 1
return
```

### DIFF
--- a/pyteal/ast/if_.py
+++ b/pyteal/ast/if_.py
@@ -6,44 +6,57 @@ from .expr import Expr
 class If(Expr):
     """Simple two-way conditional expression."""
 
-    def __init__(self, cond: Expr, thenBranch: Expr, elseBranch: Expr) -> None:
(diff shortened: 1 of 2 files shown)


derbear:
Suggestion: Have an `Unless` expression which does the opposite.

PR #23, opened 2020-08-20
Add missing Txn field

jasonpaulos:
`Txn.first_valid_time` somehow got removed when updating to TEAL v2


jasonpaulos:
Happy to leave this out since I can't imagine a use case for it.

PR #25, opened 2020-09-30
Remove travis token from build status image

onetechnical:
This token is only necessary for private repositories and could potentially leak private repository (i.e. not this repository) information.


jasonpaulos (approved):
LGTM

PR #34, opened 2020-12-02
Fix base32 validation

Ursify:
Hi Jason, Algorand Team.

Please review below regex update, I hope is fine and can pass your tests. Also, let me know if I need to point my pull request to codegen-test instead of master. This is my first pull request to a crypto project.

Cheers,
Ursify

### DIFF
--- a/pyteal/ast/bytes_test.py
+++ b/pyteal/ast/bytes_test.py
@@ -2,14 +2,25 @@ import pytest
 
 from .. import *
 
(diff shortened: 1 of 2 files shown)


jasonpaulos:
Hi @Ursify, thanks for catching this! TEAL actually supports base32 with padding and without padding, so I've made a modification to support both schemes and added some tests.

Once you sign the CLA I can approve this PR.

PR #37, opened 2020-12-04
Base 32 Pattern Check

Ursify:
Hi Jason, Algorand Team.

I hope below explanation makes sense.

BASE 32 works in groups of 8 Characters.
Valid Final 8 Characters(with / without padding):
CC======
CC
CCCC====
CCCC
CCCCC===
CCCCC
CCCCCCC=
CCCCCCC
CCCCCCCC
Invalid Final 8 Characters(with / without padding):
C=======
C
CCCCCC==
CCCCCC

Invalid 8 characters group should throw an error, since it won't be possible to uncoded them back.

Cheers,
Ursify


jasonpaulos:
Hi @Ursify, thanks for catching this too. I made a small change so that the regex accepts empty strings too and I've updated tests and an example which had invalid base32. Do you see any problems with the updated regex?

Ursify:
> Hi @Ursify, thanks for catching this too. I made a small change so that the regex accepts empty strings too and I've updated tests and an example which had invalid base32. Do you see any problems with the updated regex?

Hi @jasonpaulos , thanks for reviewing my PR. Your update is working fine. I run some test and all looks good. Cheers,

jasonpaulos (approved):

PR #41, opened 2021-01-29
Add optimization to detect duplicates

jasonpaulos:
This PR allows the compiler to use the `dup` and `dup2` ops when appropriate. The primary benefit of this optimization is to decrease the generated program size.

Edit: I've identified one case where this optimization produces bad results (stateful write ops that depend on stateful reads, and the same should be true for scratch stores that depend on scratch loads). This needs to be addressed before this is ready to merge.


jasonpaulos:
Closing this PR for now since it's significantly out of date.

In general this is still a good idea, but not something I'm able to work on again in the near future.

PR #46, opened 2021-03-23
Fix exponential blowup with blocks with `If`

fabrice102:
Previously, the functions `TealBlock.addIncoming` and
`TealBlock.validate` took exponential time with blocks with `If`.

The issue is that `If` blocks have two children (`left` and `right`),
and both of them have the same `next` child block.

To see the issue, run `examples/signature/test_many_ifs.py`
before and after this PR.


jasonpaulos:
@fabrice102 thanks for noticing this! I moved the code from `test_many_ifs.py` into its own test in `tests/compile_test.py` since I think that's a better place for it.

PR #63, opened 2021-05-08
adding PushBytes and PushInt

barnjamin:
Should say adding Tmpl.PushBytes and Tmpl.PushInt

### DIFF
--- a/pyteal/ast/tmpl.py
+++ b/pyteal/ast/tmpl.py
@@ -58,4 +58,25 @@ class Tmpl(LeafExpr):
         """
         return cls(Op.addr, TealType.bytes, placeholder)
 
+
+    @classmethod
+    def PushInt(cls, placeholder: str):
+        """Create a new PushInt template.
+        
+        Args:
+            placeholder: The name to use for this template variable. Must start with `TMPL_` and
+                only consist of uppercase alphanumeric characters and underscores.
+        """
+        return cls(Op.pushint, TealType.uint64, placeholder)
+
+    @classmethod
+    def PushBytes(cls, placeholder: str):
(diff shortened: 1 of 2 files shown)


jasonpaulos:
Hi @barnjamin, thanks for contributing. Could you please explain a situation where this new feature would be beneficial, and why #57 (which can create pushint and pushbytes ops) does not already satisfy it?

barnjamin:
I'd like to be able to know the exact stop/start byte positions of certain variables in a program so that I can verify in another program, via hash preimage, that the submission is from this template.  The intc/bytec blocks complicate things because I don't know exact positions of the bytes.  I use pushbytes to prevent the variables that I swap out with template values from being included in the constant blocks. 

From the description alone, #57 appears to enable/disable constant block generation for the entire program? If that is so, I don't think I'd like to turn off constant block generation for the whole program, since there may be other constants that would make sense to be deduplicated.

https://github.com/barnjamin/rareaf/blob/main/src/contracts/listing.py#L18-L20 example use is here

barnjamin:
> From the description alone, #57 appears to enable/disable constant block generation for the entire program

I think this is wrong, does it copy the `TMPL_...` string directly into the const block?

barnjamin:
Just actually tried it and it does output what I'm looking for.  Closing this but docs for the Tmpl/assembleConstants stuff will be useful, you want a pr for those?

Thanks

jasonpaulos:
There is some documentation about assembleConstants here: https://pyteal.readthedocs.io/en/latest/api.html#pyteal.compileTeal

If you think more documentation would be beneficial, I'd be happy to review a PR for it.

Issue #65, opened 2021-06-08
Add basic TEAL 4 opcodes to PyTeal


jdtzmn:
## Summary
Add basic opcodes (as opposed to [complex ones](https://github.com/algorand/go-algorand-internal/issues/1284)) that were added in TEAL 4. The `divmodw` and `expw` opcodes are not included because they will require additional design work.

## Scope/Requirements
Implement and write tests for:
* ~`gload`/`gloads`~ This needs some design work. Ideally you'd be able to choose specific values to "export" from one contract and be able to "import" them from another one, all without having to choose specific slot IDs. That's one way to do this at least.
* `gaid`/`gaids`
* `ExtraProgramPages` transaction field
* Int-based math ops:
  * `shl`
  * `shr`
  * `sqrt`
  * `bitlen`
  * `exp`

## Urgency/Relative Priority
Required to support TEAL 4 features.

PR #67, opened 2021-06-11
Implement TEAL 4 basic ops

jasonpaulos:
This PR adds support for the following feature from TEAL v4:
* `extra_program_pages` transaction field
* `GeneratedID` expression
* `BitLen` expression
* `Sqrt` expression
* `ShiftLeft`, `ShiftRight`, and `Exp` expressions
  *  Also overloads the Python operators for these operations


jasonpaulos:
@jdtzmn does `GeneratedID(txnIndex)` make sense for the `gaid(s)` opcode?

jdtzmn:
Yes, I think that makes sense. @jannotti any thoughts?

jannotti:
I'm ambivalent - I don't like that it implies the g is for "generated", as we're going for "group", as with gload and gtxn. But I guess you won't need the g's because it is known from context in pyteal? How will gload appear? Does the g in gtxn show up in pyteal naming?

I do like 'generated' compared to 'created' for being when clearer that it's about the ID that was made in that transaction.

So... I think it's good if you feel it fits in well with other choices in pyteal. I don't know it all that well.

jdtzmn (approved):
Looks good! I like the overloads.

jasonpaulos:
@jannotti `gload` will appear in a separate PR after some design work -- I want to explore the idea of having `Export`/`Import` operations that would allow contracts to share values using `gload`.

And yes g does show up in the `Gtxn` variable, but that's currently the only place. I think I'll merge this with `GeneratedID`, but if we change our mind about that name we'll have some time to update it before PyTeal gets released.

jasonpaulos:
Closes #65.

PR #75, opened 2021-07-14
Add ops for Byteslice arithmetic

### LINKED ISSUE #73: Support byteslice math
# Summary
TEAL v4 adds support for doing math on byteslices, meaning up to 512-bit uint math is now possible. PyTeal should support this as well.

# Scope
Supporting this properly may take 2 steps:
1. Add basic support for byteslice math by directly exposing functions like `BytesAdd(a, b)` to do math on byteslices.
2. Create a more robust type system for PyTeal. This could mean new uint types (128-bit, 256-bit, etc.) that would compile to byteslices and use byteslice math if needed.

This issue is for just the first step.

algochoi:
This commit adds support for Byteslice arithmetic operations (e.g. `b+`, `b~`, etc.) that were introduced in Teal V4 (#73). 

* Adds byteslice arithmetic operations (see [Algorand Opcodes](https://developer.algorand.org/docs/reference/teal/opcodes/))
* Updates documentation for byteslice arithmetic operations
* Add a line to README.md for installing requirements.txt 

Closes #73

### DIFF
--- a/docs/arithmetic_expression.rst
+++ b/docs/arithmetic_expression.rst
@@ -45,6 +45,36 @@ The associativity and precedence of the overloaded Python arithmetic operators a
  * :code:`Int(1) + Int(2) + Int(3)` is equivalent to :code:`Add(Add(Int(1), Int(2)), Int(3))`
  * :code:`Int(1) + Int(2) * Int(3)` is equivalent to :code:`Add(Int(1), Mul(Int(2), Int(3)))` 
 
+Byteslice Arithmetic
+~~~~~~~~~~~~~~~~~~~~
+
+Byteslice arithemetic is available for Teal V4 and above. 
+In PyTeal, byteslice arithmetic expressions include :code:`TealType.Bytes` values as arguments (with the exception of :code:`BZero`).
+The table below summarizes the byteslize arithmetic operations in PyTeal.
+
+=================================== ================================================= ===========================
+Operator                            Semantics                                         Example
+=================================== ================================================= ===========================
+:any:`BLt(a, b) <BLt>`              `1` if `a` is less than `b`, `0` otherwise        :code:`BLt(Bytes("base16", "0xFF"), Bytes("base16", "0xFE"))`
+:any:`BGt(a, b) <BGt>`              `1` if `a` is greater than `b`, `0` otherwise     :code:`BGt(Bytes("base16", "0xFF"), Bytes("base16", "0xFE"))`
+:any:`BLe(a, b) <BLe>`              `1` if `a` is no greater than `b`, `0` otherwise  :code:`BLe(Bytes("base16", "0xFF"), Bytes("base16", "0xFE"))`
+:any:`BGe(a, b) <BGe>`              `1` if `a` is no less than `b`, `0` otherwise     :code:`BGe(Bytes("base16", "0xFF"), Bytes("base16", "0xFE"))`
+:any:`BAdd(a, b) <BAdd>`            `a + b`, error (panic) if overflow                :code:`BAdd(Bytes("base16", "0x10"), Bytes("base16", "0x11"))`
+:any:`BMinus(a, b) <BMinus>`        `a - b`, error if underflow                       :code:`BMinus(Bytes("base16", "0x10"), Bytes("base16", "0x11"))`
+:any:`BMul(a, b) <BMul>`            `a * b`, error if overflow                        :code:`BMul(Bytes("base16", "0x10"), Bytes("base16", "0x11"))`
+:any:`BDiv(a, b) <BDiv>`            `a / b`, error if divided by zero                 :code:`BDiv(Bytes("base16", "0x10"), Bytes("base16", "0x11"))`
+:any:`BMod(a, b) <BMod>`            `a % b`, modulo operation                         :code:`BMod(Bytes("base16", "0xFF"), Bytes("base16", "0xFE"))`
+:any:`BEq(a, b) <BEq>`              `1` if `a` equals `b`, `0` otherwise              :code:`BEq(Bytes("base16", "0xFF"), Bytes("base16", "0xFF"))`
+:any:`BNeq(a, b) <BNeq>`            `0` if `a` equals `b`, `1` otherwise              :code:`BNeq(Bytes("base16", "0xFF"), Bytes("base16", "0xFF"))`
+:any:`BAnd(a, b) <pyteal.BAnd>`     `1` if `a > 0 && b > 0`, `0` otherwise            :code:`BAnd(Bytes("base16", "0xFF"), Bytes("base16", "0xFE"))`
+:any:`BOr(a, b) <pyteal.BOr>`       `1` if `a > 0 || b > 0`, `0` otherwise            :code:`BOr(Bytes("base16", "0xFF"), Bytes("base16", "0xFE"))`
+:any:`BXOr(a, b) <pyteal.BXor>`     `a ^ b`, bit-wise xor operation                   :code:`BXor(Bytes("base16", "0xFF"), Bytes("base16", "0xFE"))`
+:any:`BNot(a) <pyteal.BNot>`        `~a`, invert all its bits                         :code:`BNot(Bytes("base16", "0xFF00"))`
+:any:`BZero(a) <pyteal.BZero>`      return bytearray of length a with all zero bytes  :code:`BZero(Int(8))`
+=================================== ================================================= ===========================
+
+Currently, byteslice arithmetic operations are not overloaded, and must be explicitly called.
+
 .. _bit_and_byte_manipulation:
 
 Bit and Byte Operations
--- a/docs/arithmetic_expression.rst
+++ b/docs/arithmetic_expression.rst
@@ -55,26 +55,26 @@ In PyTeal, byteslice arithmetic expressions include
 and must be 64 bytes or less.
 The table below summarizes the byteslize arithmetic operations in PyTeal.
 
-======================================= ============= ====================================================================== ==================
-Operator                                Return Type   Example                                                                Example Result
-======================================= ============= ====================================================================== ==================
-:any:`BytesLt(a, b) <BytesLt>`          :code:`Int`   :code:`BytesLt(Bytes("base16", "0xFF"), Bytes("base16", "0xFE"))`      :code:`0`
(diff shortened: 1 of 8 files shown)


pyteal/ast/binaryexpr.py
```diff
@@ -259,3 +259,185 @@ def GetByte(value: Expr, index: Expr) -> BinaryExpr:
         index: The index of the byte to extract. Must evaluate to an integer less than Len(value).
     """
     return BinaryExpr(Op.getbyte, (TealType.bytes, TealType.uint64), TealType.uint64, value, index)
+
+def BAdd(left: Expr, right: Expr) -> BinaryExpr:
```
algochoi, line 263: I just used `B<name of op>` for the byteslice math ops, but it might not be explicit enough -- Thoughts on changing this to `Byte<name-of-op>`, e.g. `ByteAdd`?

jasonpaulos: I think I prefer `Bytes<name-of-op>`

pyteal/ast/binaryexpr.py
```diff
@@ -259,3 +259,185 @@ def GetByte(value: Expr, index: Expr) -> BinaryExpr:
...
+    """Add two numbers as bytes.
+    
+    Produces left + right.
```
jasonpaulos, line 266: Can you elaborate on how the byte slices are interpreted as integers for each of these descriptions, like "Produces left + right, where left and right are interpreted as big-endian unsigned integers." It would be good to mention that the arguments must be 64 bytes or less too.

Also, for &, |, ^, it's worth mentioning that the arguments are "zero-left extended to the greater of their lengths" (taken from https://github.com/algorandfoundation/specs/blob/master/dev/TEAL_opcodes.md#b-11).

docs/conf.py
```diff
@@ -18,7 +18,7 @@
 # -- Project information -----------------------------------------------------
 
 project = 'PyTeal'
-copyright = '2020, Algorand'
```
jasonpaulos, line 21: Thanks for this!

docs/arithmetic_expression.rst
```diff
@@ -45,6 +45,36 @@ The associativity and precedence of the overloaded Python arithmetic operators a
...
 
+Byteslice Arithmetic
+~~~~~~~~~~~~~~~~~~~~
+
+Byteslice arithemetic is available for Teal V4 and above. 
+In PyTeal, byteslice arithmetic expressions include :code:`TealType.Bytes` values as arguments (with the exception of :code:`BZero`).
+The table below summarizes the byteslize arithmetic operations in PyTeal.
+
+=================================== ================================================= ===========================
+Operator                            Semantics                                         Example
```
jasonpaulos, line 56: I think the `Semantics` column is a little misleading since we don't provide overloads for these operations. Instead, it would probably be more useful to provide the produced values for the examples.

docs/arithmetic_expression.rst
```diff
@@ -45,6 +45,36 @@ The associativity and precedence of the overloaded Python arithmetic operators a
  * :code:`Int(1) + Int(2) + Int(3)` is equivalent to :code:`Add(Add(Int(1), Int(2)), Int(3))`
  * :code:`Int(1) + Int(2) * Int(3)` is equivalent to :code:`Add(Int(1), Mul(Int(2), Int(3)))` 
 
+Byteslice Arithmetic
+~~~~~~~~~~~~~~~~~~~~
+
+Byteslice arithemetic is available for Teal V4 and above.
```
jasonpaulos, line 51: Could you mention that this feature allows you to perform up to 512-bit arithmetic?

jasonpaulos (changes requested):
Implementation and tests look good! I left some suggestions for docs improvements.

pyteal/ast/binaryexpr.py
algochoi: Changed to `Bytes<name-of-op>`

docs/arithmetic_expression.rst
algochoi: Agreed, changing the docs columns to:
`Operator` | `Return Type` | `Example` | `Example Result`

docs/arithmetic_expression.rst
```diff
@@ -45,6 +45,39 @@ The associativity and precedence of the overloaded Python arithmetic operators a
...
+
+Byteslice arithemetic is available for Teal V4 and above. 
+Byteslice arithmetic operators allow up to 512-bit arithmetic.
+In PyTeal, byteslice arithmetic expressions include 
+:code:`TealType.Bytes` values as arguments (with the exception of :code:`BytesZero`)
+and must be 64 bytes or less.
+The table below summarizes the byteslize arithmetic operations in PyTeal.
+
+======================================= ============= ====================================================================== ==================
+Operator                                Return Type   Example                                                                Example Result
```
jasonpaulos, line 59: Could you instead use `TealType.uint64`/`TealType.bytes` for the return type? That's more accurate, since these ops don't actually return an instance of `Int` or `Bytes`.

docs/arithmetic_expression.rst
```diff
@@ -45,6 +45,39 @@ The associativity and precedence of the overloaded Python arithmetic operators a
...
+:any:`BytesGe(a, b) <BytesGe>`          :code:`Int`   :code:`BytesGe(Bytes("base16", "0xFF"), Bytes("base16", "0xFE"))`      :code:`1`
+:any:`BytesAdd(a, b) <BytesAdd>`        :code:`Bytes` :code:`BytesAdd(Bytes("base16", "0xFF"), Bytes("base16", "0xFE"))`     :code:`0x1FD`
+:any:`BytesMinus(a, b) <BytesMinus>`    :code:`Bytes` :code:`BytesMinus(Bytes("base16", "0xFF"), Bytes("base16", "0xFE"))`   :code:`0x1`
+:any:`BytesMul(a, b) <BytesMul>`        :code:`Bytes` :code:`BytesMul(Bytes("base16", "0xFF"), Bytes("base16", "0xFE"))`     :code:`0xFD02`
+:any:`BytesDiv(a, b) <BytesDiv>`        :code:`Bytes` :code:`BytesDiv(Bytes("base16", "0xFF"), Bytes("base16", "0x11"))`     :code:`0xF`
+:any:`BytesMod(a, b) <BytesMod>`        :code:`Bytes` :code:`BytesMod(Bytes("base16", "0xFF"), Bytes("base16", "0x12"))`     :code:`0x3`
+:any:`BytesEq(a, b) <BytesEq>`          :code:`Int`   :code:`BytesEq(Bytes("base16", "0xFF"), Bytes("base16", "0xFF"))`      :code:`1`
+:any:`BytesNeq(a, b) <BytesNeq>`        :code:`Int`   :code:`BytesNeq(Bytes("base16", "0xFF"), Bytes("base16", "0xFF"))`     :code:`0`
```
jasonpaulos, line 71: The examples and results are nice! One formatting issue though: hexadecimal byte strings need to be even lengths (since 2 characters define 1 byte), so could you prefix the odd-length example results with 0, e.g. `0x3` -> `0x03`.

Also, could you move the rows for `BytesEq` and `BytesNeq` right below `BytesGe`? I think it makes sense if all the comparison ops are next to each other.

jasonpaulos (changes requested):
Two minor comments about the docs table, otherwise this looks good!

docs/arithmetic_expression.rst
algochoi: Good catch!

docs/arithmetic_expression.rst
```diff
@@ -45,6 +45,39 @@ The associativity and precedence of the overloaded Python arithmetic operators a
...
+:any:`BytesDiv(a, b) <BytesDiv>`        :code:`TealType.Bytes`  :code:`BytesDiv(Bytes("base16", "0xFF"), Bytes("base16", "0x11"))`     :code:`0x0F`
+:any:`BytesMod(a, b) <BytesMod>`        :code:`TealType.Bytes`  :code:`BytesMod(Bytes("base16", "0xFF"), Bytes("base16", "0x12"))`     :code:`0x03`
+:any:`BytesAnd(a, b) <pyteal.BytesAnd>` :code:`TealType.Bytes`  :code:`BytesAnd(Bytes("base16", "0xBEEF"), Bytes("base16", "0x1337"))` :code:`0x1227`
+:any:`BytesOr(a, b) <pyteal.BytesOr>`   :code:`TealType.Bytes`  :code:`BytesOr(Bytes("base16", "0xBEEF"), Bytes("base16", "0x1337"))`  :code:`0xBFFF`
+:any:`BytesXor(a, b) <pyteal.BytesXor>` :code:`TealType.Bytes`  :code:`BytesXor(Bytes("base16", "0xBEEF"), Bytes("base16", "0x1337"))` :code:`0xADD8`
+:any:`BytesNot(a) <pyteal.BytesNot>`    :code:`TealType.Bytes`  :code:`BytesNot(Bytes("base16", "0xFF00"))`                            :code:`0x00FF`
+:any:`BytesZero(a) <pyteal.BytesZero>`  :code:`TealType.Bytes`  :code:`BytesZero(Int(8))`                                              :code:`0x00000000`
```
jasonpaulos, line 76: The example return value should actually be `0x0000000000000000` for this one

docs/arithmetic_expression.rst
```diff
@@ -45,6 +45,39 @@ The associativity and precedence of the overloaded Python arithmetic operators a
...
+======================================= ======================= ====================================================================== ==================
+
+Currently, byteslice arithmetic operations are not overloaded, and must be explicitly called.
+
 .. _bit_and_byte_manipulation:
 
 Bit and Byte Operations
```
jasonpaulos, line 83: I just noticed this section gets pulled under the one above.

Screen Shot 2021-07-15 at 2 22 49 PM

It should probably be at the same level instead, so could you change the underline type from `---` to `~~~` here?

jasonpaulos (changes requested):
This is the last change request I promise 😄

docs/arithmetic_expression.rst
algochoi: True! I will change this to `BytesZero(Int(4)` and keep the return as `0x00000000`, i.e. 4 Bytes of 0s

jasonpaulos (approved):
Looks good! Feel free to squash + merge this and delete your branch

PR #79, opened 2021-07-23
Expose gload op-codes and allow manual scratch slot id assignment

### LINKED ISSUE #74: Support importing/exporting values between contracts
# Summary
TEAL v4 introduces the `gload(s)` ops, which can load values from the scratch space of prior app call transactions in the same group. PyTeal should support this feature as well.

# Scope
edit: commenting out the initial idea in favor of a simpler approach inspired by @StylishTriangles's suggestion.

This issue requires the following changes:

1. Allow `ScratchSlot` to have an additional optional parameter `requestedSlotId`, which would be an integer from 0-255 which specifies the slot ID that compiler **must** assign to this slot. This would also require modifications to `compileTeal` to actually assign the correct ID to any `ScratchSlot`s with this parameter, and if multiple unique `ScratchSlot` request the same slot ID, compilation must fail.
2. Also add an optional `requestedSlotId` parameter to `ScratchVar` which would be passed directly to its underlying `ScratchSlot`.
3. Expose the `gload` op in a PyTeal expression (maybe as `ImportScratchValue(txnIndex: Expr, slotId: int)`)

~Ideally we can have a system where one contract can export many values (perhaps with helpful names that don't make it onto the chain) with something like `Export("bidAmount", computedBidAmount)`. Then, another contract can import this exported value with something like `computedBidAmount = ImportFrom(txnIndex, contract1.exports("bidAmount"))`.~

~This would require programs to be compiled in a specific order to make dependencies work, or all at once and the compiler can sort out dependencies.~

~Additionally, the compile method would need to return more information, perhaps a Python object that contains a dict of all exports from the contract and their assigned slots -- it's important to make this information accessible by the user so that they can properly debug, and it also lets other non-PyTeal contracts import their exports. This object then needs to be obtained by the importing contract in order to know which slots to actually load. Again, here it will be important to allow imports from non-PyTeal contracts (meaning essentially direct access to gload(s)).~

algochoi:
Exposes `gload` and `gloads` op codes and allows users to manually specify scratch slot IDs in Pyteal code.

Closes #74

### DIFF
--- a/pyteal/ast/gload_test.py
+++ b/pyteal/ast/gload_test.py
@@ -1,0 +1,60 @@
+from pyteal.compiler.compiler import compileTeal
+from pyteal.config import NUM_SLOTS
+import pytest
+
+from .. import *
+from .gload import ImportScratchValue
+# this is not necessary but mypy complains if it's not included
+from .. import MAX_GROUP_SIZE, CompileOptions
+
+teal3Options = CompileOptions(version=3)
+teal4Options = CompileOptions(version=4)
+
+def test_gload_teal_3():
+    with pytest.raises(TealInputError):
+        ImportScratchValue(0, 0).__teal__(teal3Options)
+
+    with pytest.raises(TealInputError):
+        ImportScratchValue(Int(0), 0).__teal__(teal3Options)
+
+def test_gload():
+    expr = ImportScratchValue(0, 0)
+    assert expr.type_of() == TealType.uint64
+
+    expected = TealSimpleBlock([
+        TealOp(expr, Op.gload, 0, 0)
+    ])
+    
+    actual, _ = expr.__teal__(teal4Options)
+
+    assert actual == expected
+
+def test_gload_dynamic():
+    arg = Int(0)
+    expr = ImportScratchValue(arg, 0)
+    assert expr.type_of() == TealType.uint64
+
+    expected = TealSimpleBlock([
+        TealOp(arg, Op.int, 0),
+        TealOp(expr, Op.gloads, 0)
+    ])
+    
+    actual, _ = expr.__teal__(teal4Options)
+    actual.addIncoming()
+    actual = TealBlock.NormalizeBlocks(actual)
+
+    assert actual == expected
+
+def test_gload_invalid():
+    with pytest.raises(TealInputError):
+        ImportScratchValue(-1, 0)
+
+    with pytest.raises(TealInputError):
+        ImportScratchValue(MAX_GROUP_SIZE, 0)
+
+    with pytest.raises(TealInputError):
+        ImportScratchValue(0, -1)
+
+    with pytest.raises(TealInputError):
+        ImportScratchValue(0, NUM_SLOTS)
+    
--- a/pyteal/ast/gload_test.py
+++ b/pyteal/ast/gload_test.py
@@ -1,11 +1,8 @@
-from pyteal.compiler.compiler import compileTeal
-from pyteal.config import NUM_SLOTS
 import pytest
 
 from .. import *
-from .gload import ImportScratchValue
 # this is not necessary but mypy complains if it's not included
-from .. import MAX_GROUP_SIZE, CompileOptions
+from .. import MAX_GROUP_SIZE, NUM_SLOTS, CompileOptions
 
 teal3Options = CompileOptions(version=3)
 teal4Options = CompileOptions(version=4)
@@ -19,7 +16,7 @@ def test_gload_teal_3():
 
 def test_gload():
     expr = ImportScratchValue(0, 0)
-    assert expr.type_of() == TealType.uint64
+    assert expr.type_of() == TealType.anytype
 
     expected = TealSimpleBlock([
         TealOp(expr, Op.gload, 0, 0)
@@ -32,7 +29,7 @@ def test_gload():
 def test_gload_dynamic():
     arg = Int(0)
     expr = ImportScratchValue(arg, 0)
-    assert expr.type_of() == TealType.uint64
+    assert expr.type_of() == TealType.anytype
 
     expected = TealSimpleBlock([
         TealOp(arg, Op.int, 0),
--- a/pyteal/ast/gload_test.py
+++ b/pyteal/ast/gload_test.py
@@ -9,17 +9,17 @@ teal4Options = CompileOptions(version=4)
 
 def test_gload_teal_3():
     with pytest.raises(TealInputError):
-        ImportScratchValue(0, 0).__teal__(teal3Options)
+        ImportScratchValue(0, 1).__teal__(teal3Options)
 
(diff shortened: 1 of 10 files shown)


pyteal/ast/scratchvar_test.py
```diff
@@ -66,3 +67,24 @@ def test_scratchvar_load():
...
+    otherVar = ScratchVar(TealType.uint64, 10)
+    prog  = Seq([
+                myvar.store(Int(5)),
+                otherVar.store(Int(0))
+            ])
+    with pytest.raises(TealInternalError):
+        compileTeal(prog, mode=Mode.Signature, version=4) 
+
+    # Users cannot assign a slot that has been automatically
+    # assigned by the compiler
```
algochoi, line 82: In the current implementation, users cannot assign a slot ID that has been automatically assigned by the compiler. This is a little tricky as users have to ultimately keep track of the compiler assigned slots (which are deterministically assigned in ascending order, starting from 0, for a valid program) if they also want user defined slot IDs. 

Maybe there should be a stricter requirement for users to specify all the slot IDs or let the compiler automatically assign all of them. Mixing the two could _potentially_ be a source for bugs?

Looking forward, optimizing for "dead" slots may also affect this?

jasonpaulos: I think I did not fully explain the intended behavior when mixing compiler-assigned slots and user-defined ones.

What should happen in this situation is that the compiler will first "reserve" all of the user-defined slot IDs (and throw an error if multiple slots request the same user-defined ID). Then, for the compiler-assigned slots, it can arbitrarily assign slots IDs from the range 0-255, excluding any reserved IDs that were already assigned to user-defined slots. (Ideally the compiler-assigned IDs will be backwards compatible with the current code to assign IDs, so that the generated TEAL code won't change after this update.)

Also, sorry for the confusing naming here, but the `id` field is not actually the TEAL slot ID that the `ScratchSlot` will have in the compiled program. It's just supposed to be a unique way to identify and deterministically order different `ScratchSlot` objects. If you look at the `compileTeal` function (specifically the loop `for index, slot in enumerate(sorted(slots, key=lambda slot: slot.id))`), you'll see that the `.id` property isn't actually what's assigned to the compiled source.

If you want to rename the `id` field to make it less confusing, I would support that.

jasonpaulos (changes requested):

pyteal/ast/scratchvar_test.py
algochoi: I think I understand now. To double check:

* The user is specifying a slot in the **compiled source**, so instead of assigning slots by `index, slot`, the compiler needs to assign the user-specified slots, then assign any extra ones in between (starting from 0 up to 255). 
* Basically there is a mapping between the `id` field and the actual slot number.

jasonpaulos: Yes, that's correct

pyteal/ast/scratch.py
```diff
@@ -9,12 +11,30 @@
...
+        Args:
+            requestedSlotId (optional): A scratch slot id that the compiler must store the value. 
+            This id may be a Python int in the range [0-256).       
+        """
+        if requestedSlotId is None:
+            self.id = ScratchSlot.nextSlotId
+            ScratchSlot.nextSlotId += 1
+            self.isReservedSlot = False
+        else:
+            # TODO: Is there a way to check whether the user hasn't alloted more than
```
jasonpaulos, line 31: To answer the TODO, no I don't think that's possible here. Since multiple programs can be defined in the same Python runtime, you can't figure out if the slots are exhausted until compile time.

pyteal/ast/scratchvar.py
```diff
@@ -16,15 +16,15 @@ class ScratchVar:
             ])
     """
 
-    def __init__(self, type: TealType = TealType.anytype):
+    def __init__(self, type: TealType = TealType.anytype, slotId: int = None):
         """Create a new ScratchVar with an optional type.
 
         Args:
             type (optional): The type that this variable can hold. An error will be thrown if an
```
jasonpaulos, line 23: The docstring is missing the `slotId` param

pyteal/ast/scratchvar_test.py
```diff
@@ -66,3 +67,27 @@ def test_scratchvar_load():
...
+
+def test_scratchvar_double_assign_invalid():
+    myvar    = ScratchVar(TealType.uint64, 10)
+    otherVar = ScratchVar(TealType.uint64, 10)
+    prog  = Seq([
+                myvar.store(Int(5)),
+                otherVar.store(Int(0))
+            ])
+    with pytest.raises(TealInternalError):
+        compileTeal(prog, mode=Mode.Signature, version=4)
```
jasonpaulos, line 93: These tests would actually fit in really well in the `pyteal/compiler/compiler_test.py` file, since they are testing parts of `compileTeal`. It would be great if you could compare the output of `compileTeal` to an expected source string too, which will help us guard against changing how slot assignments resolve in future versions.

For this file though, perhaps you could add version of `test_scratchvar_store` and `test_scratchvar_load` that use a user-defined slot id?

jasonpaulos (changes requested):

pyteal/ast/gload_test.py
```diff
@@ -0,0 +1,57 @@
...
+from .. import *
+# this is not necessary but mypy complains if it's not included
+from .. import MAX_GROUP_SIZE, NUM_SLOTS, CompileOptions
+
+teal3Options = CompileOptions(version=3)
+teal4Options = CompileOptions(version=4)
+
+def test_gload_teal_3():
+    with pytest.raises(TealInputError):
+        ImportScratchValue(0, 0).__teal__(teal3Options)
```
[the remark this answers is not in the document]
algochoi: Good point, will change.

pyteal/ast/scratchvar_test.py
algochoi: I think that makes more sense -- I moved these tests to the `compile_test.py` file and added some unit tests regarding load/store here like you suggested.

pyteal/compiler/compiler.py
```diff
@@ -1,9 +1,10 @@
-from typing import List
+from typing import List, Set
 
 from ..ast import Expr
 from ..ir import Mode, TealComponent, TealOp, TealBlock
 from ..errors import TealInputError, TealInternalError
 from ..config import NUM_SLOTS
+from ..ast import ScratchSlot
```
jasonpaulos, line 7: nit: there's already an import from `ast` above

pyteal/ast/scratch.py
jasonpaulos: Is this TODO meant to get committed?

jasonpaulos (approved):
Looks great! 1 question and 1 nit, but this can be merged after addressing those

PR #82, opened 2021-07-26
Update documentation to explain new if syntax

algochoi:
This change corrects a typo and adds an example for the new `If().Then().Else()` style syntax.


jasonpaulos (approved):
Looks good!

Issue #84, opened 2021-07-28
[TEAL 5] Add `AppParam` class


jasonpaulos:
## Summary
A new opcode, `app_params_get`, will be added to TEAL v5 (from this PR: https://github.com/algorand/go-algorand/pull/2301). PyTeal should expose this opcode with a new `AppParam` class, similar to the current `AssetParam` class.

## Scope
* Create and test a new `AppParam` class that exposes the `app_params_get` opcode.
* Since this is the first part of TEAL v5 that will be added to PyTeal, the following files will need to be modified as well:
  * `pyteal/compiler/compiler.py`: increase `MAX_TEAL_VERSION` to 5
  * `pyteal/ir/ops.py`: add new v5 opcodes to the `Op` enum (any opcodes from https://github.com/algorand/go-algorand/blob/master/data/transactions/logic/opcodes.go that have `Version` set to 5, e.g. `cover`, `uncover`, etc.). Note that new opcodes are still being added, so we will likely need to do this again once TEAL 5 is out of development.

PR #85, opened 2021-07-29
Add `AppParam` opcode for TEAL v5

ahangsu:
This commit introduces `AppParam` class for a new opcode in TEAL v5 (#84). Minor changes are also applied on `AssetParam` and compiler version.

### DIFF
--- a/pyteal/ast/app_test.py
+++ b/pyteal/ast/app_test.py
@@ -255,49 +255,209 @@ def test_global_del_invalid():
         App.globalDel(Int(2))
 
 
-def test_app_param_approval_program_valid(): ...
+def test_app_param_approval_program_valid():
+    _app = Int(1)
+    expr = AppParam.approvalProgram(_app)
+    assert expr.type_of() == TealType.none
+    assert expr.value().type_of() == TealType.bytes
+
+    expected = TealSimpleBlock([
+        TealOp(_app, Op.int, 1),
+        TealOp(expr, Op.app_params_get, "AppApprovalProgram"),
+        TealOp(None, Op.store, expr.slotOk),
+        TealOp(None, Op.store, expr.slotValue)
+    ])
+
+    actual, _ = expr.__teal__(options)
+    actual.addIncoming()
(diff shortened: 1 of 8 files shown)


pyteal/ast/app_test.py
```diff
@@ -162,94 +162,302 @@ def test_local_put():
...
     actual = TealBlock.NormalizeBlocks(actual)
-    
+
     assert actual == expected
 
 def test_global_del_invalid():
     with pytest.raises(TealTypeError):
         App.globalDel(Int(2))
+
+
```
algochoi, line 257: nit: consider using single line breaks here for consistency with the above tests

pyteal/ast/app_test.py
```diff
@@ -162,94 +162,302 @@ def test_local_put():
...
+def test_app_param_approval_program_valid():
+    _app = Int(1)
```
algochoi, line 259: nit: consider `arg` for consistency with the above tests

algochoi (commented):
Looks good! Just made some style suggestions. Unfortunately I'm not very familiar with the new `AppParam` so might want to have Jason approve it as well, but the ops and immediates seem to match the docs.

We might also want to update the docs at a later point for TEAL V5 ops.

pyteal/compiler/compiler_test.py
```diff
@@ -109,7 +109,7 @@ def test_compile_version_3():
 
 def test_compile_version_4():
```
jasonpaulos, line 110: Could you add a `test_compile_version_5` check similar to this one?

jasonpaulos (changes requested):
Great work! I just have one additional test request.

jasonpaulos:
Also, I just created a `teal5` branch. Can you change this PR so it merges onto that instead of `master`?

pyteal/compiler/compiler_test.py
ahangsu: Sure, 1 sec

jasonpaulos (approved):
Looks great! Once tests pass you can squash merge this

PR #90, opened 2021-08-04
Direct Reference for Application/Account/Asset

ahangsu:
This PR adds supports on direct reference for foreign app/account/asset, corresponding to the PR [#2264](https://github.com/algorand/go-algorand/pull/2264) in `go-algorand`.

Minor changes are applied to the following components:
- `AssetParam`: add support for passing asset id in `Txn.assets`.
- `AssetHolding`: add support for passing account address in `Txn.Accounts` or is `Txn.Sender`.
- `MinBalance, Balance` in `unaryexp.py`: add support for passing account address.
- `App`:
  - `optedIn, localGet, localGetEx, localPut, localDel`: add support for passing account address.
  - `globalGetEx`: add support for passing application id in `Txn.applications`.

Corresponding documents and test cases are modified.

### DIFF
--- a/docs/assets.rst
+++ b/docs/assets.rst
@@ -55,17 +55,17 @@ account holds. For example,
 
 .. code-block:: python
 
-    # get the balance of the sender for asset 31566704
+    # get the balance of the sender for asset `Txn.assets[0]`
     # if the account is not opted into that asset, returns 0
-    senderAssetBalance = AssetHolding.balance(Txn.sender(), Txn.assets[31566704])
+    senderAssetBalance = AssetHolding.balance(Txn.sender(), Txn.assets[0])
     program = Seq([
         senderAssetBalance,
         senderAssetBalance.value()
     ])
 
-    # get the balance of Txn.accounts[1] for asset 27165954
+    # get the balance of Txn.accounts[1] for asset `Txn.assets[1]`
     # if the account is not opted into that asset, exit with an error
-    account1AssetBalance = AssetHolding.balance(Txn.accounts[1], Txn.assets[27165954])
+    account1AssetBalance = AssetHolding.balance(Txn.accounts[1], Txn.assets[1])
     program = Seq([
         account1AssetBalance,
         Assert(account1AssetBalance.hasValue()),
@@ -80,17 +80,17 @@ A value of :code:`1` indicates frozen and :code:`0` indicates not frozen. For ex
 
 .. code-block:: python
 
-    # get the frozen status of the sender for asset 31566704
+    # get the frozen status of the sender for asset `Txn.assets[0]`
     # if the account is not opted into that asset, returns 0
-    senderAssetFrozen = AssetHolding.frozen(Txn.sender(), Txn.assets[31566704])
+    senderAssetFrozen = AssetHolding.frozen(Txn.sender(), Txn.assets[0])
     program = Seq([
         senderAssetFrozen,
         senderAssetFrozen.value()
     ])
 
-    # get the frozen status of Txn.accounts[1] for asset 27165954
+    # get the frozen status of Txn.accounts[1] for asset `Txn.assets[1]`
     # if the account is not opted into that asset, exit with an error
-    account1AssetFrozen = AssetHolding.frozen(Txn.accounts[1], Txn.assets[27165954])
+    account1AssetFrozen = AssetHolding.frozen(Txn.accounts[1], Txn.assets[1])
     program = Seq([
         account1AssetFrozen,
         Assert(account1AssetFrozen.hasValue()),
@@ -136,7 +136,7 @@ Here's an example that uses an asset parameter:
 
 .. code-block:: python
 
-    # get the total number of units for Txn.assets[0]
+    # get the total number of units for asset `Txn.assets[0]`
     # if the asset is invalid, exit with an error
     assetTotal = AssetParam.total(Txn.assets[0])
 
--- a/docs/assets.rst
+++ b/docs/assets.rst
@@ -152,6 +152,9 @@ Here's an example that uses an asset parameter:
     # get the total number of units for Txn.assets[0]
     # if the asset is invalid, exit with an error
     assetTotal = AssetParam.total(Int(0))
+    # another way to get the total number of units for Txn.assets[0]
+    assetTotalDirectRef = AssetParam.total(Txn.assets[0])
+
     program = Seq([
         assetTotal,
         Assert(assetTotal.hasValue()),
--- a/docs/assets.rst
+++ b/docs/assets.rst
@@ -15,7 +15,9 @@ For example,
 .. code-block:: python
 
     senderBalance = Balance(Int(0)) # get the balance of Txn.accounts[0] (the sender)
+    senderBalanceDirectRef = Balance(Txn.Sender()) # get the balance of the sender by passing the account address (bytes)
     account1Balance = Balance(Int(1)) # get the balance of Txn.accounts[1]
+    account1BalanceDirectRef = Balance(Txn.accounts[1]) # get the balance of Txn.accounts[1] by passing the account address (bytes)
 
 The :any:`MinBalance` expression can be used to find an account's `minimum balance <https://developer.algorand.org/docs/features/accounts/#minimum-balance>`_.
 This amount is also in microAlgos. For example,
@@ -23,7 +25,9 @@ This amount is also in microAlgos. For example,
 .. code-block:: python
 
     senderMinBalance = MinBalance(Int(0)) # get the minimum balance of Txn.accounts[0] (the sender)
+    senderMinBalanceDirectRef = MinBalance(Txn.Sender()) # get the minimum balance of the sender by passing the account address (bytes)
     account1MinBalance = MinBalance(Int(1)) # get the minimum balance of Txn.accounts[1]
+    account1MinBalance = MinBalance(Txn.accounts[1]) # get the minimum balance of Txn.accounts[1] by passing the account address (bytes)
 
 Additionally, :any:`Balance` and :any:`MinBalance` can be used together to calculate how many Algos
 an account can spend without closing. For example,
@@ -58,6 +62,8 @@ account holds. For example,
     # get the balance of Txn.accounts[0] (the sender) for asset 31566704
     # if the account is not opted into that asset, returns 0
     senderAssetBalance = AssetHolding.balance(Int(0), Int(31566704))
+    # get the balance of the sender by passing the sender's account address
+    senderAssetBalanceDirectRef = AssetHolding.balance(Txn.Sender(), Int(31566704))
     program = Seq([
         senderAssetBalance,
         senderAssetBalance.value()
@@ -66,6 +72,8 @@ account holds. For example,
     # get the balance of Txn.accounts[1] for asset 27165954
     # if the account is not opted into that asset, exit with an error
     account1AssetBalance = AssetHolding.balance(Int(1), Int(27165954))
+    # get the balance of Txn.accounts[1] by passing the account address
+    account1AssetBalanceDirectRef = AssetHolding.balance(Txn.Sender(), Int(27165954))
     program = Seq([
         account1AssetBalance,
         Assert(account1AssetBalance.hasValue()),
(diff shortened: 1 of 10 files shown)


pyteal/ast/app_test.py
```diff
@@ -92,38 +134,60 @@ def test_local_get_ex():
...
-        App.localGetEx(Txn.sender(), Int(0), Bytes("key"))
-    
+        App.localGetEx(Txn.sender(), Int(0), Int(0x123456))
+
     with pytest.raises(TealTypeError):
         App.localGetEx(Int(0), Bytes("app"), Bytes("key"))
 
     with pytest.raises(TealTypeError):
-        App.localGetEx(Int(0), Int(0), Int(1))
+        App.localGetEx(Txn.sender(), Int(0), Int(1))
```
jasonpaulos, line 175: It seems like this case is testing the same as `App.localGetEx(Txn.sender(), Int(0), Int(0x123456))`? If so, it can be removed.

pyteal/ast/asset.py
```diff
@@ -48,9 +50,9 @@ def total(cls, asset: Expr) -> MaybeValue:
             asset: An index into Txn.ForeignAssets that corresponds to the asset to check. Must
```
jasonpaulos, line 50: The docstring should also say that the `asset` arg can be the asset ID starting in v4.

pyteal/ast/asset.py
```diff
@@ -48,9 +50,9 @@ def total(cls, asset: Expr) -> MaybeValue:
...
                 evaluate to uint64.
         """
-        require_type(asset.type_of(), TealType.uint64)
+        require_type(asset.type_of(), TealType.anytype)
```
jasonpaulos, line 53: This should not change because the type remains a uint64, it just can now refer to the asset ID instead of its index in ForeignAssets.

docs/assets.rst
```diff
@@ -15,15 +15,19 @@ For example,
 .. code-block:: python
 
     senderBalance = Balance(Int(0)) # get the balance of Txn.accounts[0] (the sender)
+    senderBalanceDirectRef = Balance(Txn.sender()) # get the balance of the sender by passing the account address (bytes)
```
jasonpaulos, line 18: To simplify the docs, I think you should modify all the examples to only use direct references. Users can still view the generated API documentation for the functions to see that indexes are also acceptable.

docs/state.rst
```diff
@@ -91,8 +91,11 @@ and the third argument is the value to write. For example:
 .. code-block:: python
 
     App.localPut(Int(0), Bytes("role"), Bytes("admin")) # write a byte slice to Txn.accounts[0], the sender's account
+    App.localPut(Txn.sender(), Bytes("role"), Bytes("admin")) # write a byte slice to the sender's account
```
jasonpaulos, line 94: Same thing as the above comment, I think all examples should only use direct references. In this file you will need to modify some of the text in addition to the examples too.

jasonpaulos (changes requested):

pyteal/ast/app_test.py
ahangsu: Yea it is a duplicated case. Removed now.

docs/assets.rst
```diff
@@ -130,6 +142,7 @@ Expression                        Type                    Description
 :any:`AssetParam.reserve()`       :code:`TealType.bytes`  The address of the asset's reserve account.
 :any:`AssetParam.freeze()`        :code:`TealType.bytes`  The address of the asset's freeze account.
 :any:`AssetParam.clawback()`      :code:`TealType.bytes`  The address of the asset's clawback account.
+:any:`AssetParam.creator()`       :code:`TealType.bytes`  The address of the asset's creator account.
```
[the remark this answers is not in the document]
ahangsu: Removed from document.

pyteal/ast/asset.py
ahangsu: docstring updated.

pyteal/ast/asset.py
ahangsu: `AssetParam` type checking fixed.

docs/assets.rst
ahangsu: Document updated.

docs/state.rst
ahangsu: Document updated.

docs/state.rst
```diff
@@ -160,39 +158,39 @@ function.
 
 In order to use this function you need to pass in an integer that represents an application to
 read from. This integer corresponds to the index of an application in the
-:any:`Txn.applications <TxnObject.applications>` array.
-
-This means that in order to read or manipulate an external application's local state, that application
-must be present in the application call transaction's :code:`Txn.applications` array.
+:any:`Txn.applications <TxnObject.applications>` array, or since TEAL version 4, an byte string
```
jasonpaulos, line 161: The application argument can never be a byte array. It must be either an integer that's an index into the `Txn.applications` array, or the ID of an application. To avoid confusion, I think it's ok for this page to only mention that the ID can be the real ID of an application. Users can consult the API docs for the functions to learn about the other case.

(Since this is an ambiguous case and app IDs are all large, we actually interpret any integer less than the length of `Txn.applications` as an index, see here for implementation: https://github.com/algorand/go-algorand/blob/3d3feb5183da7b726f5207c18648779d8e9821b3/data/transactions/logic/eval.go#L2993)

jasonpaulos (changes requested):

docs/state.rst
```diff
@@ -85,14 +83,14 @@ Writing Local State
 ~~~~~~~~~~~~~~~~~~~
 
 To write to the local state of an account, use the :any:`App.localPut` function. The first argument
-is an integers corresponding to the account to write to, the second argument is the key to write to,
-and the third argument is the value to write. For example:
+is an integers or a byte string corresponding to the account to write to, the second argument is the
```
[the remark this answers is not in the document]
ahangsu: Fixed.

docs/assets.rst
```diff
@@ -55,17 +55,17 @@ account holds. For example,
 
 .. code-block:: python
 
-    # get the balance of Txn.accounts[0] (the sender) for asset 31566704
+    # get the balance of the sender for asset 31566704
     # if the account is not opted into that asset, returns 0
-    senderAssetBalance = AssetHolding.balance(Int(0), Int(31566704))
+    senderAssetBalance = AssetHolding.balance(Txn.sender(), Txn.assets[31566704])
```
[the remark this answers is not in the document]
ahangsu: My bad. I should have consulted the document about asset number upper-limit. The index is way too large here, which does not make sense. As far as I know, one account can have at most 1000 assets, right? (Just saying)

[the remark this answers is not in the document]
ahangsu: But anyways, `Txn.assets[i]` returns an uint64, which is of same type as the index.

docs/state.rst
```diff
@@ -72,10 +72,8 @@ Local State
 Local state consists of key-value pairs that are stored in a unique context for each account that
 has opted into your application. As a result, you will need to specify an account when manipulating
 local state. This is done by passing in an integer that corresponds to the index of the account in
-the :any:`Txn.accounts <TxnObject.accounts>` array.
-
-In order to read or manipulate an account's local state, that account must be present in the
-application call transaction's :code:`Txn.accounts` array.
+the :any:`Txn.accounts <TxnObject.accounts>` array (or since TEAL version 4, by passing in a byte
+string that appears in :code:`Txn.accounts` array).
```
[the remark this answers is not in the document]
ahangsu: Fixed.

docs/state.rst
ahangsu: Good catch. This indicates a mistake both of us overlooked in corresponding `globalPutEx` class, where I let type checking for application id / index to be `anytype`. It should be `uint64`.

I fixed the code section. For document, I think we can just omit the "too-much detailed" information about the argument we passed in.

docs/state.rst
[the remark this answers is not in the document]
ahangsu: Fixed other similar related sections in document to avoid confusion.

docs/state.rst
```diff
@@ -220,31 +216,31 @@ For example:
...
+    otherAppSenderRole = App.localGetEx(Txn.sender(), Int(31), Bytes("role"))
     program = Seq([
         otherAppSenderRole,
         If(otherAppSenderRole.hasValue(), otherAppSenderRole.value(), Bytes("none"))
     ])
 
     # get "role" from the local state of Txn.accounts[1] for the app with ID 31
     # if "role" has not been set, returns "none"
-    otherAppOtherAccountRole = App.localGetEx(Int(1), Int(31), Bytes("role"))
+    otherAppOtherAccountRole = App.localGetEx(Txn.sender(), Int(31), Bytes("role"))
```
jasonpaulos, line 243: Same here

docs/state.rst
```diff
@@ -159,40 +156,39 @@ To read a value from the global state of another application, use the :any:`App.
...
-This means that in order to read or manipulate an external application's local state, that application
-must be present in the application call transaction's :code:`Txn.applications` array.
-
 **Note:** The :code:`Txn.applications` array does not behave like a normal array. It's actually a
 :code:`1`-indexed array with a special value at index :code:`0`, the current application's ID.
 See :ref:`txn_special_case_arrays` for more details.
 
-Now that you have an integer that represents an application to read from, pass this as the first
-argument to :any:`App.globalGetEx`, and pass the key to read as the second argument. For example:
+Now that you have an integer or a byte string that represents an application to read from, pass
```
jasonpaulos, line 166:
```suggestion
Now that you have an integer that represents an application to read from, pass
```

jasonpaulos (changes requested):
Lots of small fixes, but overall this is looking good

docs/state.rst
```diff
@@ -85,14 +83,14 @@ Writing Local State
...
+third argument is the value to write. For example:
 
 .. code-block:: python
 
-    App.localPut(Int(0), Bytes("role"), Bytes("admin")) # write a byte slice to Txn.accounts[0], the sender's account
-    App.localPut(Int(0), Bytes("balance"), Int(10)) # write a uint64 to Txn.accounts[0], the sender's account
-    App.localPut(Int(1), Bytes("balance"), Int(10)) # write a uint64 to Txn.accounts[1]
+    App.localPut(Txn.sender(), Bytes("role"), Bytes("admin")) # write a byte slice to the sender's account
+    App.localPut(Txn.sender(), Bytes("balance"), Int(10)) # write a uint64 to the sender's account
+    App.localPut(Txn.account[1], Bytes("balance"), Int(10)) # write a uint64 to Txn.account[1]
```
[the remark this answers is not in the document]
ahangsu: Fixed

docs/state.rst
```diff
@@ -102,14 +100,13 @@ Reading Local State
...
-is an integer corresponding to the account to read from and the second argument is the key to read.
-For example:
+is the address of the account to read from, and the second argument is the key to read. For example:
 
 .. code-block:: python
 
-    App.localGet(Int(0), Bytes("role")) # read from Txn.accounts[0], the sender's account
-    App.localGet(Int(0), Bytes("balance")) # read from Txn.accounts[0], the the sender's account
-    App.localGet(Int(1), Bytes("balance")) # read from Txn.accounts[1]
+    App.localGet(Txn.sender(), Byte("role")) # read from the sender's account
```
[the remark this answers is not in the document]
ahangsu: Fixed

docs/state.rst
```diff
@@ -159,40 +156,39 @@ To read a value from the global state of another application, use the :any:`App.
...
+Now that you have an integer or a byte string that represents an application to read from, pass
+this as the first argument to :any:`App.globalGetEx`, and pass the key to read as the second
+argument. For example:
 
 .. code-block:: python
 
     # get "status" from the global context of Txn.applications[0] (the current app)
     # if "status" has not been set, returns "none"
-    myStatus = App.globalGetEx(Int(0), Bytes("status"))
+    myStatus = App.globalGetEx(Txn.application[0], Bytes("status"))
```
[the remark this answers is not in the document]
ahangsu: Fixed

docs/state.rst
```diff
@@ -220,31 +216,31 @@ For example:
...
+    myAppSenderRole = App.localGetEx(Txn.sender(), Int(0), Bytes("role"))
     program = Seq([
         myAppSenderRole,
         If(myAppSenderRole.hasValue(), myAppSenderRole.value(), Bytes("none"))
     ])
 
     # get "role" from the local state of Txn.accounts[1] for the current app
     # if "role" has not been set, returns "none"
-    myAppOtherAccountRole = App.localGetEx(Int(1), Int(0), Bytes("role"))
+    myAppOtherAccountRole = App.localGetEx(Txn.sender(), Int(0), Bytes("role"))
```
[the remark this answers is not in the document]
ahangsu: Fixed

docs/state.rst
ahangsu: Fixed

docs/state.rst
ahangsu: Fixed

jasonpaulos (approved):

PR #94, opened 2021-08-06
Add PR Template

sepandhaghighi:
Add pull requests template


jasonpaulos:
Apologies for the late reply, we've opted not to use this exact template, but we may add a different one in the future

PR #95, opened 2021-08-10
loop

shiqizng:
This PR adds While and For loops to Pyteal.


pyteal/ir/tealsimpleblock.py
```diff
@@ -24,16 +26,35 @@ def replaceOutgoing(self, oldBlock: TealBlock, newBlock: TealBlock) -> None:
...
             repr(self.ops),
             repr(self.nextBlock),
         )
+
+        self.visited.pop()
+        return s
     
     def __eq__(self, other: object) -> bool:
+        # check for loop
+        if self in self.blocks:
```
jasonpaulos, line 47: I don't love that these functions need to modify the object in order to work, but I can't think of a better solution right now -- I might make a follow up PR if I can think of something better.

In any case, `self.blocks` and `self.visited` do not need to be lists, since they only ever contain `self` or nothing. A cleaner solution would be to make a single `self.visited` boolean attribute, and the functions can set this to true/false to indicate whether `self` has been visited.

jasonpaulos (changes requested):
Nice work, nested loops and continue/break seem to be working as expected! My primary concern is about the sorting function, which I explain more below.

examples/signature/split.teal
```diff
@@ -13,24 +13,7 @@ global ZeroAddress
...
-int 0
-==
-&&
-txn FirstValid
-int 3000
->
-&&
-b l3
-l2:
+bz l2
```
algochoi, line 16: Are these auto-generated from the pyteal file?

algochoi (commented):
Looking good, just had some small questions and comments - I think we might want to run some formatter (e.g. `black`) before pushing to make spacing/line breaks more consistent and delete any commented out code blocks.

pyteal/compiler/compiler_test.py
```diff
@@ -117,27 +118,27 @@ def test_compile_version_4():
...
-        compileTeal(program, Mode.Application, version=2)
-
-    program = App.globalGetEx(Int(0), Bytes("key")).hasValue()
-    with pytest.raises(TealInternalError):
-        compileTeal(program, Mode.Application, version=2)
-    
-    program = ScratchVar().load()
-    with pytest.raises(TealInternalError):
-        compileTeal(program, Mode.Application, version=2)
+# def test_slot_load_before_store():
```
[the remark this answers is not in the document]
jasonpaulos: Actually this is OK to stay for now. I'm making a lot of changes to scratch slot validation to support subroutines, and in my PR I will uncomment this again.

pyteal/compiler/sort.py
```diff
@@ -12,22 +12,27 @@ def sortBlocks(start: TealBlock) -> List[TealBlock]:
...
-        for i, m in enumerate(n.getOutgoing()):
-            for i, block in enumerate(m.incoming):
-                if n is block:
-                    m.incoming.pop(i)
-                    break
-            if len(m.incoming) == 0:
-                if i == 0:
-                    S.insert(0, m)
-                else:
+        if type(n) == TealConditionalBlock:
```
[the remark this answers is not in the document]
ahangsu: ^ The previous version should be good here, even if `For`, `While` introduce cycles to the graph...

ahangsu (changes requested):
Looks good, I left some minor comments here and there for now.

examples/signature/split.teal
ahangsu: Yep, it seems Shiqi updated `split.teal` and `dutch_auction.teal` with new generated version.

pyteal/compiler/sort.py
```diff
@@ -12,22 +12,27 @@ def sortBlocks(start: TealBlock) -> List[TealBlock]:
...
+                    continue
+                if type(m) != TealConditionalBlock and len(m.incoming) != 0:
                     S.append(m)
-    
+                else:
+                    S.insert(0, m)
+
+        order.append(n)
+        visited.append(n)
+
```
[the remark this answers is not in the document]
shiqizng: you're correct. I simplified the algorithm as suggested.

pyteal/ast/for_.py
```diff
@@ -0,0 +1,103 @@
...
+        if self.step is None:
+            raise TealCompileError("For expression must have a end", self)
+        if self.doBlock is None:
+            raise TealCompileError("For expression must have a thenBranch", self)
+        
+        return "(For {} {} {} {})".format(self.start, self.cond, self.step, self.doBlock)
+
+    def type_of(self):
+        if str(self.doBlock) == str(Seq([Int(0)])):
+            raise TealCompileError("For expression must have a thenBranch", self)
```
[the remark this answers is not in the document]
shiqizng: yes, it should be checking for None.

pyteal/compiler/compiler.py
```diff
@@ -20,6 +20,9 @@ def __init__(
     ):
         self.mode = mode
         self.version = version
+        self.currentLoop: Optional[Expr] = None
```
algochoi, line 23: Should these be in `CompileOptions`? Or perhaps we can create another class for them, but I'm personally not sure if/how it will be affected when passing it around in the `__teal__` methods.

algochoi (commented):
The new formatting is much easier to read -- just one question about passing variables within `CompileOptions`

pyteal/ast/break_test.py
```diff
@@ -0,0 +1,42 @@
...
+from .. import *
+
+# this is not necessary but mypy complains if it's not included
+from .. import CompileOptions
+
+options = CompileOptions()
+
+
+def test_break_fail():
+
```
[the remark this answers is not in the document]
shiqizng: these are added by black.

pyteal/compiler/sort.py
```diff
@@ -13,22 +14,30 @@ def sortBlocks(start: TealBlock) -> List[TealBlock]:
...
-            if len(m.incoming) == 0:
-                if i == 0:
-                    S.insert(0, m)
-                else:
-                    S.append(m)
+        visited.add(id(n))
+
+    endIndex = -1
+    for i, block in enumerate(order):
+        if block == end:
```
jasonpaulos, line 33: Can you check if `block is end` instead? This is a slightly stronger check, since we want to make sure we find the end block exactly, not a different block which happens to have the same ops as the end block, which is what `==` checks.

pyteal/ir/tealsimpleblock.py
```diff
@@ -10,6 +10,8 @@ class TealSimpleBlock(TealBlock):
     def __init__(self, ops: List[TealOp]) -> None:
         super().__init__(ops)
         self.nextBlock: Optional[TealBlock] = None
+        self.blocks: List[TealBlock] = []
```
jasonpaulos, line 13: This variable isn't needed anymore

jasonpaulos (changes requested):
After these are addressed, this looks good to merge

pyteal/compiler/compiler.py
jasonpaulos: We decided to put these variables on `CompileOptions` because that class is essentially the de facto context manager during compilation. Maybe we can improve on this in the future, but I think it's ok for now

jasonpaulos (approved):
Looks good!

ahangsu (approved):
Should be good to go

PR #96, opened 2021-08-13
Align Seq with rest of the API

StylishTriangles:
It doesn't make sense for Seq to have a different API than all of the other operations.
This change allows to write
```
Seq(
    expr1,
    expr2
)
```
directly without needlessly wrapping the arguments in an array. This is especially convenient in repositories that use black to format source code. Before this patch the formatted code has too much indentation in a simple Seq:
```
Seq(
    [
        expr1,
        expr2,
    ]
)
```
This change doesn't break backwards compatibility.


jasonpaulos (approved):
This is a good change, thank you for the contribution

PR #99, opened 2021-08-19
Add subroutine support

### LINKED ISSUE #71: Add subroutines
# Summary
TEAL v4 introduces the `callsub` and `retsub` ops, which can be used to create subroutines in TEAL programs. PyTeal should have this functionality as well.

# Scope
Introduce a `@Subroutine` function decoration. This decoration would modify the function and it might look like:

```python
def Subroutine(targetFunc):
    subroutineID = # get a unique ID for this subroutine
    
    # save a reference to targetFunc somehow so the compiler can execute it later
    globalVarAllSubroutines[subroutineID] = targetFunc

    def subroutineCall(*args):
        # don't actually execute targetFunc, instead return a PyTeal expression indicating that you want to call it with args
        return CallSubroutine(subroutineID, args)

    return subroutineCall
```

Then during compilation, the compiler will collect all referenced subroutines, call the underlying functions to get their ASTs, and add them to the program's AST.

```python
@Subroutine
def isEven(i: Expr) -> Expr:
    return If(i == Int(0),
        Int(1),
        If(i == Int(1),
            Int(0),
            isEven(i - Int(2))
        )
    )

    # using proposed If.Then.Else.ElseIf syntax (see #70)
    return If(i == Int(0))
        .Then(Int(1))
        .ElseIf(i == Int(1))
        .Then(Int(0))
        .Else(isEven(i - Int(2)))
```

## Arguments

During compilation, arguments for subroutines can probably just be placed directly on the stack, like normal opcodes. The return value, if any, can also be placed on the stack for the caller to inspect.

## Caller vs Callee Scratch Slots

~~In order to prevent subroutines from overriding scratch space values set by their calling code (the caller), PyTeal needs a notion of which scratch slots are caller owned and callee owned. Whenever a caller wishes to call a subroutine, it needs to move all values it cares about out of the callee-owned slots (either into caller-owned ones or spill onto the stack). Similarly, whenever a subroutine wishes to store something in a caller-owned slot, it needs to remember the value of that slot and restore it before the subroutine finishes (again, probably by spilling onto the stack).~~

New idea: don't have an explicit notation of caller- vs callee-owned slots; instead, identify all the slots each subroutine (and the main program) uses, then divide up the available slots according to this mapping. Spilling some slots to the stack would probably be necessary if not enough slots are available for every subroutine to have unique ones (and in the case of recursion), but this scheme would minimize slot storing/loading.

## Recursion

Recursion with subroutines should be possible. To enable this, when the compiler executes the actual subroutine functions and collects their ASTs, it must be able to identify all the referenced subroutines in that subroutine _without_ them executing, just like a call into a first-level subroutine. Then the same logic can be used to find all referenced subroutines in that subroutine, etc., until no new subroutines are referenced.

## Inlining Subroutines

It would likely be a good optimization to identify subroutines that can be inlined directly into the caller code instead of being a legitimate subroutine. I expect many users will want to create subroutines for readability, not out of necessity.

jasonpaulos:
This draft PR contains an implementation of subroutines.

To do before this is ready for review:
- [ ] Rebase/merge with master
- [ ] Add more tests
- [ ] Fix and use `validateSlots` again (that's the source of the one failing test)
- [ ] Continue documenting new functions and classes

Closes #71


pyteal/ast/scratch.py
```diff
@@ -59,14 +59,12 @@ def load(self, type: TealType = TealType.anytype) -> "ScratchLoad":
         """
         return ScratchLoad(self, type)
 
+    def __repr__(self):
+        return "ScratchSlot({})".format(self.id)
+
     def __str__(self):
         return "slot#{}".format(self.id)
 
-    def __eq__(self, other):
```
jasonpaulos, line 65: I had to remove this so that `ScratchSlot(1) != ScratchSlot(1)`, which is required in order to detect duplicate requested slot IDs when I add all slots to a set.

pyteal/ir/labelref.py
```diff
@@ -0,0 +1,20 @@
+class LabelReference:
```
jasonpaulos, line 1: This class it basically a pass-by-reference string. It was added so that `flattenSubroutines` can easily prefix all the labels in a given subroutine in order to differentiate them.

PR #106, opened 2021-09-01
add Txn field Nonparticipation

shiqizng:


pyteal/ast/txn.py
```diff
@@ -85,6 +85,7 @@ class TxnField(Enum):
     local_num_uints = (54, "LocalNumUint", TealType.uint64, 3)
     local_num_byte_slices = (55, "LocalNumByteSlice", TealType.uint64, 3)
     extra_program_pages = (56, "ExtraProgramPages", TealType.uint64, 4)
+    nonparticipation = (57, "Nonparticipation", TealType.uint64, 2)
```
jasonpaulos, line 88: Unfortunately this isn't documented anywhere, but the last element of this tuple is the minimum TEAL version required to access that field, so it should be 5.

pyteal/ast/gtxn_test.py
```diff
@@ -675,6 +675,19 @@ def test_gtxn_id_dynamic():
     assert actual == expected
 
 
+def test_gtxn_nonparticipation():
```
jasonpaulos, line 678: Could you also add a test called `test_gtxn_nonparticipation_dynamic` which tests the same thing as this, but uses a PyTeal expression (e.g. `Int(0)`) as the txn index? There are a bunch of tests like that in this file too.

Additionally, once you change the minimum TEAL version to 5, can you add a check to verify compiling with an older version of TEAL raises an error? See `test_txn_extra_program_pages` for an example

jasonpaulos (commented):
I just created #107 to merge the `teal5` branch to master. Since that updates the compiler to work with v5, that will have to be merged before this.

jasonpaulos (approved):
Looks good!

PR #114, opened 2021-09-13
Use cover and uncover to simplify recursive subroutines

### LINKED ISSUE #110: Simplify recursive subroutine calling in TEAL 5
## Summary
TEAL 5 introduces the [`cover` and `uncover` opcodes](https://github.com/algorand/go-algorand/pull/2541) for stack manipulation, which can be used to greatly reduce the amount of opcodes and stack space needed to implement recursive subroutine calling.

## Scope
Instead of using `dig` and `pop` to help spill scratch slots during a potentially recursive subroutine call in `spillLocalSlotsDuringRecursion`, `cover` and `uncover` can do a better job. There are comments in this section of code describing the exact improvements that can be made: https://github.com/algorand/pyteal/blob/1157c57a66205bd7aa24390fe2d0254fbf609872/pyteal/compiler/subroutines.py#L119-L167

Currently 

## Priority
Not strictly needed, but this is a worthwhile improvement for subroutines.

jasonpaulos:
This PR uses the new `cover` and `uncover` TEAL 5 opcodes to make recursive subroutine calling more efficient.

For example, consider the following subroutine:

```python
@Subroutine(TealType.uint64)
def multiplyByAdding(a, b):
    return (
        If(a == Int(0))
        .Then(Return(Int(0)))
        .Else(Return(b + multiplyByAdding(a - Int(1), b)))
    )
```

With TEAL 4, the following code will be generated (comments added by me):

```
sub0: // multiplyByAdding
store 1
store 0
load 0
int 0
==
bnz sub0_l2
load 1
load 0
int 1
-
load 1 # At this point, the top of the stack contains the 2 arguments for the recursive call
load 0 # <-- Here, local scratch slots 0 and 1 are spilled to the stack so they don't get overwritten in the recursive call
load 1 # <-/
dig 3 # <-- These two lines use dig to pull the 2 arguments back to the top of the stack
dig 3 # <-/
callsub sub0 # the recursive subroutine call
store 0 # the return value is placed in slot 0 temporarily
store 1 # local slot 1 is restored
load 0 # the return value is placed back on the stack
swap # the return value and the previously spilled value for slot 0 are swapped
store 0 # local slot 0 is restored
swap # <-- Because we used dig to pull the arguments above the spilled slots, the original arguments are
pop  # <--| still there. These 4 lines get rid of them and make sure the return value ends at the top of the stack.
swap # <--|
pop  # <--/
+
retsub
sub0_l2:
int 0
retsub
```

With TEAL 5, now the following code will be generated:

```
sub0: // multiplyByAdding
store 1
store 0
load 0
int 0
==
bnz sub0_l2
load 1
load 0
int 1
-
load 1 # At this point, the top of the stack contains the 2 arguments for the recursive call
load 0 # <-- Here, local scratch slots 0 and 1 are spilled to the stack so they don't get overwritten in the recursive call
load 1 # <-/
uncover 3 # <-- Now, uncover is used to pull the 2 arguments to the top of the stack
uncover 3 # <-/
callsub sub0 # the recursive subroutine call
cover 2 # the return value is pushed below the 2 spilled slots on the stack
store 1 # <-- The spilled slots are restored
store 0 # <-/
+
retsub
sub0_l2:
int 0
retsub
```

The TEAL 5 version is much easier to understand and more efficient in terms of program size and stack usage.

Closes #110


pyteal/compiler/compiler_test.py
```diff
@@ -967,7 +967,56 @@ def isEven(i: Expr) -> Expr:
...
+sub0_l4:
+int 0
+b sub0_l3
+sub0_l5:
+int 1
+sub0_l6:
+retsub
+    """.strip()
+    actual = compileTeal(program, Mode.Application, version=5, assembleConstants=False)
+    assert actual == expected
```
ahangsu, line 1021:
```suggestion


def test_compile_subroutine_recursive_cover_5():
    @Subroutine(TealType.uint64)
    def multiplyByAdding(a, b):
        return (
            If(a == Int(0))
            .Then(Return(Int(0)))
            .Else(
                Return(b + multiplyByAdding(a - Int(1), b))
            )
        )
    program = Return(multiplyByAdding(Int(3), Int(5)))

    expected = """#pragma version 5
int 3
int 5
callsub sub0
return
sub0: // multiplyByAdding
store 1
store 0
load 0
int 0
==
bnz sub0_l2
load 1
load 0
int 1
-
load 1
load 0
load 1
uncover 3
uncover 3
callsub sub0
cover 2
store 1
store 0
+
retsub
sub0_l2:
int 0
retsub
    """.strip()

    actual = compileTeal(program, Mode.Application, version=5, assembleConstants=False)
    assert actual == expected

```

If you don't mind, I think it might be great to add the example illustrated in PR, since it tests the `(un)cover` op-code optimization in subroutine recursion.

pyteal/compiler/subroutines.py
```diff
@@ -108,63 +110,90 @@ def spillLocalSlotsDuringRecursion(
...
+                            )
                         )
-                    )
-                    # because we are stuck using dig instead of uncover in TEAL 4, we'll need to
-                    # pop all of the dug up arguments after the function returns
+                        # because we are stuck using dig instead of uncover in TEAL 4, we'll need to
+                        # pop all of the dug up arguments after the function returns
 
-                preserveReturnValue = False
+                coverAvailable = version >= Op.cover.min_version
```
ahangsu, line 157: I think this logic condition can be put before line 117 (some if condition deciding if `cover` available), what do you think?

pyteal/compiler/compiler_test.py
jasonpaulos: Good idea, I just added this test and a few more

pyteal/compiler/subroutines.py
jasonpaulos: Good idea

pyteal/compiler/subroutines.py
```diff
@@ -108,63 +112,89 @@ def spillLocalSlotsDuringRecursion(
...
-                    # is more efficient
-                    before.append(
-                        TealOp(
-                            None,
-                            Op.dig,
-                            len(slots) + subroutine.argumentCount() - 1,
+                    # local slots, if needed
+
+                    stackDistance = len(slots) + numArgs - 1
+
```
ahangsu, line 141: The `stackDistance` seems to be an invariant: slot length and number of args of subroutine are not modified in the loop. Can we put this ahead the for loop over `numArgs`?

pyteal/compiler/subroutines.py
```diff
@@ -108,63 +112,89 @@ def spillLocalSlotsDuringRecursion(
...
+                digArgs = True
+                coverSpilledSlots = False
+                uncoverArgs = False
+                if coverAvailable:
+                    digArgs = False
+                    if len(slots) < numArgs:
+                        coverSpilledSlots = True
+                    else:
+                        uncoverArgs = True
```
ahangsu, line 124: Could you please explain what this condition checks? I am not sure in what scenario `len(slots) < numArgs` is triggered, since the current check only cover condition `len(slots) == numArgs`.

Sorry in advance for the questions...

jasonpaulos: That's a good question. Before a recursive subroutine call, the arguments for the call are already on the stack. The compilers need to spill all local slots to the stack, somehow pull the arguments to the top of the stack, and then invoke the subroutine (and afterwards restore the slots).

There are 2 options to "somehow pull the arguments to the top of the stack":
1. Use `cover` to place each spilled slot below the arguments. E.g. with 3 args and 2 local slots:
```
arg 1
arg 2
arg 3
# the stack is now [arg1, arg2, arg3]

load 0
cover 3
load 1
cover 3
# the stack is now [slot0, slot1, arg1, arg2, arg3]
```
(I just realized the argument should be `numArgs`, not `len(slots) + numArgs - 1` 😅 -- thank you for questioning the lack of tests!)

2. Use `uncover` to pull up each argument above the spilled slots. E.g. with 3 args and 2 local slots:
```
arg 1
arg 2
arg 3
# the stack is now [arg1, arg2, arg3]

load 0
load 1
# the stack is now [arg1, arg2, arg3, slot0, slot1]

uncover 4
uncover 4
uncover 4
# the stack is now [slot0, slot1, arg1, arg2, arg3]
```

They are both equivalent, but depending on the number of arguments vs spilled slots, one might use more opcodes than the other -- in the examples using `cover` is more efficient.

I will add more tests to make sure all cases are evaluated.

ahangsu: All good all good, I believe once this is resolved, it should be good to merge. Thanks for adding more tests!

pyteal/compiler/subroutines.py
jasonpaulos: I'd rather leave that variable there, since it's the only place it's used

pyteal/compiler/subroutines.py
jasonpaulos: More tests added, but the implementation was actually correct. I thought it used the wrong value, but it uses `numArgs`

ahangsu: I see, the point here is to pick the method using least operation. Very neat.

ahangsu (approved):
LGTM. Should be good to go.

PR #115, opened 2021-09-15
Support inner transactions

### LINKED ISSUE #109: Dynamic Array Reference
## Problem

I'd like to be able to reference the elements of the Txn.application_args array dynamically using the looping construct with ScratchVars.  Currently this is not supported but I see support is available in the Gtxn array.

example of what I'd like to do:
```python
def set_properties(acct, args, start):
    idx = ScratchVar()

    init = idx.store(Int(start))
    cond = idx.load()<args.length()
    step = idx.store(idx.load() + Int(2))

    return Seq([
        For(init, cond, step).Do(
            App.localPut(acct, args[idx.load()], args[idx.load() + Int(1)])
        ),
        Int(1)
    ])

```
This may be supported in some way but I can't seem to find it.

## Solution

Accept the Union[int,Expr] in the Txn array items `__getitem__` call

### LINKED ISSUE #112: Support for creating inner transactions
## Summary
algorand/go-algorand#2661 introduced application actions, aka inner transactions which application can create. This issue is to expose that functionality in PyTeal as well.

## Scope

Due to the simplicity of the `tx_begin`, `tx_field`, and `tx_submit` opcodes, the design is also pretty simple.

Four new functions should be created:

1. `InnerTxnBuilder.Begin()`: This will begin an inner transaction. It will evaluate to the `tx_begin` opcode.

2. `InnerTxnBuilder.SetField(field: TxnField, value: Expr)`: This will set a single field on an inner transaction. This must be called after `InnerTxnBuilder.Begin()` to be a valid operation. The `field` argument must be from the  existing [`TxnField` enum](https://github.com/algorand/pyteal/blob/7186838c8f44899b71e8ac61acc0ad231c758266/pyteal/ast/txn.py#L30). The `value` argument must be a PyTeal expression which evaluates to a type that is compatible with the indicated `field` type. The compiler will throw an error if the types do not match, for instance if a byteslice is set to the `TxnField.amount` field.

3. `InnerTxnBuilder.SetFields(values: Dict[TxnField, Expr])`: This function is similar to `InnerTxnBuilder.SetField`, except it can accept multiple fields and values for convenience. The `values` argument must be a dictionary from `TxnField` to PyTeal expressions. Note that starting in Python 3.6 (which is coincidentally the lowest version PyTeal supports), dictionaries have a defined order, so this will not introduce any nondeterminism.

4. `InnerTxnBuilder.Submit()`: This will submit an inner transaction. It will evaluate to the `tx_submit` opcode.

PyTeal will not do any validation of inner transaction fields beyond the simple typechecking described above. This means that when the AVM adds support for more inner transaction features (e.g. the `acfg` txn type), PyTeal does not need to be updated.

### Example

A PyTeal-ized version of https://github.com/algorand/go-algorand/blob/e32b2c2ef9423526bbec7d749d912d3a6153410b/test/scripts/e2e_subs/tealprogs/app-escrow.teal#L116-L142

```python
withdraw_amount = Btoi(Txn.application_args[1])
handle_withdraw = Seq(
    App.localPut(Int(0), balance_key, App.localGet(Int(0), balance_key) - withdraw_amount),
    InnerTxnBuilder.Begin(),
    # set one field
    InnerTxnBuilder.SetField(TxnField.type_enum, TxnType.Payment),
    # set multiple fields at once
    InnerTxnBuilder.SetFields({
        TxnField.amount: withdraw_amount,
        TxnField.receiver: Txn.sender()
    }),
    InnerTxnBuilder.Submit(),
    Approve(),
)
```

## Priority
Needed for TEAL 5 support.

jasonpaulos:
Closes #112

In addition to supporting inner transactions, this PR also:
* Adds all TEAL 5 opcodes to the `Op` enum
* Adds new global and txn fields from TEAL 5
* Allows zero element `Seq` expressions (this is a no-op, but it's needed to support `InnerTxnBuilder.SetFields({})`)
* Enables dynamic array access, closes #109
* Consolidates all txn opcode tests (`txn`, `gtxn`, `itxn`)


shiqizng (commented):
this looks good to me. adding descriptions to the new classes, such as InnerTxnActionExpr, would be helpful. and what happened to the tests for gtxn/gtxnas?

jasonpaulos:
I've just added documentation for inner transactions -- thank you for reminding me!

The existing tests for txn/gtxn were very repetitive, so instead of repeating them again for itxn, I got rid of all the handwritten tests in favor of the new method `test_txn_fields` in `txn_test.py`. This methods iterates through all possible fields and all the different txn objects to make sure they get converted properly into the IR layer.

pyteal/ast/gtxn.py
```diff
@@ -45,34 +45,59 @@ def __teal__(self, options: "CompileOptions"):
...
+            return TealBlock.FromOp(options, op, cast(Expr, self.index))
+
+        if type(self.index) is int:
+            op = TealOp(self, opToUse, self.field.arg_name, cast(int, self.index))
+            return TealBlock.FromOp(options, op, cast(Expr, self.txnIndex))
+
+        op = TealOp(self, opToUse, self.field.arg_name)
+        return TealBlock.FromOp(
+            options, op, cast(Expr, self.txnIndex), cast(Expr, self.index)
+        )
```
ahangsu, line 100: nit: can remove some unnecessary `cast` to following:

```suggestion
        if type(self.txnIndex) is int:
            if type(self.index) is int:
                op = TealOp(self, opToUse, self.txnIndex, self.field.arg_name, self.index)
                return TealBlock.FromOp(options, op)

            op = TealOp(self, opToUse, self.txnIndex, self.field.arg_name)
            return TealBlock.FromOp(options, op, cast(Expr, self.index))

        if type(self.index) is int:
            op = TealOp(self, opToUse, self.field.arg_name, self.index)
            return TealBlock.FromOp(options, op, cast(Expr, self.txnIndex))

        op = TealOp(self, opToUse, self.field.arg_name)
        return TealBlock.FromOp(
            options, op, cast(Expr, self.txnIndex), cast(Expr, self.index)
        )
```

still passes `mypy pyteal` on my side.

shiqizng (approved):
looks good to me

pyteal/ast/gtxn.py
jasonpaulos: Good suggestion. I updated my mypy and this is passing now 👍

Issue #80, opened 2021-07-26
Identity element for logic operations


CiottiGiorgio:
## Problem

Right now if I write `Int(1)` as argument of `And` it gets actually compiled to the code. It would be nice if the compiler could calculate that an identity element doesn't change the result of the operation.
For example if I want to insert a check depending on the value of some parameter I actually have to change the whole expression.

`return core_checks if value == 0 else And(core_checks, other_checks)`

should become

`return And(core_checks, identity_element if value == 0 else other_checks)`

## Solution

The compiler should detect identity elements for each operator and simplify them.

## Urgency

No urgency at all.

jasonpaulos:
Allowing single argument NaryExpressions sounds reasonable and simple to do.

I also agree that it would be nice if PyTeal could intelligently optimize away constants like the original issue suggests. There are many places this could happen, and identify elements in `And` and `Or` are one of them.

Another good place would be conditions, since it's pretty common to check if a number is equal to 0 or 1. However, this is a redundant check in TEAL, since 0 and 1 can already be used to represent boolean values. For example, the TEAL code `<number-to-check>; int 0; ==` is equivalent to just `<number-to-check>; !`.

PR #117, opened 2021-09-17
Support logs, dynamic arg indexes, and other improvements

### LINKED ISSUE #113: Support for logging
## Summary
TEAL 5 introduces the `log` opcode, which can be used to add byteslices to a list of logged values for the current transaction.

## Scope
PyTeal should expose this functionality as well. Perhaps a simple `Log` expression will work for this?

## Priority
Needed to support TEAL 5.

jasonpaulos:
This PR adds:
* Support for logging with the `Log` expression
* Support for dynamic indexes into the `Arg` expression
* Support for single-element NaryExpr, e.g. `And(Int(1))`
* Support for creating `Bytes` expressions directly from `bytes` and `bytearray` objects, e.g. `Bytes(b"value")`

Closes #113.

### DIFF
--- a/pyteal/ast/bytes.py
+++ b/pyteal/ast/bytes.py
@@ -28,8 +28,8 @@ class Bytes(LeafExpr):
 
         For UTF-8 strings:
             Pass the string as the only argument. For example, ``Bytes("content")``.
-        For raw byte or bytearray objects:
-            Pass the byte or bytearray as the only argument. For example, ``Bytes(b"content")``.
+        For raw bytes or bytearray objects:
+            Pass the bytes or bytearray as the only argument. For example, ``Bytes(b"content")``.
         For base16, base32, or base64 strings:
             Pass the base as the first argument and the string as the second argument. For example,
             ``Bytes("base16", "636F6E74656E74")``, ``Bytes("base32", "ORFDPQ6ARJK")``,
--- a/pyteal/ast/bytes.py
+++ b/pyteal/ast/bytes.py
@@ -1,4 +1,4 @@
-from typing import TYPE_CHECKING
+from typing import Union, cast, overload, TYPE_CHECKING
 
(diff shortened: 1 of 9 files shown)


shiqizng (commented):
it looks good to me. But I don't see why support for single-element naryExpr is needed. why is Arg(Int(1)) or Concat(Bytes("a")) useful?

jasonpaulos:
Single element NaryExpr was a request from #80. The use case is to write something like `And(*[validate(thing) for thing in things])` without worrying about a special case if the array has only 1 element.

pyteal/ast/bytes.py
```diff
@@ -29,35 +39,42 @@ def __init__(self, *args: str) -> None:
...
-            self.base, byte_str = args
+        if arg2 is None:
+            if type(arg1) is str:
+                self.base = "utf8"
+                self.byte_str = escapeStr(arg1)
+            elif type(arg1) in (bytes, bytearray):
+                self.base = "base16"
+                self.byte_str = cast(Union[bytes, bytearray], arg1).hex()
+            else:
+                raise TealInputError("Unknown argument type: {}".format(type(arg1)))
```
algochoi, line 50: Small suggestion: consider using [f-strings](https://www.python.org/dev/peps/pep-0498/) instead of `.format()`

algochoi (approved):
Code and tests look good to me 🚢 
Just one small suggestion on formatting strings with f-strings

shiqizng (approved):

pyteal/ast/bytes.py
jasonpaulos: Good idea, but this should probably be a separate issue, since `format` is used in so many places in the codebase.

algochoi: Sounds good to me

PR #118, opened 2021-09-22
Add extract expressions

jasonpaulos:
Adds expressions for using new TEAL 5 extract opcodes.

### DIFF
--- a/pyteal/ast/binaryexpr_test.py
+++ b/pyteal/ast/binaryexpr_test.py
@@ -5,7 +5,7 @@ from .. import *
 # this is not necessary but mypy complains if it's not included
 from .. import CompileOptions
 
-options = CompileOptions()
+options = CompileOptions(version=5)
 
 
 def test_add():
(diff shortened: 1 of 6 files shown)


pyteal/ast/binaryexpr_test.py
```diff
@@ -1385,3 +1385,40 @@ def test_b_ge_invalid():
...
+        )
+
+        actual, _ = expr.__teal__(options)
+        actual.addIncoming()
+        actual = TealBlock.NormalizeBlocks(actual)
+
+        assert actual == expected
+
+
+def test_extract_uint_invalid():
```
algochoi, line 1418: Should there be a test that fails when `Teal4Options` are passed in?

algochoi (commented):
The code looks good to me, just had two questions:

1. Will there be a way to force the `substring3` or `extract3` TEAL op code using Pyteal? I guess there isn't much of a usecase right now, but those ops do save a couple bytes (assuming they are uint8) if I recall correctly. 
2. There is a somewhat hidden shortcut in `extract` where if you specify the length to be `0` as an immediate, it extracts until the end of the string. I think this is linked to point 1 and wondering if it should be supported.

pyteal/ast/binaryexpr_test.py
jasonpaulos: Good point, adding this now

jasonpaulos:
I agree it would be nice if there was a way to generate the `substring` and `extract` opcodes, since they are more efficiently encoded.

Normally if an operation can accept either a stack value or an immediate, PyTeal allows `Union[int, Expr]` and internally chooses the correct opcode based on the arguments. However, with these I think a better approach would be to have some sort of optimization that checks if the input expressions can be transformed into integers, and if that's the case and the integers aren't too large, then the immediate form of the opcode is used. Furthermore, in most cases it will probably be more efficient to use `extract` over `substring`, so the optimization could also make this happen when it's beneficial.

For `extract` the length arg being 0 indicates the end of the string, but for `extract3` there is no equivalent to this, correct? I.e. if I pass in 0 as the length arg, I will get a 0 length substring. If this is the case, that's a constraint we'll need to keep in mind when writing an optimization for this.

Since there's a fair bit of complexity here, I think it would be better to address this in a separate issue.

pyteal/ast/ternaryexpr_test.py
```diff
@@ -72,6 +72,38 @@ def test_substring_invalid():
         Substring(Bytes("my string"), Int(0), Txn.sender())
 
 
+def test_extract():
```
algochoi, line 75: We might want to check for the `Teal4Options` compile error here too?

algochoi (commented):
I think that sounds good -- Yes the `0` length trick will only work with the immediates `extract` opcode rather than the `extract3` op code.

pyteal/ast/ternaryexpr_test.py
jasonpaulos: Good catch, I'll add this

algochoi (approved):
LGTM 🚢

PR #120, opened 2021-09-24
Remove unnecessary branches

jasonpaulos:
This PR adds additional code to `TealBlock.NormalizeBlocks` to remove unnecessary intermediate blocks with 0 ops. Blocks like this are added by `If` and other branching expressions in order to converge divergent branches. They are harmless, but it would be more efficient to get rid of them.

For example, consider the following expression:
```python
If(Int(1)).Then(Int(2)).ElseIf(Int(3)).Then(Int(4)).Else(Int(5))
```

Before this PR, it compiles to:
```
#pragma version 2
int 1
bnz main_l5
int 3
bnz main_l4
int 5
main_l3:
b main_l6
main_l4:
int 4
b main_l3
main_l5:
int 2
main_l6:
return
```

With this PR, it now compiles to:
```
#pragma version 2
int 1
bnz main_l4
int 3
bnz main_l3
int 5
b main_l5
main_l3:
int 4
b main_l5
main_l4:
int 2
main_l5:
return
```

### DIFF
--- a/pyteal/ir/tealblock.py
+++ b/pyteal/ir/tealblock.py
@@ -215,10 +215,7 @@ class TealBlock(ABC):
                     for i, incomingBlock in enumerate(outgoingBlock.incoming):
                         if block is incomingBlock:
                             # remove block from incoming of outgoing
-                            outgoingBlock.incoming = (
-                                outgoingBlock.incoming[:i]
-                                + outgoingBlock.incoming[i + 1 :]
-                            )
+                            outgoingBlock.incoming.pop(i)
                             break
 
                     for prev in block.incoming:
--- a/pyteal/ir/tealblock.py
+++ b/pyteal/ir/tealblock.py
@@ -204,6 +204,31 @@ class TealBlock(ABC):
                     if prev is start:
                         start = block
 
+        for block in TealBlock.Iterate(start):
+            if len(block.ops) == 0:
+                outgoing = block.getOutgoing()
+                if len(outgoing) == 1:
+                    # if block has 0 ops and 1 outgoing edge, directly connect every incoming block
+                    # to the single outgoing block, thereby removing an unnecessary intermediate
+                    # jump to this block
+                    outgoingBlock = outgoing[0]
+                    for i, incomingBlock in enumerate(outgoingBlock.incoming):
+                        if block is incomingBlock:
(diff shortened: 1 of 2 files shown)


jannotti (commented):
How do we version this stuff?  Since it changes compilation, logicsig authors may be sad.  Do you only introduce these in new teal versions?

jasonpaulos:
This change will get released with v0.9.0 of pyteal, which is allowed to introduce breaking changes. In the past we've changed the code generator on releases like this out of necessity (and always with a warning in the changelog), so there is precedence for this.

pyteal/ir/tealblock.py
```diff
@@ -204,6 +204,31 @@ def NormalizeBlocks(cls, start: "TealBlock") -> "TealBlock":
...
+                outgoing = block.getOutgoing()
+                if len(outgoing) == 1:
+                    # if block has 0 ops and 1 outgoing edge, directly connect every incoming block
+                    # to the single outgoing block, thereby removing an unnecessary intermediate
+                    # jump to this block
+                    outgoingBlock = outgoing[0]
+                    for i, incomingBlock in enumerate(outgoingBlock.incoming):
+                        if block is incomingBlock:
+                            # remove block from incoming of outgoing
+                            outgoingBlock.incoming = (
```
algochoi, line 218: Is there a difference between using `incoming.remove(i)` here ? Are we afraid of indexing errors?

algochoi (approved):
Code looks good to me :shipit: 
Just had one question about manual list splicing.

pyteal/ir/tealblock.py
jasonpaulos: `remove` compares by value (by calling `__eq__`), but here we want to compare by reference. There may be multiple blocks with the same code (e.g. `int 1; return` is a common block that can appear more than once in a program) which return true under value comparison, but false under reference comparison.

algochoi: Sorry, I think I meant `incoming.pop(index)` (or some equivalent of removing an element by the index) -- you're totally right about the `remove` I think.

jasonpaulos: Good idea, pop would be better here -- I'll update this now

PR #136, opened 2021-11-08
rm created_application_id

barnjamin:
application calls not yet allowed on inner transactions


jasonpaulos:
Closing this because that field is still valid, it just won't be useful until TEAL 6 (which is on the way)

PR #142, opened 2021-12-06
Fix `App.globalGetEx` docstring

jasonpaulos:
The docstring for `App.globalGetEx` mistakenly says that it will accept bytes for the app ID.


tzaffi (approved):
LGTM

PR #146, opened 2021-12-20
up max teal version

barnjamin:
Just upping the max version allowed so if folks are interested in toying around they can. 

Generally I think we should try to do this for future versions as soon as a new teal version hits master/feature branches of go-algorand, even if we haven't defined the new ops yet

Optionally I can target some new feature branch here instead of trying to merge to master

### DIFF
--- a/pyteal/compiler/compiler_test.py
+++ b/pyteal/compiler/compiler_test.py
@@ -134,7 +134,7 @@ def test_compile_version_invalid():
         compileTeal(expr, Mode.Signature, version=1)  # too small
 
     with pytest.raises(TealInputError):
-        compileTeal(expr, Mode.Signature, version=6)  # too large
+        compileTeal(expr, Mode.Signature, version=MAX_TEAL_VERSION+1)  # too large
 
     with pytest.raises(TealInputError):
         compileTeal(expr, Mode.Signature, version=2.0)  # decimal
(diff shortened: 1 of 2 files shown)


tzaffi (approved):
LGTM as we already have some opcodes in `go-algorand` that [require v6](https://github.com/algorand/go-algorand/blob/v3.2.2-stable/data/transactions/logic/opcodes.go#L320) so seems useful. @algochoi - could you chime in as well?

tzaffi:
But there are some failing tests that are checking for version _not_ being greater than 5, so these need to modd'ed as well.

algochoi:
Looks fine to me, but I'm not too sure how useful this might be without defining any TEAL 6 opcodes like mentioned above.

barnjamin:
Teal 6 allows for some extra inner transaction fields to be set, namely the one I wanted was rekey_to and its only supported >= v6. 

I found myself generating the teal then tweaking the first line but this seems easier.

algochoi (approved):

pyteal/compiler/compiler_test.py
```diff
@@ -134,7 +134,7 @@ def test_compile_version_invalid():
         compileTeal(expr, Mode.Signature, version=1)  # too small
 
     with pytest.raises(TealInputError):
-        compileTeal(expr, Mode.Signature, version=6)  # too large
+        compileTeal(expr, Mode.Signature, version=MAX_TEAL_VERSION + 1)  # too large
```
tzaffi, line 137: nit: can we just say `version=7` instead of `MAX_TEAL_VERSION +_ 1`.

Yes, it will fail again when we up the version again, but that will remind us that we have this test in place.

jasonpaulos (commented):
It would be nice if you could add a `test_compile_version_6` test too 😄 

Otherwise, looks good

jasonpaulos (approved):

PR #147, opened 2021-12-23
using the subroutine name for the label

barnjamin:


barnjamin:
@jasonpaulos @tzaffi what do you think of the above? 

submitted this pr just to demonstrate how it was done but not sure its a good idea so closed it

PR #148, opened 2021-12-27
Formatting subroutines with name and newline

barnjamin:
To make the TEAL output a little more readable, add a newline before the labels.  This will be removed during assembly to bytecode so should be no change to bytecode/addresses. 

Updated the tests and examples to account for this as well.

### DIFF
--- a/examples/signature/dutch_auction.teal
+++ b/examples/signature/dutch_auction.teal
@@ -12,6 +12,7 @@ int 1
 ==
 bnz main_l4
 err
+
 main_l4:
 txn RekeyTo
 global ZeroAddress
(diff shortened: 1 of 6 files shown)


jasonpaulos:
This actually adds a newline before subroutine labels _and_ internal labels used for conditional branching (e.g. from if statements). It seem like it would be better if only subroutine labels had a newline before them?

barnjamin:
Good point, reworked to add new lines in comment if its set, this serves to break up the spacing but only for subroutines with a comment.

jasonpaulos (approved):
Looks good

PR #153, opened 2021-12-30
`method` pseudo-op support for ABI methods

### LINKED ISSUE #150: Use the `method` pseudo-op when routing to ABI methods
## Problem

Currently the `method` pseudo-op which converts an ABI method's signature to its selector is not available in PyTEAL. Utilizing this method, should make it easier to understand and debug ABI apps that are written in PyTEAL.

## Solution (at 30,000 feet)

Introduce `Op.method(a_method)` that takes the ABI method `a_method` with signature `THE_SIGNATURE` and when converted to TEAL has the effect of creating the following snippet:
```
method "THE_SIGNATURE"
```

That's just a shot in the dark, there are probably better ways to achieve this.

## Urgency

Moderate - the bang for the buck on this one is pretty high.

ahangsu:
## Tasks

- [x] Implement and test `Method`
- [x] Additional `Method` support in `Subroutine`

Closes #150 

## NOTE

Currently, if the subroutine in PyTeal is specified as an ABI method (by providing a method signature in subroutine declaration), the produced Teal code's subroutine label's comment is changed to method signature (not Py function name).

Do we want to also change the label name in compiler? I think in https://github.com/algorand/pyteal/pull/147 we can do this.


pyteal/ast/method_test.py
```diff
@@ -0,0 +1,25 @@
...
+from pyteal.ast.method import Method
+
+from .. import *
+
+
+def test_method():
+    expr = Method("add(uint64,uint64)uint64")
+    assert expr.type_of() == TealType.bytes
+
+    expected = TealSimpleBlock([TealOp(expr, Op.method, "add(uint64,uint64)uint64")])
```
jasonpaulos, line 12: The method name should be surrounded in quotes. Right now it outputs `method add(uint64,uint64)uint64`, but it should be `method "add(uint64,uint64)uint64"`.

See the implementation here: https://github.com/jannotti/go-algorand/blob/168cfe58f1d99b0289accc67098f69482d400295/data/transactions/logic/assembler.go#L655-L671

pyteal/ast/subroutine.py
```diff
@@ -19,7 +18,10 @@ class SubroutineDefinition:
     nextSubroutineId = 0
 
     def __init__(
-        self, implementation: Callable[..., Expr], returnType: TealType
+        self,
+        implementation: Callable[..., Expr],
+        returnType: TealType,
+        methodSignature: str = None,
```
jasonpaulos, line 24: I'm less sure about the changes to this file, since not every subroutine will represent a method.

What if you changed `methodSignature` to just `name`? I.e. if you provide the `name` arg, then it would override whatever the function's `__name__` is.

@barnjamin thoughts on this?

jasonpaulos (commented):
You probably also want to modify the `createConstantBlocks` function to recognize the method pseudo op -- it already handles addr in a similar way: https://github.com/algorand/pyteal/blob/a9c42e307e548cc2e6a1fa8b38149e240ec07ceb/pyteal/compiler/constants.py#L97

pyteal/ast/subroutine.py
barnjamin: I had thought we'd be able to infer the method signature from the method definition, including name and types it accepts. That would mean defining the ABI types with accessors.

pyteal/ast/method_test.py
ahangsu: oops...

pyteal/ast/subroutine.py
barnjamin: I started something here https://github.com/barnjamin/pyteal-utils/blob/main/pytealutils/abi/types.py but maybe this kind of thing belongs in PyTEAL proper.

We should _definitely_ have some representation of them so decoding from the over-the-wire format is painless

barnjamin: I ended up just adding the ABI types to utils for now using the py-sdk to test encoding/decoding:  https://github.com/algorand/pyteal-utils/blob/a5f35c9e217b2c6e259b8a725bb988ec631e6fc8/pytealutils/abi/test_fixed_point.py

ahangsu: Just changed `methodSignature` name to `nameStr`, and override function's `__name__` on initialization, but I still feel this needs to be polished as an implementation, and should be modified after ABI in PyTeal is introduced.

I would like to keep this thread open for more discussion on the design of subroutine (in combination with ABI methods).

pyteal/compiler/constants.py
```diff
@@ -94,6 +90,20 @@ def extractAddrValue(op: TealOp) -> Union[str, bytes]:
...
+def extractMethodValue(op: TealOp) -> bytes:
+    """Extract the constant value being loaded by a TealOp whose op is Op.method.
+
+    Returns:
+        The bytes of method selector computed from the method signature that the op is loading.
+    """
+    if len(op.args) != 1 or type(op.args[0]) != str:
+        raise TealInternalError("Unexpected args in method opcode: {}".format(op.args))
+
+    methodSignature = cast(str, op.args[0])
```
jasonpaulos, line 102: I think need to discard the leading and trailing `"` before hashing

pyteal/compiler/constants_test.py
```diff
@@ -63,6 +64,31 @@ def test_extractAddrValue():
...
+        (TealOp(None, Op.method, "empty()void"), b"\xa8\x8c\x26\xa5"),
+        (TealOp(None, Op.method, "payment(pay,uint64)bool"), b"\x3e\x3b\x3d\x28"),
+        (
+            TealOp(
+                None,
+                Op.method,
+                "referenceTest(account,application,account,asset,account,asset,asset,application,application)uint8[9]",
+            ),
+            b"\x0d\xf0\x05\x0f",
+        ),
```
jasonpaulos, line 84: These tests should have quotes around the method signature args to be accurate

pyteal/ast/method.py
```diff
@@ -0,0 +1,43 @@
...
+
+from ..types import TealType
+from ..ir import TealOp, Op, TealBlock
+from .leafexpr import LeafExpr
+
+if TYPE_CHECKING:
+    from ..compiler import CompileOptions
+
+
+class Method(LeafExpr):
```
jasonpaulos, line 14: Renaming this to `MethodSignature` would probably be more clear. PyTeal may want to use `Method` to refer to a method definition in the future

pyteal/compiler/constants.py
ahangsu: The info stored in `Method` is unprocessed, i.e., without leading/tailing `"`: see https://github.com/algorand/pyteal/blob/feature/method-pseudo-op/pyteal/ast/method.py#L33, where we add `"` in compilation.

pyteal/compiler/constants_test.py
ahangsu: This is taken from your test script: https://gist.github.com/jasonpaulos/99e4f8a75f2fc2ec9b8073c064530359, and the tests come from hash without quotes...

pyteal/compiler/constants.py
```diff
@@ -94,6 +90,22 @@ def extractAddrValue(op: TealOp) -> Union[str, bytes]:
...
+    if methodSignature[0] == methodSignature[-1] and methodSignature.startswith('"'):
+        methodSignature = methodSignature[1:-1]
```
jasonpaulos, line 104: The quotes are not optional, so this code should probably throw an error if the argument does not start and end with a quote.

jasonpaulos (approved):
Looks good!

PR #154, opened 2022-01-07
ABI Types

barnjamin:
Implements the following ABI types in PyTeal:
- Uint8 (and Byte), Uint16, Uint32, Uint64
- Bools
- Tuples
- Arrays
  - Static
  - Dynamic

Other types and uses for the ABI types will be addressed by other issues.

Checklist:
- [x] Figure out the best way to represent these types in PyTeal
- [x] Implement the above types 
- [x] Support bools when encoding and decoding arrays and tuples
- [x] For increased decoding efficiency, add more methods for different decoding contexts. Right now we just have `decode(encoded,start,length)`, but other useful forms would be `decodeAll(encoded)` and `decodeSubstring(encoded,start,end)` (?)
- [x] Add bounds checks for small integers
- [x] Test all the types
- [ ] Add docstrings
- [x] In the interest of getting this PR merged as quickly as possible, remove the array `map` function and make a separate issue/PR for it


pyteal/ast/abi/collections.py
```diff
@@ -194,28 +346,41 @@ def decode(self, encoded: Expr, offset: Expr = None) -> "ABIDynamicArrayType":
...
-        return Concat(self.len.encode(), self.value)
+        return encodeArray(self.data, True)
 
     def __getitem__(self, index: Union[int, Expr]) -> ABIValue:
         if type(index) is int:
             index = Int(index)
-        return indexArray(self.type.elementType, self.value, index)
+        return indexArray(self.type.elementType, self.data, index)
+
+    def forEach(self, action: Callable[[ABIValue, Expr], Expr]) -> Expr:
```
barnjamin, line 379: ❤️

pyteal/ast/abi/uint.py
```diff
@@ -0,0 +1,99 @@
...
+    def decode(self, encoded: Expr, offset: Expr, length: Expr) -> Expr:
+        return Seq(
+            Assert(length == Int(self.byte_length_static())),  # TODO: remove?
+            self.set(ExtractUint16(encoded, offset)),
+        )
+
+    def encode(self) -> Expr:
+        # value might exceed a uint16, need to check at runtime
+        return Seq(
+            Assert(Not(self.get() >> Int(2))),
```
CiottiGiorgio, line 69:
```suggestion
            Assert(Not(self.get() >> Int(16))),
```

jasonpaulos: Good catch! Though why 17 and not 16? `2**16 >> 16 = 1`

CiottiGiorgio: Yes exactly. Furthermore, if every bit until the 16th is set, `2**17-1 >> 16 = 1`

jasonpaulos: I think we disagree by 1. The maximum value of a 16-bit unsigned int is `2**16-1 = 65535`. So it should be sufficient to right shift by 16 to see if anything is equal to or greater than that.

My original code meant 2 bytes, but of course shifting is measured in bits, not bytes, so it should be 16.

CiottiGiorgio: Ah yes you are right. I didn't have my engineering hat on.

pyteal/ast/abi/uint.py
```diff
@@ -0,0 +1,195 @@
...
+
+    def new_instance(self) -> "Uint16":
+        return Uint16()
+
+    def set(self, value: Union[int, Expr]) -> Expr:
+        if type(value) is int:
+            if value >= 2 ** 16:
+                raise ValueError("Value exceeds Uint16 maximum: {}".format(value))
+            value = Int(value)
+        # TODO: check dynamic value bounds?
```
ahangsu, line 101: nit: is it possible (or worth it) to try something like
```python
Seq(
    Assert(value < Int(2 ** 16)),
    self.stored_value.store(cast(Expr, value)
)
```
for the stack variable (which is the dynamic value bound here)?

jasonpaulos: Yes, that's the kind of check this comment is referencing. I think for correctness we want this, but for efficiency we would also want an optimization step that removes as much of these checks as possible if they are not necessary.

ahangsu: I see, the question is more on if it "worth" adding these checks on stack variables then.

jasonpaulos: I think I've decides, yes, this check should be here. A later PR can attempt to optimize it

ahangsu (commented):
This will become a quite large PR, so I decided to separate my review into small reviews for components. Sorry for the inconvenience in advance...

I reviewed the `uint` part, looks largely good to me, have some followup questions:
- We currently have support for `uint` with bit size of power of 2 (e.g., 8/16/32/64). Will we later support the following:
  - `uint` types will non pow-of-2 bit size within [8, 64]
  - `uint` types with bit size larger than 64

  If there are tickets for more implementation, that would be great.
  An approach I am thinking of is using metaclass to make implementation smaller (by parameterizing the classes, rather than objects), but I am unsure how much this would make our days easier.
- `ufixed` support is still tbd, wondering how this would be approached, since underlying Teal (does not?) support fixed point number.
  The only thing I am thinking of is deriving `ufixed` from `uint`, and treat `ufixed` bytes just like `uint`. Not sure how you would approach this...

pyteal/ast/abi/uint.py
```diff
@@ -0,0 +1,225 @@
...
+        if type(value) is Uint8:
+            value = value.get()
+            checked = True
+
+        if checked:
+            return self.stored_value.store(cast(Expr, value))
+
+        return Seq(
+            self.stored_value.store(cast(Expr, value)),
+            Assert(self.stored_value.load() < Int(2 ** self.bit_size)),
```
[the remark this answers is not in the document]
jasonpaulos: The assertion is here so we fail as early as possible if the program tries to put a value that's too large into this integer type.

Previously this check was in the `encode` method, but I realized that was too late, since the value might be used in a computation before being encoded.

And the fact that the assertion happens after the store is because this order is more efficient if the value being stored is a compound expression like `Int(1) + Int(2)`. It's the difference between

```
int 1
int 2
+
int 2 ** self.bit_size # not valid code, but this is the value from pyteal
<
assert
int 1
int 2
+
store X
```

and

```
int 1
int 2
+
store X
load X
int 2 ** self.bit_size
<
assert
```

Practically asserting before or after the value is stored is the same thing from a correctness point of view.

pyteal/ast/abi/uint.py
```diff
@@ -0,0 +1,225 @@
...
+            startIndex = Int(0)
+        return self.stored_value.store(GetByte(encoded, startIndex))
+
+    def encode(self) -> Expr:
+        return SetByte(Bytes(b"\x00"), Int(0), self.get())
+
+
+Uint8.__module__ = "pyteal"
+
+Byte = Uint8
```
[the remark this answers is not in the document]
jasonpaulos: Good point. Originally I thought these two could use the same class, but since `uint8` and `byte` are different at the method signature level, I think `byte` needs to be its own class with its own `__str__`

jasonpaulos:
@ahangsu to answer your questions:
1. Yes I created a follow up issue to finish the other integers: https://github.com/algorand/pyteal/issues/181. Also, I completely agree that parameterizing the uint classes seems like the best way to approach this, but I wasn't familiar enough with that process to attempt it in this PR.
2. You're right that the AVM does not support fixed point numbers natively, but it's possible to simulate them from a higher level language. @barnjamin's original PR has an implementation of them: https://github.com/algorand/pyteal/blob/2e1447f12c5429aa5cad08cd213907e3dd7af77a/pyteal/abi/ufixed.py, and we also have a follow up issue: https://github.com/algorand/pyteal/issues/184

pyteal/ast/abi/uint.py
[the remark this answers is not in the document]
ahangsu: I see, I overlooked the perspective of generating smaller quantity of code (since it's `Expr`). That's a good point.

pyteal/ast/abi/uint.py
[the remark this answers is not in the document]
ahangsu: I can write a small pr to this one for `byte`, hope we can bring this out with more complete features. 
I am thinking of inherit directly from `uint8` and reuse most of the things, but I do not have a clear idea how to do conversion between `Byte` and `Uint8`.

pyteal/ast/abi/uint_test.py
```diff
@@ -0,0 +1,251 @@
...
+        actual, _ = expr.__teal__(options)
+
+        assert actual == expected
+
+
+def test_Uint_decode():
+    for test in testData:
+        for startIndex in (None, Int(1)):
+            for endIndex in (None, Int(2)):
+                for length in (None, Int(3)):
```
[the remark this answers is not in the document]
ahangsu: I agree with your thoughts, so long as `extract` has enough bytes to take, then it should be fine.

pyteal/ast/abi/uint.py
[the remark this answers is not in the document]
ahangsu: PR in #194

pyteal/ast/abi/type.py
```diff
@@ -0,0 +1,139 @@
...
+
+    @abstractmethod
+    def __str__(self) -> str:
+        pass
+
+
+Type.__module__ = "pyteal"
+
+
+class ComputedType(ABC, Generic[T]):
```
ahangsu, line 90: nit: I think we will still need something inherit from `ComputedType` to reference returned ABI value from subroutine, since the `store_into` is a placeholder right now...

pyteal/ast/abi/bool.py
```diff
@@ -0,0 +1,98 @@
...
+        if type(t) is Bool:
+            numBools = consecutiveBools(types, i)
+            ignoreNext = numBools - 1
+            length += boolSequenceLength(numBools)
+            continue
+        length += t.byte_length_static()
+    return length
+
+
+def consecutiveBools(types: Sequence[Type], startIndex: int) -> int:
```
ahangsu, line 78: nit: consider renaming to `consecutiveBoolNum` for readability.

pyteal/ast/abi/bool.py
```diff
@@ -0,0 +1,98 @@
...
+        if ignoreNext > 0:
+            ignoreNext -= 1
+            continue
+        if type(t) is Bool:
+            numBools = consecutiveBools(types, i)
+            ignoreNext = numBools - 1
+            length += boolSequenceLength(numBools)
+            continue
+        length += t.byte_length_static()
+    return length
```
ahangsu, line 75: nit: personally prefer to use `for i in range(...)` and just let `i += numBools - 1` to skip iteration. Can ignore this, not a big deal.

pyteal/ast/abi/bool_test.py
```diff
@@ -0,0 +1,332 @@
...
+
+    assert actual == expected
+
+
+def test_Bool_decode():
+    boolType = abi.Bool()
+    for startIndex in (None, Int(1)):
+        for endIndex in (None, Int(2)):
+            for length in (None, Int(3)):
+                encoded = Bytes("encoded")
```
ahangsu, line 131: nit: can move this line out of loop

pyteal/ast/abi/bool.py
```diff
@@ -0,0 +1,98 @@
...
+        return False
+
+    def byte_length_static(self) -> int:
+        # Not completely accurate since up to 8 consecutive bools will fit into a single byte
+        return 1
+
+    def get(self) -> Expr:
+        return self.stored_value.load()
+
+    def set(self, value: Union[bool, Expr, "Bool"]) -> Expr:
```
ahangsu, line 33: nit: there is a possibility that one can directly set a `TealType.uint64` inside scratch value, do you think it would cause an issue?

pyteal/ast/abi/bool_test.py
```diff
@@ -0,0 +1,332 @@
...
+        assert actual == expected
+
+
+def test_Bool_get():
+    boolType = abi.Bool()
+    expr = boolType.get()
+    assert expr.type_of() == TealType.uint64
+    assert not expr.has_return()
+
+    expected = TealSimpleBlock([TealOp(expr, Op.load, boolType.stored_value.slot)])
```
ahangsu, line 119: just one dumb question from my side: what is the difference for `TealOp` to have first argument to have either `Expr` or `None`? Or so to say, in what scenario `TealOp` will want to accept `Expr`, and otherwise when do we want to give `None`?

ahangsu (commented):
This is the review for `type` and `bool` part, still reading `array` part. Mostly look good to me.

As I was reading `type.py`, a new idea about subroutine and subroutine name spun up in my mind: Essentially, on the perspective of ABI router, the registered method must satisfy all argument abiType and return abiType. This infers that we can remove the `name` part in subroutine decorator, and directly compute method name/signature from subroutine definition, which simplifies the syntax again.

pyteal/ast/abi/bool.py
jasonpaulos: I don't think that actually skips iteration? I believe it only changes the current value of `i` and does not advance the iterator.

ahangsu: oof... I was thinking I am writing cpp...

pyteal/ast/abi/bool.py
jasonpaulos: Good point, we may want to validate that the type is only 0 or 1 🤔

pyteal/ast/abi/bool_test.py
jasonpaulos: The first argument is the `Expr` which created the `TealOp`, used only for error messages. For testing, if it's easy to construct the expected response which correct `Expr`s, then generally I think we should do that. Otherwise, None should be used for the `Expr` and the context `TealComponent.Context.ignoreExprEquality()` should be enabled before comparing with `==`

pyteal/ast/abi/bool.py
jasonpaulos: I will add a check for this similar to uint max values

ahangsu (commented):
I was reviewing `array.py` right now, some question arose about `ArrayElements`, and I am reading `tuple.py` now before I review again on `array` testing.

pyteal/ast/abi/array.py
```diff
@@ -0,0 +1,217 @@
...
+            "[]" if self._static_length is None else "[{}]".format(self._static_length)
+        )
+
+
+Array.__module__ = "pyteal"
+
+# until something like https://github.com/python/mypy/issues/3345 is added, we can't make the size of the array a generic parameter
+class StaticArray(Array[T]):
+    def __init__(self, valueType: T, length: int) -> None:
+        super().__init__(valueType, length)
```
ahangsu, line 101: oh snap I forgot, need a check of `length` here to be strictly positive.

pyteal/ast/abi/array.py
```diff
@@ -0,0 +1,217 @@
...
+        length: Expr = None
+    ) -> Expr:
+        extracted = substringForDecoding(
+            encoded, startIndex=startIndex, endIndex=endIndex, length=length
+        )
+        return self.stored_value.store(extracted)
+
+    def set(self, values: Sequence[T]) -> Expr:
+        if not all(self._valueType.has_same_type_as(value) for value in values):
+            raise TealInputError("Input values do not match type")
```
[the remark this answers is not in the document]
jasonpaulos: Good suggestion, I've modified this slightly and I'll commit it soon

pyteal/ast/abi/array.py
```diff
@@ -0,0 +1,217 @@
...
+        return self._valueType.is_dynamic()
+
+    def byte_length_static(self) -> int:
+        if self.is_dynamic():
+            raise ValueError("Type is dynamic")
+        if type(self._valueType) is Bool:
+            return boolSequenceLength(self.length_static())
+        return self.length_static() * self._valueType.byte_length_static()
+
+    def set(self, values: Sequence[T]) -> Expr:
```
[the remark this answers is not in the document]
jasonpaulos: Good point, I'll add this to Static and Dynamic arrays

pyteal/ast/abi/array.py
```diff
@@ -0,0 +1,217 @@
...
+
+DynamicArray.__module__ = "pyteal"
+
+
+class ArrayElement(ComputedType[T]):
+    def __init__(self, array: Array[T], index: Expr, length: Expr = None) -> None:
+        super().__init__(array._valueType)
+        self.array = array
+        self.index = index
+        self.length = length
```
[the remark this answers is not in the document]
jasonpaulos: Good point, this is leftover when when we had `Array.map` and `Array.forEach`. In those cases, it's more efficient to calculate the length once and then reuse it every time you need to load an element. But since we don't have those functions anymore, I will remove this

pyteal/ast/abi/array.py
```diff
@@ -0,0 +1,217 @@
...
+        else:
+            arrayLength = self.array.length()
+
+        encodedArray = self.array.encode()
+
+        if not self.array._valueType.has_same_type_as(output):
+            raise TealInputError("Output type does not match value type")
+
+        if type(output) is Bool:
+            return cast(Bool, output).decodeBit(encodedArray, offsetBit)
```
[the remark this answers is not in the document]
jasonpaulos: A great suggestion

pyteal/ast/abi/array.py
```diff
@@ -0,0 +1,217 @@
...
+class ArrayElement(ComputedType[T]):
+    def __init__(self, array: Array[T], index: Expr, length: Expr = None) -> None:
+        super().__init__(array._valueType)
+        self.array = array
+        self.index = index
+        self.length = length
+
+    def store_into(self, output: T) -> Expr:
+        offsetBit = self.index
+        offsetIndex = Int(self.array._stride) * self.index
```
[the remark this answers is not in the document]
jasonpaulos: I chose `byteIndex` to mirror `bitIndex` for the bool case

pyteal/ast/abi/array.py
```diff
@@ -0,0 +1,217 @@
...
+            raise TealInputError("Output type does not match value type")
+
+        if type(output) is Bool:
+            return cast(Bool, output).decodeBit(encodedArray, offsetBit)
+
+        if self.array._has_offsets:
+            valueStart = ExtractUint16(encodedArray, offsetIndex)
+            valueEnd = (
+                If(self.index + Int(1) == arrayLength)
+                .Then(arrayLength * Int(self.array._stride))
```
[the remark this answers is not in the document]
jasonpaulos: Ah, I think you're right. The expression `arrayLength * Int(self.array._stride)` calculates the end of the "header" portion of the array, not the entire array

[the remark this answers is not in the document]
jasonpaulos: I've just changed it, let me know what you think

pyteal/ir/tealblock.py
```diff
@@ -228,5 +228,40 @@ def NormalizeBlocks(cls, start: "TealBlock") -> "TealBlock":
...
+                if expectedSlot in mapFromActualToExpected.values():
+                    # this value was already seen
+                    return False
+                mapFromActualToExpected[actualSlot] = expectedSlot
+                continue
+
+            if mapFromActualToExpected[actualSlot] != expectedSlot:
+                return False
+
+        return True
```
ahangsu, line 264: These lines of code smells algorithmic, I suppose you are trying to decide a map between `actual` and `expected` satisfying some property (at least it feels acyclic).
My understanding is that, these lines of code ensures that
- cardinality of `actual` >= cardinality of `expected`
- overlapping part of `actual` and `expected` are self-mapping
- allowing more than 1 element map from `actual` to `expected`, but not reversely.

Would be nice to add some more explanation in docstring.

ahangsu (commented):
Largely finished the first pass over (almost) all the code in this PR. If you wont mind, I would like to help writing some docstring to explain the mechanism in some components, for the sake of further maintenance...

pyteal/ast/abi/array.py
```diff
@@ -98,6 +97,10 @@ def __str__(self) -> str:
 # until something like https://github.com/python/mypy/issues/3345 is added, we can't make the size of the array a generic parameter
 class StaticArray(Array[T]):
     def __init__(self, valueType: T, length: int) -> None:
+        if length < 0:
```
ahangsu, line 100: nit: by parsing https://github.com/algorand/go-algorand/blob/master/data/abi/abi_type.go#L122, it is not allowed to let array length == 0.

ahangsu (commented):
The array implementation now largely feels correct to me

pyteal/ast/abi/tuple.py
```diff
@@ -0,0 +1,284 @@
...
+        ignoreNext = 0
+        for i, typeAfter in enumerate(valueTypes[index + 1 :], start=index + 1):
+            if ignoreNext > 0:
+                ignoreNext -= 1
+                continue
+
+            if type(typeAfter) is Bool:
+                boolLength = consecutiveBoolNum(valueTypes, i)
+                nextDynamicValueOffset += boolSequenceLength(boolLength)
+                ignoreNext = boolLength - 1
```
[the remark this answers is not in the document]
jasonpaulos: Thank you! The current tests don't have coverage for this, I'll work on adding one that does

pyteal/ast/abi/array.py
jasonpaulos: I think it _should_ be allowed by the SDKs, since it is allowed by the [spec](https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0004.md#types) (says `N >= 0`), but it's not very important

Related issue: https://github.com/algorand/js-algorand-sdk/issues/497

ahangsu: oof, though it is meaningless to let static array length to be 0, which is 0 byte, but it should be okay to be valid... does not feel harmed.

pyteal/ir/tealblock.py
jasonpaulos: I added docs for this, let me know if it makes sense

ahangsu: looks good!

jasonpaulos:
Since this is being merged to a feature branch, I think it's ok if some docstrings are missing for now. Before the feature branch is merged to master, we should make sure to fix that though

ahangsu (approved):
Let's merge it in feature branch, and keep in mind to add lost docstring from time to time.

PR #163, opened 2022-01-18
Add BytesSqrt

StylishTriangles:
Added bsqrt from upcoming TEAL 6


pyteal/ast/unaryexpr_test.py
```diff
@@ -357,6 +358,27 @@ def test_b_not_invalid():
...
+    actual, _ = expr.__teal__(teal6Options)
+    actual.addIncoming()
+    actual = TealBlock.NormalizeBlocks(actual)
+
+    assert actual == expected
+
+
+def test_bsqrt_invalid():
+    with pytest.raises(TealTypeError):
+        BytesNot(Int(2 ** 64 - 1))
```
jasonpaulos, line 379:
```suggestion
        BytesSqrt(Int(2 ** 64 - 1))
```

This should be `BytesSqrt`

jasonpaulos (changes requested):
Thanks for adding this. I noticed one issue, other than that it looks good

pyteal/ast/unaryexpr_test.py
StylishTriangles: Whoops

jasonpaulos (approved):

PR #165, opened 2022-01-19
Acct params get

barnjamin:
Adding acct_params_get opcode as MaybeValue since it returns 2 items on the stack

`Stack: ..., A: uint64 → ..., X: any, Y: uint64`


pyteal/ast/acct.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class AcctParam:
+    @classmethod
+    def balance(cls, acct: Expr) -> MaybeValue:
+        """Get the current balance in microalgos an account.
+
+        Args:
+            acct: An index into Txn.accounts that corresponds to the account to check.
+                Must evaluate to uint64.
```
jasonpaulos, line 19: I believe the account can be an integer index into the accounts array or the actual address of the account.

pyteal/ast/acct.py
```diff
@@ -0,0 +1,62 @@
...
+from ..types import TealType, require_type
+from ..ir import Op
+from .expr import Expr
+from .maybe import MaybeValue
+
+if TYPE_CHECKING:
+    from ..compiler import CompileOptions
+
+
+class AcctParam:
```
jasonpaulos, line 12: nit: maybe rename to `AccountParam`? That seems more readable to me

pyteal/ast/acct.py
```diff
@@ -0,0 +1,62 @@
...
+        return MaybeValue(
+            Op.acct_params_get,
+            TealType.uint64,
+            immediate_args=["AcctMinBalance"],
+            args=[acct],
+        )
+
+    @classmethod
+    def authAddr(cls, acct: Expr) -> MaybeValue:
+        """Get the authorizing address for an account.
```
jasonpaulos, line 47: It would be great if you could mention that an account that hasn't been rekeyed will return the empty address for this

ahangsu:
Looks good to me, would be great with some minor test cases by mimicking `app_params_get`: https://github.com/algorand/pyteal/blob/master/pyteal/ast/app_test.py

pyteal/ast/acct.py
ahangsu: I am not sure about this: see https://github.com/algorand/go-algorand/blame/feature/c2c/data/transactions/logic/TEAL_opcodes.md#L956

barnjamin: the eval.go does call out to this method which just takes a `stackValue` and allows either an address or a ref to uint8 https://github.com/algorand/go-algorand/blob/4b1dfed080122c317df7e12e0d83c55871da7ff4/data/transactions/logic/eval.go#L3282

ahangsu: my bad, it seems README is not quite up to dated.

pyteal/ast/acct.py
```diff
@@ -0,0 +1,59 @@
...
+
+class AccountParam:
+    @classmethod
+    def balance(cls, acct: Expr) -> MaybeValue:
+        """Get the current balance in microalgos an account.
+
+        Args:
+            acct: An index into Txn.accounts that corresponds to the application to check or an address in the Txn.accounts array.
+                May evaluate to uint64 or an address.
+        """
```
ahangsu, line 20: nit: I think you might still want to use `require_type` here: e.g., https://github.com/algorand/pyteal/blob/master/pyteal/ast/app.py#L98, where we can require `anyType`.

Same applies to other class methods.

pyteal/ast/acct.py
```diff
@@ -0,0 +1,59 @@
...
+    from ..compiler import CompileOptions
+
+
+class AccountParam:
+    @classmethod
+    def balance(cls, acct: Expr) -> MaybeValue:
+        """Get the current balance in microalgos an account.
+
+        Args:
+            acct: An index into Txn.accounts that corresponds to the application to check or an address in the Txn.accounts array.
```
ahangsu, line 18: nit: I think the document for `acct` is not quite accurate, since by https://github.com/algorand/go-algorand/blob/4b1dfed080122c317df7e12e0d83c55871da7ff4/data/transactions/logic/eval.go#L3277, the app's own address is also allowed

ahangsu (commented):
Looks largely good to me, just some minor things to polish

pyteal/ast/acct.py
barnjamin: ah good call

jannotti (approved):
I'm no expert on pyteal, but at least to my ability to pattern match - this seems right.

jasonpaulos (approved):
Looks great

PR #171, opened 2022-01-27
Change Subroutine Wrapped Callable to a class with call method

ahangsu:
# Summary

Previous implementation of subroutine wraps a function implementation and return a callable object, allowing for subroutine syntax definition and subroutine call syntax. But this does not allow for much information extractable about underlying subroutine.

This PR allows for more information (name, return type, has return) about the subroutine extractable.

### DIFF
--- a/pyteal/__init__.pyi
+++ b/pyteal/__init__.pyi
@@ -114,6 +114,7 @@ __all__ = [
     "SubroutineDefinition",
     "SubroutineDeclaration",
     "SubroutineCall",
+    "SubroutineFnWrapper",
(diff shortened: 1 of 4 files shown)


jasonpaulos (commented):
Looks mostly good, but could you add a line to here testing that `mySubroutine` is now an instance of `SubroutineFnWrapper`? https://github.com/algorand/pyteal/blob/5bcee03e1c00691434ac70d185e0070d883adbf6/pyteal/ast/subroutine_test.py#L151

jasonpaulos (approved):

PR #174, opened 2022-01-28
adding new teal6 ops, no pyteal expressions defined for them yet

barnjamin:


algoanne:
for clarity, this change is adding the following new TEAL functionality into pyteal: 
- the `LastLog` transaction field
- the `divw`, `itxnas`, and `gitxnas` opcodes.

ahangsu (approved):
Both Ben's and Mike's code look good to me.

pyteal/ast/txn.py
```diff
@@ -695,6 +707,17 @@ def created_application_id(self) -> TxnExpr:
         """
         return self.makeTxnExpr(TxnField.created_application_id)
 
+    def last_log(self) -> TxnExpr:
+        """Get the last logged message from a transaction.
```
jasonpaulos, line 711:
```suggestion
        """A convenience method for getting the last logged message from a transaction.
```

pyteal/ast/txn.py
```diff
@@ -108,6 +108,7 @@ class TxnField(Enum):
     num_logs = (59, "NumLogs", TealType.uint64, False, 5)
     created_asset_id = (60, "CreatedAssetID", TealType.uint64, False, 5)
     created_application_id = (61, "CreatedApplicationID", TealType.uint64, False, 5)
+    last_log = (62, "LastLog", TealType.bytes, False, 6)
```
jasonpaulos, line 111: You could add `StateProofPK` as well, it's the last item in the table at: https://github.com/algorand/go-algorand/blob/master/data/transactions/logic/TEAL_opcodes.md#txn-f

jasonpaulos (commented):
Looks good, I just have some docs suggestions, plus I noticed `StateProofPK` is not included here.

* The methods `logs`, `created_application_id`, and `created_asset_id` currently state that they only work on inner txns, but as of TEAL 6, they also work on top-level transactions as well.
* It would be nice if you could update `docs/accessing_transaction_field.rst` with the new txn fields

pyteal/ast/txn.py
michaeldiamant: @jasonpaulos Thanks for the documentation pointer.  I added `StateProofPK` via https://github.com/algorand/pyteal/pull/174/commits/a181b737aefd63edcf821837a9b6845c991e64c2.

However, I'm mostly unfamiliar with state proof feature details.  So, I welcome feedback/pointers on additional, relevant docstring remarks.

ahangsu: If I got a vote, I think it is correct. What we are doing (PyTeal) is not quite related with stateproof, essentially we are building a correspondence to Teal code, so exposing the `StateProofPK` field (and adding corresponding test) seems right to me.

jasonpaulos (approved):
This looks good to merge. Updating the docs would be nice, but that can happen in a separate PR (there are more things to add to the docs anyway), so I wouldn't consider that a hard requirement right now

PR #176, opened 2022-02-01
Expose ABI Type names

ahangsu:
## Summary

Pretty minor PR, want to expose the names of ABI types (does not goes in TEAL level, only on PyTEAL level) for later development.


jasonpaulos (approved):
Looks good, thanks

PR #182, opened 2022-02-02
Subroutine Type Annotations

tzaffi:
Assert that `implementation` param of `SubroutineDefinition` has the expected annotation types.

The followup PR #183 expands the allowable annotations types as needed.

### DIFF
--- a/pyteal/ast/subroutine.py
+++ b/pyteal/ast/subroutine.py
@@ -1,3 +1,4 @@
+from __future__ import annotations
 from typing import Callable, List, Optional, TYPE_CHECKING
 from inspect import Parameter, signature
 
@@ -47,6 +48,14 @@ class SubroutineDefinition:
                     "Function has a parameter with a default value, which is not allowed in a subroutine: {}".format(
                         name
                     )
+                )
+
+        for var, var_type in implementation.__annotations__.items():
+            if var_type != Expr:
+                stub = "return " if var == 'Return' else ("parameter " + var)
+                    
+                raise TealInputError(
+                    "Function has {} of disallowed type {}. Only type Expr is allowed".format(stub, var_type)
                 )
 
         self.implementation = implementation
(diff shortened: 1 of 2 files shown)


pyteal/ast/subroutine_test.py
```diff
@@ -72,12 +76,21 @@ def fnWithKeywordArgs(a, *, b):
     def fnWithVariableArgs(a, *b):
         return Return()
 
+    def fnWithNonExprReturnAnnotation(a, b) -> TealType.uint64:
+        return Return()
+
```
tzaffi, line 81:
```suggestion
```

pyteal/ast/subroutine.py
```diff
@@ -49,6 +49,16 @@ def __init__(
                     )
                 )
 
+        for var, var_type in implementation.__annotations__.items():
```
ahangsu, line 52: nit: I believe this line can take advantage of previously defined `sig` on line 32.

tzaffi: Looking closer at it, I believe I could. Use the `sig` approach I would have query the parameters in a different way from the return so it's a bit more cumbersome. Do you see any clear advantage to changing the logic to use `sig` instead of `implementation.__annotations__` ?

pyteal/ast/subroutine.py
```diff
@@ -49,6 +49,16 @@ def __init__(
...
+            if var_type != Expr:
```
jasonpaulos, line 53:
```suggestion
            if var_type is not Expr:
```
Equality operators can "lie" by being overloaded, but `is` will never lie. Though I see you change this check in the follow up PR, so it's a bit of a moot point.

jasonpaulos (commented):
Looks good

jasonpaulos (approved):
I have no preference on the `sig` vs `implementation.__annotations__` issue, so I'll approve now

ahangsu (approved):
I am fine with the code rn

PR #183, opened 2022-02-02
handle ScratchVar and Expr subtypes for the return

tzaffi:


pyteal/ast/scratchvar.py
```diff
@@ -17,7 +17,15 @@ class ScratchVar:
             ])
     """
 
-    def __init__(self, type: TealType = TealType.anytype, slotId: int = None):
+    MAX_STACK_DEPTH = 700 * 256 / 2
```
tzaffi, line 20: need to remove this

pyteal/ast/scratch.py
```diff
@@ -126,7 +142,10 @@ def __str__(self):
     def __teal__(self, options: "CompileOptions"):
         from ..ir import TealOp, Op, TealBlock
 
-        op = TealOp(self, Op.store, self.slot)
+        store_op = Op.stores if self.slot.slotIdFromStack else Op.store
+        op_args = [] if self.slot.slotIdFromStack else [self.slot]
```
tzaffi, line 146: the "real" teal changes for store

pyteal/ast/scratch.py
```diff
@@ -93,7 +106,10 @@ def __str__(self):
     def __teal__(self, options: "CompileOptions"):
         from ..ir import TealOp, Op, TealBlock
 
-        op = TealOp(self, Op.load, self.slot)
+        load_op = Op.loads if self.slot.slotIdFromStack else Op.load
+        op_args = [] if self.slot.slotIdFromStack else [self.slot]
```
tzaffi, line 110: the "real" teal changes for load

pyteal/ast/subroutine_test.py
```diff
@@ -1,7 +1,11 @@
+from distutils.version import StrictVersion
```
ahangsu, line 1: nit: unused import?

pyteal/ast/scratch.py
```diff
@@ -17,26 +17,36 @@ class ScratchSlot:
...
-            self.isReservedSlot = False
+        assert (
+            requestedSlotId is None or not slotIdFromStack
+        ), "cannot specify requestedSlotId when slotIdFromStack"
+
+        self.slotIdFromStack = slotIdFromStack
+
+        if self.slotIdFromStack:
+            self.id = -1
+            self.isReservedSlot = True
```
ahangsu, line 35: nit: in this condition branch, `requestedSlotId` is unchecked, while I would love to enforce it to be `None` to make things clearer.

pyteal/ast/subroutine_test.py
tzaffi: 👍

pyteal/ast/scratch.py
tzaffi: I'll add an extra assertion. Even though in theory, you can never get to this branch when `requestedSlotId != None` given the assertion on line 28 above, explicit is better than implicit.

ahangsu: oh no, I overlooked things...

tzaffi: ok, will not add the extra check

pyteal/ast/scratchvar.py
```diff
@@ -17,7 +17,15 @@ class ScratchVar:
...
+
+    def __init__(
+        self,
+        type: TealType = TealType.anytype,
+        slotId: int = None,
+        *,
+        slodIdFromStack: bool = False,
```
jasonpaulos, line 27:
```suggestion
        slotId: Union[int, Expr] = None,
```
Based on the conversation we had, I suggest doing something like this so that you'll be able to pass in the slot to use with the dynamic version of the opcodes. I imagine this could work like this:
* If `slotId` is an int, then `store` & `load` should be used to access it like they do currently
* If `slotId` is an `Expr`, then is represents a dynamic, runtime index, so `stores` and `loads` should be used to access it.

jasonpaulos: Also I'd probably recommend making the static vs dynamic decision inside of `ScratchVar` so that `ScratchSlot` does not change too much since it's used extensively in the compiler as a way to represent a static slot. Maybe a `DynamicScratchSlot` is needed to address the dynamic case?

pyteal/ast/scratch.py
```diff
@@ -9,7 +9,58 @@
...
+        """Get an expression to store a value in this slot.
+
+        Args:
+            value (optional): The value to store in this slot. If not included, the last value on
+            the stack will be stored. NOTE: storing the last value on the stack breaks the typical
+            semantics of PyTeal, only use if you know what you're doing.
+        """
+        if value is not None:
+            return ScratchStore(self, value)
+        return ScratchStackStore(self)
```
tzaffi, line 28: Gotta figure out the impact of this `ScratchStackStore` in the case of no value provided... not sure it works out of the box with `DynamicSlot`

pyteal/ast/scratch.py
```diff
@@ -9,7 +9,58 @@
...
+            type (optional): The type being loaded from this slot, if known. Defaults to
+                TealType.anytype.
+        """
+        return ScratchLoad(self, type)
+
+
+Slot.__module__ = "pyteal"
+
+
+class DynamicSlot(Slot):
```
jasonpaulos, line 43: I like this direction, but still I don't see how `DynamicSlot` knows which slot it should store to/load from.

In my opinion, `DynamicSlot` needs a constructor that takes an `Expr` that indicates this value.

tzaffi: Will keep this in mind

tzaffi: I think this isn't needed. Imagine we are trying store the value `val: Expr`. Let's assume that `ScratchVar.slot: Union[DynamicSlot, ScratchSlot]` is set during `ScratchVar`'s construction based on `slotId`'s type and follow the code path:
```python
ScrachVar(slotId).store(val) -> dispatches ->
    (DynamicSlot|ScratchSlot).store(val) ->
        ScratchStore(val) : Expr
```
So in either case, we end up staring `val` via `ScratchStore(val)`

docs/control_structures.rst
```diff
@@ -296,17 +296,17 @@ Subroutines
 .. note::
```
tzaffi, line 296: Unrelated doc changes.

examples/application/pass_by_reference.py
```diff
@@ -0,0 +1,245 @@
+from pathlib import Path
```
tzaffi, line 1: Clean this up...

pyteal/__init__.pyi
```diff
@@ -17,154 +17,156 @@ from .errors import TealInternalError, TealTypeError, TealInputError, TealCompil
 from .config import MAX_GROUP_SIZE, NUM_SLOTS
 
 __all__ = [
-    "Expr",
```
tzaffi, line 20: sorting alphabetically

pyteal/ast/__init__.py
```diff
@@ -120,138 +120,147 @@
...
+    ScratchLoad,
+    ScratchStore,
+    ScratchStackStore,
+    Slot,
+)
 from .scratchvar import ScratchVar
 from .maybe import MaybeValue
 
 __all__ = [
-    "Expr",
```
tzaffi, line 128: alphabetizing

pyteal/ast/scratch.py
```diff
@@ -59,6 +38,11 @@ def load(self, type: TealType = TealType.anytype) -> "ScratchLoad":
         """
         return ScratchLoad(self, type)
 
+    # TODO: Can I get this to work?
```
tzaffi, line 41: don't leave as is

pyteal/ast/scratch.py
```diff
@@ -69,13 +53,98 @@ def __hash__(self):
...
+        assert isinstance(
+            slotIdExpr, Expr
+        ), "slotIdExpr must be an Expr but was provided {}".format(type(slotIdExpr))
+
+        self.id: Expr = slotIdExpr
+
+    def dynamic(self) -> bool:
+        return True
+
+    # TODO: Can I get this to work?
```
tzaffi, line 83: don't leave as is

pyteal/ast/scratch.py
```diff
@@ -69,13 +53,97 @@ def __hash__(self):
...
+        """
+        super().__init__()
+
+        assert slotIdExpr is not None, "cannot create a DynamicSlot with no slotIdExpr"
+
+        assert isinstance(
+            slotIdExpr, Expr
+        ), "slotIdExpr must be an Expr but was provided {}".format(type(slotIdExpr))
+
+        self.id: Expr = slotIdExpr
```
tzaffi, line 77: Code smell: `id` property can either be an `int` or `Expr` based on Slot subclass

pyteal/compiler/compiler.py
```diff
@@ -145,17 +145,24 @@ def compileSubroutine(
     subroutineBlocks[currentSubroutine] = start
 
     referencedSubroutines: Set[SubroutineDefinition] = set()
+    subroutineCallbacks: Dict[SubroutineDefinition, SubroutineCall] = {}
     for stmt in teal:
         for subroutine in stmt.getSubroutines():
             referencedSubroutines.add(subroutine)
+            assert isinstance(
```
tzaffi, line 152: This new assertion would break any previous code that has a misbehaving `getSubroutines()` method

pyteal/ast/subroutine.py
```diff
@@ -66,10 +78,16 @@ def __init__(
...
+        anns = self.implementation.__annotations__
+        return arg in anns and anns[arg] is ScratchVar
+
+    def getDeclaration(
+        self, callback: "SubroutineCall" = None
+    ) -> "SubroutineDeclaration":
         if self.declaration is None:
             # lazy evaluate subroutine
-            self.declaration = evaluateSubroutine(self)
+            self.declaration = evaluateSubroutine(self, callback=callback)
```
tzaffi, line 90: Code smell: lazy evaluation is no longer valid because a parameter is being introduced. In this case, I believe we're ok (when you follow the branches), but it's preferred to either:
* get rid of the laziness here and take a performance hit
* or split this into a new method that expects the parameter and isn't lazy

tzaffi: **JUST PLAIN WRONG** ...gonna fix

tzaffi:
Closing for now - too much detritus. Need fresh start with less open ended approach.

tzaffi:
Might re-base this off of #195

PR #186, opened 2022-02-03
WIP: Subroutine ABI type annotation

ahangsu:


pyteal/ast/subroutine.py
```diff
@@ -1,18 +1,25 @@
...
 from .scratchvar import ScratchVar
+from . import abi
 
 if TYPE_CHECKING:
     from ..compiler import CompileOptions
 
+# NOTE trying to do return, rather than "write to reference", experimenting here
 
 class SubroutineDefinition:
+    PARAM_ANNOTATION_TYPES = (Expr, ScratchVar, abi.Type)
```
tzaffi, line 18: kewl

ahangsu: nope, this seem to be wrong, let me explain: the check sees if an annotation type is **exactly** in `PARAM_ANNOTATION_TYPES`, but for `abi.Type`, what we are looking for is its sub-types.
Some changes need to be done here...

pyteal/ast/subroutine_test.py
```diff
@@ -40,6 +44,15 @@ def fnWithOnlyArgExprAnnotations(a: Expr, b: Expr):
     def fnWithPartialExprAnnotations(a, b: Expr) -> Expr:
         return Return()
 
+    def fnWithReturnTernaryExprAnnotations(a, b: Expr) -> Nonce:
```
tzaffi, line 47: @jasonpaulos @ahangsu 

I have a PR out that makes the following proposal - add `input_types` optional paramater to **Subroutine** decorator. Will this contradict the approach here, or can they work together? EG:

```python
@Subroutine(TealType.bytes, input_types=[TealType.bytes, TealType.uint64])
def string_mult(s: ScratchVar, n):
    i = ScratchVar(TealType.uint64)
    tmp = ScratchVar(TealType.bytes)
    start = Seq(i.store(Int(1)), tmp.store(s.load()), s.store(Bytes("")))
    step = Seq(i.store(i.load() + 1), s.store(Concat(s.load(), tmp.load())))
    return Seq(
        For(start, i.load() <= n, step),
        s.load(),
    )
```

PR #193, opened 2022-02-14
Add opcode support for itxnas and gitxnas

michaeldiamant:
Adds opcode support for `itxnas` and `gitxnas`.

Child PR to  #174.


pyteal/ast/itxn.py
```diff
@@ -120,7 +120,7 @@ def Submit(cls) -> Expr:
         :any:`InnerTxnBuilder.Begin` and :any:`InnerTxnBuilder.SetField` must be called before
         submitting an inner transaction.
 
-        This will fail fail if 16 inner transactions have already been executed, or if the
+        This will fail if 16 inner transactions have already been executed, or if the
```
barnjamin, line 123: I think this is not true anymore, In teal 6 I can submit up to 256 individual transactions

pyteal/ast/gitxn.py
```diff
@@ -46,8 +47,10 @@ def __teal__(self, options: "CompileOptions"):
 class GitxnaExpr(TxnaExpr):
     """An expression that accesses an inner transaction array field from an inner transaction in the last inner group."""
 
-    def __init__(self, txnIndex: int, field: TxnField, index: int) -> None:
-        super().__init__(Op.gitxna, None, "Gitxna", field, index)
+    def __init__(
+        self, txnIndex: Union[int, Expr], field: TxnField, index: Union[int, Expr]
+    ) -> None:
+        super().__init__(Op.gitxna, Op.gitxnas, "Gitxna", field, index)
```
ahangsu, line 53: nit: I might be wrong but according to: https://github.com/algorand/go-algorand/tree/master/data/transactions/logic#inner-transactions, it seems `txnIndex` is still an immediate value, which can only be `int`.

What makes the difference in choice of `gitxna` and `gitxnas` is where the last `index` comes from:
- `index` is a stack value (`Expr`), then you want `gitxnas`
- `index` is an int (immediate value), then you want `gitxna`

pyteal/ast/gitxn.py
```diff
@@ -57,18 +60,35 @@ def __str__(self):
...
+            op = TealOp(self, opToUse, self.txnIndex, self.field.arg_name)
+            return TealBlock.FromOp(options, op, cast(Expr, self.index))
+
+        if type(self.index) is int:
+            op = TealOp(self, opToUse, self.field.arg_name, self.index)
+            return TealBlock.FromOp(options, op, cast(Expr, self.txnIndex))
+
+        op = TealOp(self, opToUse, self.field.arg_name)
+        return TealBlock.FromOp(
+            options, op, cast(Expr, self.txnIndex), cast(Expr, self.index)
```
ahangsu, line 90: nit: if my previous argument about line 50-53 is correct, then you can just use the condition branch between line 76-82, since there's no need in distinguishing `txnIndex`'s type...

pyteal/ast/itxn.py
```diff
@@ -204,7 +204,8 @@ def SetFields(cls, fields: Dict[TxnField, Union[Expr, List[Expr]]]) -> Expr:
 InnerTxnBuilder.__module__ = "pyteal"
 
 InnerTxn: TxnObject = TxnObject(
-    TxnExprBuilder(Op.itxn, "InnerTxn"), TxnaExprBuilder(Op.itxna, None, "InnerTxna")
+    TxnExprBuilder(Op.itxn, "InnerTxn"),
+    TxnaExprBuilder(Op.itxna, Op.itxnas, "InnerTxna"),
```
ahangsu, line 208: looks correct to me

ahangsu (commented):
good job on first patch of PRs, a few things might need to be fixed. I will also look into how spec for inner group size changed...

pyteal/ast/itxn.py
michaeldiamant: @barnjamin Thanks for catching.   Can you resolve if it looks good now?

* I updated the docstring in https://github.com/algorand/pyteal/pull/193/commits/dc5289f1ea0e0368a000f775b2d230bef4379a6a.  
  * Aside:  I'm surprised docstrings _cannot_ reference variables.  I considered a larger change to make _256_ a constant and reference it with one of these methods:  https://stackoverflow.com/questions/10307696/how-to-put-a-variable-into-python-docstring.  Since there's only 1 reference, I opted to for the quick change.  Though I might circle back in a subsequent PR.
* I opened https://github.com/algorand/docs/pull/743 because I noticed a stale reference there too.

pyteal/ast/gitxn.py
michaeldiamant: @ahangsu Thanks for catching and for the context.  

You're definitely right.  I misunderstood the change's implications.  Via https://github.com/algorand/pyteal/pull/193/commits/1a78b6beac9a70da56e16dbf4d5cc96c6d7e5931, I attempted to fix the mistake.

ahangsu: Nice, let me take a pass again :-)

pyteal/ast/gitxn.py
ahangsu: The line 75 is not necessary, since we no longer need to discuss `self.txnIndex`'s type anymore.

michaeldiamant: @ahangsu Agree - I tried to remove the obsolete logic via https://github.com/algorand/pyteal/pull/193/commits/1e94439f30c8aae8152135cf45cd19a85bbca3a3.  Let me know if it looks good to you.

ahangsu: looks correct, this aligns with my thoughts.

pyteal/ast/gitxn.py
```diff
@@ -57,18 +58,23 @@ def __str__(self):
 
     def __teal__(self, options: "CompileOptions"):
         verifyFieldVersion(self.field.arg_name, self.field.min_version, options.version)
-        if type(self.txnIndex) is not int or type(self.index) is not int:
```
jasonpaulos, line 60: This should still raise an error if `self.txnIndex` is an int, since we don't have a version of `gitxn` that takes transaction indexes from the stack yet

jasonpaulos (commented):
Additionally, line 101 of `pyteal/ast/gitxn.py` no longer needs to cast the index to an int, since `GitxnaExpr` can handle `Expr`s.

```python
        return TxnObject(
            lambda field: GitxnExpr(txnIndex, field),
            lambda field, index: GitxnaExpr(txnIndex, field, cast(int, index)),
        )
```
->
```python
        return TxnObject(
            lambda field: GitxnExpr(txnIndex, field),
            lambda field, index: GitxnaExpr(txnIndex, field, index),
        )
```

pyteal/ast/gitxn.py
michaeldiamant: @jasonpaulos Thanks for catching the runtime type check mistake and the unnecessary cast.  
* I attempted to fix both issues via https://github.com/algorand/pyteal/pull/193/commits/57725e0812577bbbfb4720c3271617f3ea6700dc.
* Pointers appreciated if I'm doing something in a _non-Pythonic_ way.  I'm trying to balance researching the language while getting something to work.

Aside:   What conventions (if any) exist for unit tests confirming an exception does _not_ happen?
* I ran out of time to look closely today.  A quick google suggests writing a test _without_ assertions like https://stackoverflow.com/a/20275035.
* `gitxn_test.py` seems to apply the _test without assertion_ pattern, so I extended it.  Though it's unclear to me if it's preferable.

pyteal/ast/gitxn_test.py
```diff
@@ -21,10 +21,21 @@ def test_gitxn_invalid():
...
+    with pytest.raises(TealInputError):
+        GitxnaExpr("Invalid_type", TxnField.application_args, 1).__teal__(teal6Options)
+
+    with pytest.raises(TealInputError):
+        GitxnaExpr(0, TxnField.application_args, "Invalid_type").__teal__(teal6Options)
+
+
+def test_gitxna_valid():
+    GitxnaExpr(0, TxnField.application_args, 1).__teal__(teal6Options)
+    GitxnaExpr(0, TxnField.application_args, Assert(Int(1))).__teal__(teal6Options)
```
jasonpaulos, line 38: This exposes an invalid case we aren't currently checking for any txna opcodes currently. It _should_ be invalid to use an `Expr` that evaluates to anything but a `TealType.uint64` for the index here. `Assert(Int(1))` returns an Expr with the type `TealType.none`, so we should error.

Ideally `TxnaExpr.__init__` would check this with `require_type(index, TealType.uint64)` when `index` is not a Python int. And I suppose the same problem exists for `GtxnExpr` and `GtxnaExpr` with respect to `txnIndex`, so they should have a similar check as well.

(The reason for erroring in the `__init__` method is so that a reasonable stack trace is generated which will likely contain code that the user explicitly called to create their program. If we fail later, the stack trace will come from the `compileTeal` method, which is less useful.)

We could also modify `pyteal/ast/txn_test.py` to make sure an error is raised if a non-`TealType.uint64` expression is given as an index.

@michaeldiamant if you feel confident enough to make those changes in this PR, feel free to do so. If not, it can be addressed in a different PR.

pyteal/ast/gitxn.py
jasonpaulos: 57725e0812577bbbfb4720c3271617f3ea6700dc looks good, thanks for addressing my comments. And yes, we tend to use the test with assertions approach if the only goal is to make sure something does not raise an error.

Your unit test made me realized there's a hole in our type checking, which I elaborated more on in another comment.

pyteal/ast/gitxn_test.py
michaeldiamant: @jasonpaulos Feedback makes sense to me.  I had wondered _why_ `Expr` was applied so loosely.  

Since the changes felt large enough to me, I opened a child PR to isolate my attempt to address the feedback:  https://github.com/michaeldiamant/pyteal/pull/1. 

If you feel the changes to test setup are counterproductive, let me know and I'll back out stylistic changes.

michaeldiamant: Resolving because the child PR (https://github.com/michaeldiamant/pyteal/pull/1) is merged into this PR.

jasonpaulos (approved):

pyteal/ast/txn.py
```diff
@@ -161,6 +161,15 @@ def type_of(self):
...
 
+    @staticmethod
+    def __validate_index_or_throw(index: Union[int, Expr]):
+        if not isinstance(index, (int, Expr)):
+            raise TealInputError(
+                f"Invalid index type:  Expected int or Expr, but received {index}."
+            )
+        if isinstance(index, Expr):
+            require_type(index, TealType.uint64)
+
```
ahangsu, line 172: nit: I believe this function has a same implementation in `gtxn.py` (which is written by you), right? I believe this can be reused somehow?

pyteal/ast/gtxn_test.py
```diff
@@ -2,19 +2,57 @@
...
+        with pytest.raises(e):
+            f()
+
+
+def test_gtxn_expr_invalid():
+    for f, e in [
+        (lambda: GtxnExpr(Assert(Int(1)), TxnField.sender), TealTypeError),
+    ]:
+        with pytest.raises(e):
+            f()
```
ahangsu, line 24: this smells functional, I like it.

ahangsu (commented):
looks largely good to me, just one small thing here.

pyteal/ast/txn.py
michaeldiamant: @ahangsu I agree the method can be reworked to abstract over the field (`index` vs `txn_index`).  Though, I don't know where it's appropriate for a common definition to live.  Do you have a suggestion?

After considering a few options, I don't feel strongly that the duplication ought to be addressed now.  But, I'm on-board to make a change if you have a low effort suggestion that maintains repo organizational norms.

The only _plausible_ suggestion I came away with is:  
* Define static methods in `LeafExpr` to support validation for select fields (`index` and `txn_index` to start).
* It's unclear to me if adding validation methods to `LeafExpr` fits with the existing repo organization.

I also considered:
* In my mind, a common definition ought to live with the definition of the type (`Union[int, Expr]`).  Since a custom type doesn't exist, there isn't an associated place to define the validation.
* Alternatively, I can imagine defining a more generic validation object with an arbitrary `isinstance` check for all members of a `Union` data type.
  * I don't know enough about Python to know how to write the check.  It feels to me a bit over-the-top for the problem at-hand.

ahangsu: hmmm... I don't think defining static method in `LeafExpr` is a good option, since some other stuffs might inherit from `LeafExpr`, which "contaminates" the basic definition.

An idea I am thinking of:, extending `validate_txn_index_or_throw` to allow optional default argument, i.e., of form
```python
def validate_txn_index_or_throw(txnIndex: Union[int, Expr], fieldName: str = "txn_index"):
```
and probably rename the function to a most general `validate_index_or_throw`. This gives a low effort modification, and saves some lines of somewhat duplicated definition.

As per the "more generic validation object", I agree the idea is appealing, but it might take too much effort to get into this. For the spirit of this PR, I don't feel strong about that...

ahangsu: Yea I am convinced that it might causing more obfuscation than simplicity here, providing the static method here is fine for me.

ahangsu (approved):
LGTM

PR #194, opened 2022-02-14
ABI Bytes from ABI Uint8

ahangsu:
Minor improvement from ABI type implementation for `abi.Byte`:
- Changed `__str__` for `abi.Byte`
- Allowing `abi.Byte` and `abi.Uint8` to set from: `int` or `Expr` or `abi.Uint8` or `abi.Expr`.
- Tested.

The reason for putting `Byte` in `uint.py` is that, python would complain circular import in this case...


jasonpaulos (approved):
Looks good!

PR #196, opened 2022-02-14
MultiValue expression implemented to support opcodes that return multiple values

algoidurovic:
MultiValue is intended to supersede the older implementation of MaybeValue. While MaybeValue will continue to be available, it will inherit from the parent MultiValue class and include the option for the user to specify an output reducer function. If the reducer function isn't provided then the behavior of MaybeValue will be the same as before and won't require users to change existing PyTeal code.

Testing:
Unit tests were used to test MultiValue/MaybeValue behavior w/ and w/o user specific user functions. The unit tests cover all currently supported opcodes that return multiple values. When new opcodes are introduced in the future that don't follow the MaybeValue pattern, these tests will need to be extended and generalized.


pyteal/ast/maybe.py
```diff
@@ -29,62 +30,23 @@ def __init__(
...
+        types = [TealType.uint64, type]
+        super().__init__(op, types, immediate_args=immediate_args, args=args)
 
     def hasValue(self) -> ScratchLoad:
         """Check if the value exists.
 
         This will return 1 if the value exists, otherwise 0.
         """
-        return self.slotOk.load(TealType.uint64)
+        return self.output_slots[0].load(self.types[0])
```
jasonpaulos, line 41: If you wanted to be super backwards compatible, you could use the `@property` decorator to create getter functions for `slotOk` and `slotValue` that return the appropriate value from `self.output_slots`.

It shouldn't be strictly necessary since those fields were not meant to be used by consumers (though I have written some hacky code that does access the slots directly), but there's no downside to it.

pyteal/ast/multi.py
```diff
@@ -0,0 +1,85 @@
...
+            store = slot.store()
+            storeStart, storeEnd = store.__teal__(options)
+            curEnd.setNextBlock(storeStart)
+            curEnd = storeEnd
+
+        if self.reducer_set:
+            input = [
+                slot.load(self.types[i]) for i, slot in enumerate(self.output_slots)
+            ]
+            reducerStart, reducerEnd = self.reducer(*input).__teal__(options)
```
jasonpaulos, line 75: I'd like to suggest you flip the order of the `output_slots` when constructing `input`. The reason being, I think it's most natural for the first argument that the reducer sees to be the first value the opcode pushes to the stack, not the last.

Right now, if an opcode adds `A, B, C` to the stack, the reducer would receive the arguments `C, B, A`.

For the existing `MaybeValue` opcodes, flipping doesn't make a huge difference, but for more advanced opcodes like `divmodw` and `ecdsa_pk_decompress`, it gives their values a more logical order.

pyteal/ast/multi.py
```diff
@@ -0,0 +1,85 @@
...
+        self.immediate_args = immediate_args if immediate_args is not None else []
+        self.args = args if args is not None else []
+
+        self.output_slots = [ScratchSlot() for _ in self.types]
+        self.reducer_set = False
+
+    def outputReducer(self, reducer: Callable[..., Expr]) -> Expr:
+        self.reducer = reducer
+        self.reducer_set = True
+        return self
```
jasonpaulos, line 44: Have you considered an alternative implementation like this?

```python
def outputReducer(self, reducer: Callable[..., Expr]) -> Expr:
    # taken from __teal__
    input = [
        slot.load(self.types[i]) for i, slot in enumerate(self.output_slots)
    ]
    return Seq(self, reducer(*input))
```

It seems possibly simpler and more pure, since calling `outputReducer` does not modify the `MultiValue` object, plus the `MultiValue` is no longer responsible for converting the reducer function output to the IR. But I have not thought through all the implications.

jasonpaulos (commented):
Both the implementation and tests look great. I like how nicely this integrates with the existing `MaybeValue`.

I left a few suggestions, the most important of which is the one about slot ordering.

pyteal/ast/maybe.py
```diff
@@ -1,16 +1,17 @@
...
-from .leafexpr import LeafExpr
-from .scratch import ScratchSlot, ScratchLoad
+from .scratch import ScratchLoad
 
 if TYPE_CHECKING:
     from ..compiler import CompileOptions
 
 
-class MaybeValue(LeafExpr):
+class MaybeValue(MultiValue):
```
michaeldiamant, line 14: @algoidurovic and reviewers - Trying to understand the thinking:  Do we envision removing `MaybeValue`?

As I understand it, PyTeal retains `MaybeValue` for backwards compatibility.  If that's the case, I'm trying to gauge deprecation policy options.

jasonpaulos: I'd be in favor of deprecating `MaybeValue.hasValue` and `.value` to encourage new users to use the reducer function.

While we have no strict requirement to maintain backwards compatibility between most releases (we are in major version 0), I think we should make a best effort attempt to do so.

pyteal/ast/multi.py
```diff
@@ -0,0 +1,85 @@
...
+        super().__init__()
+        self.op = op
+        self.types = types
+        self.immediate_args = immediate_args if immediate_args is not None else []
+        self.args = args if args is not None else []
+
+        self.output_slots = [ScratchSlot() for _ in self.types]
+        self.reducer_set = False
+
+    def outputReducer(self, reducer: Callable[..., Expr]) -> Expr:
```
michaeldiamant, line 41: @algoidurovic For my understanding:  What is the tradeoff between exposing _reducer_ as a constructor argument vs a separate method (like the PR)?

I'm trying to gauge if/how requiring the _reducer_ at construction complicates the picture.  Probably because my understanding is lacking, I am wondering _when_ the caller benefits from the separate method to customize the _reducer_.

michaeldiamant: Aside:  I see the test cases reuse a `MultiValue` instance and change the _reducer_.  It's unclear to me if we'd expect regular re-use in production applications.

pyteal/ast/maybe.py
michaeldiamant: @jasonpaulos Thanks for the reply.  A few follow up questions:
* If I follow the reply, you're imagining retaining `MaybeValue` without implementation-specific methods.  If I've followed, in what use cases do you envision favoring `MaybeValue` over `MultiValue`?
  * I'm wondering if there's specific use cases you see.  Or if you generally imagine an implementation specific to a single value is helpful.  
  * I'll likely defer to your judgement though I'm wondering about the maintenance cost outweighing the benefits.
 * For any deprecation we undertake, I'm wondering _how_ we enforce the deprecation?
   * A quick google shows a library like https://pypi.org/project/deprecation/.  While that library may not be the way forward, I'm gauging _how_ and _when_ we'd enforce removal of deprecated functionality.

algoidurovic: My understanding is that we'll keep MaybeValue for the currently supported opcodes (the ticket associated with this PR gives some examples). We can even use MaybeValue for future opcodes if the returned values fit the pattern. Because it now extends the MultiValue class, its features should match MultiValue's so it's not behind feature-wise. This PR changes MaybeValue into a lightweight class that basically just adds a thin layer of specificity on top of MultiValue.

pyteal/ast/multi.py
algoidurovic: 1. I think it makes sense stylistically to "chain" a transformation function/reducer rather than pass it to the constructor. It meshes better with PyTeal code style imo.
2. The user doesn't usually construct the MultiValue/MaybeValue directly. A higher level function (for example, AccountParams.balance(..)) encapsulates the construction of the MaybeValue so the user couldn't pass in a reducer.

I wouldn't expect the reducer to be redefined in production apps very often but that option is available. It could be useful if you want to apply more than one transformation to the outputs in a chain or separately. As you pointed out, it's nice for testing as well.

pyteal/ast/multi.py
algoidurovic: I agree, this approach is better.

pyteal/ast/multi.py
michaeldiamant: @algoidurovic Got it - thanks for the context.  All set on my end - resolving.

pyteal/ast/maybe.py
michaeldiamant: @algoidurovic Got it - thanks for the context.  Let's keep as is then - resolving.

pyteal/ast/multi.py
algoidurovic: One issue I thought of. If someone does decide to call outputReducer() on MaybeValue output then the result won't be a MaybeValue, so hasValue() and value() won't be able to be called. I don't think this is a problem because if outputReducer() is called then it's because the user wants to do something with the return values different from the default. Another consideration is that reducers won't be able to be chained but we never intended for that to be supported anyway so imo not an issue. Any thoughts?

jasonpaulos (approved):
Changes look good, I just have one test suggestion.

Once that's added, this looks good to merge

pyteal/ast/multi.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def __teal__(self, options: "CompileOptions"):
+        tealOp = TealOp(self, self.op, *self.immediate_args)
+        callStart, callEnd = TealBlock.FromOp(options, tealOp, *self.args)
+
+        curEnd = callEnd
+        # the list is reversed in order to preserve the ordering of the opcode's returned
+        # values. ie the output to stack [A, B, C] should correspond to C->output_slots[2]
+        # B->output_slots[1], and A->output_slots[0].
+        for slot in reversed(self.output_slots):
```
PabloLION, line 67: I would like to ask if there's possibility to improve the performance of these lines.

The process of "reading an global state `EX_APP_KEY` of app `ID` then assert" in TEAL was like this:

```teal
int ID
byte "EX_APP_KEY"
app_global_get_ex
assert
```

But in PyTeal, with this MultiValue, the generated code becomes:

```teal
int ID
byte "EX_APP_KEY"
app_global_get_ex
store 1
store 0
load 1
assert
load 0
```

The OpCost of this process was doubled from 4 to 8 (Or 2 to 6, if the two arguments are already on the stack). In other words, the performance was reduced by half or a third. 



I want to [PR](https://github.com/algorand/pyteal/pull/541) for `addw` and `mulw`. But using this seems would change the OpCost from 1 (in TEAL) of `addw` to 5 (in PyTeal) of these following lines:

```teal
addw
store 0
store 1
load 1
load 0
```

This would be a huge limitation when we trying to use this multiple times. Can I get some advice / work-around on this?
Thanks in advance.

PabloLION: Using scratch slots seems very reasonable to me. I'm not trying to remove it, just to improve the OpCost. I hope ideas 
 to be helpful:

1. Add a method like `pyteal.TealBlock.NormalizeBlocks` to minimize the generated teal code like `store X; store Y; load Y; load X;`
2. Add a functionality like [yesql](https://www.npmjs.com/package/yesql) allowing users to write subroutines with teal code (with allowed op codes from pyteal.Op) directly.

michaeldiamant: @PabloLION Let me know if the note helps to address your question + resolves the thread?

* Double checking:  Have you seen https://pyteal.readthedocs.io/en/stable/compiler_optimization.html?  We've considered extending the capability though it's not at the top of our priority list.  
* I _think_ https://github.com/algorand/pyteal-utils#inline-assembly fulfills the desire for access to write TEAL source.

PabloLION: I didn't know about the "Compiler Optimization", and the InlineAssembly is amazing! Thanks for the clarification🥰

PR #198, opened 2022-02-15
Pass-by-Ref / Dynamic Scratch Variables via the `loads` and `stores` opcodes

tzaffi:
# Dynamic Scratch Variables

## New Capabilities
This PR aims to allow dynamically-indexed scratch variables. For example here are some nice-to-have examples:

### Pass-by-Reference Semantics
You can see the resulting teal (including my comments) [here](https://github.com/algorand/pyteal/blob/f67947a0916357766c424db1dbe6c64d4b3983e9/tests/teal/swapper_expected.teal).
```python
@Subroutine(TealType.none)
def swap(x: ScratchVar, y: ScratchVar):
    z = ScratchVar(TealType.anytype)
    return Seq(
        z.store(x.load()),
        x.store(y.load()),
        y.store(z.load()),
    )
```

### Use a Dynamic ScratchVar to "iterate" over other ScratchVar's

You can see the resulting [teal with comments](https://github.com/algorand/pyteal/blob/f67947a0916357766c424db1dbe6c64d4b3983e9/tests/teal/wilt_the_stilt_expected.teal) here.
```python
def wilt_the_stilt():
    player_score = DynamicScratchVar(TealType.uint64)

    wilt = ScratchVar(TealType.uint64, 129)
    kobe = ScratchVar(TealType.uint64)
    dt = ScratchVar(TealType.uint64, 131)

    return Seq(
        player_score.set_index(wilt),
        player_score.store(Int(100)),
        player_score.set_index(kobe),
        player_score.store(Int(81)),
        player_score.set_index(dt),
        player_score.store(Int(73)),
        Assert(player_score.load() == Int(73)),
        Assert(player_score.index() == Int(131)),
        player_score.set_index(wilt),
        Assert(player_score.load() == Int(100)),
        Assert(player_score.index() == Int(129)),
        Int(100),
    )
```

## Other improvements
I've also introduced another way to run end-to-end tests. The script `tests/pass_by_ref_test.py` works as follows:

* iterates over a list of PyTEAL approval functions defined in the file
* compiles the program and saves into `tests/teal` with a file name that is dynamically generated from the approval program function name
* loads pre-compiled companion TEAL with an extra `_expected` suffix in the filename
* compares the just-compiled TEAL against the expected TEAL line by line. The expected TEAL can have extra information -such as comments- but they have to start out exactly the same for the test to pass

## Testing
* Unit tests have been added to:
  * `pyteal/ast/scratch_test.py`
  * `pyteal/ast/scratchvar_test.py`
  * `pyteal/ast/subroutine_test.py`
* The end to end test `tests/pass_by_ref_test.py` has been added
* Temporary PR #203 runs all the backwards compatible tests in `tests/pass_by_ref_test.py` to ensure that the generated TEAL hasn't changes due to this PR

## Summary of Changes
* Various `__init__*`'s
  * To bubble up the visibility of the new `DynamicScratchVar` and make **mypy** happy
  * Also, modified `scripts/generate_init.py` now sorts and uniquifies `__all__`
* Resolve `mypy` complaints:
  * `pyteal/ast/multi_test.py`
* pyteal/ast/scratch.py: 
  * ability to query a slot's index at runtime via the new `ScratchIndex` class and various `index()` methods
* `pyteal/ast/scratchvar.py` - new `DynamicScratchIndex` class that allows subroutines to access and modify external `ScratchVar`'s and also allows it to act as an "iterator" over pre-existing `ScratchVar`'s. It uses the `loads` and `stores` op codes in order to achieve its capabilities.
* `pyteal/ast/subroutine.py` - modify the to allow pass-by-reference semantics
  * `SubroutineDefinition` now:
    * has a `by_ref_args` member set so one can tell if an argument is meant to be by-reference or by-value
    * allows `return` types to be any subtype of `Expr`, not just `Expr`
    * allows argument types to be `ScratchVar`'s in addition to the previous `Expr`
    * has a new `arguments()` method
  * `SubroutineCall` now can handle by-ref arguments, by detecting such arguments and calling their new `.index()` method to pass their index on the stack, instead of their value
  * `SubroutineDeclaration` now can handle by-ref arguments as well. **THIS IS THE TRICKIEST PART OF THE PR**. The code changes are deceptively small. There are some new comments which attempt to explain the goings on.
* New end-to-end test `tests/pass_by_ref_test.py`
  * Various expected outputs of compilation `tests/teal/*_expected.py`


pyteal/__init__.pyi
```diff
@@ -17,154 +17,156 @@ from .errors import TealInternalError, TealTypeError, TealInputError, TealCompil
 from .config import MAX_GROUP_SIZE, NUM_SLOTS
 
 __all__ = [
-    "Expr",
```
tzaffi, line 20: alphabetize and add `DynamicScratchVar`

pyteal/ast/scratch.py
```diff
@@ -59,40 +59,89 @@ def load(self, type: TealType = TealType.anytype) -> "ScratchLoad":
...
 
 class ScratchLoad(Expr):
     """Expression to load a value from scratch space."""
 
-    def __init__(self, slot: ScratchSlot, type: TealType = TealType.anytype):
+    def __init__(
+        self,
+        slot: Optional[ScratchSlot],
+        type: TealType = TealType.anytype,
+        index_expression: Expr = None,
```
tzaffi, line 109: Allowing the creation of `ScratchLoad` (and similarly for `ScratchStore`) by providing an expression for the index _instead_ of the currently required `slot`

pyteal/ast/subroutine.py
```diff
@@ -136,28 +151,55 @@ def has_return(self):
...
+    ) -> None:
         super().__init__()
         self.subroutine = subroutine
         self.args = args
 
         for i, arg in enumerate(args):
-            if arg.type_of() == TealType.none:
+            arg_type = None
+
+            if not isinstance(arg, (Expr, ScratchVar)):
```
jasonpaulos, line 164: I don't see any checks to make sure the arguments match the specific subroutine definition. Meaning if the subroutine has args `(ScratchVar, ScratchVar)`, you could currently pass in `(Expr, Expr)` without any errors.

jasonpaulos (commented):
I've made a single pass and things look pretty good, though I need to take a closer look at the new testing infrastructure and `evaluateSubroutine`

pyteal/ast/subroutine.py
tzaffi: 👍

pyteal/ast/gtxn_test.py
```diff
@@ -2,6 +2,8 @@
 
 from .. import *
 
+from .bytes import Bytes
```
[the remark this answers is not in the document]
tzaffi: don't know why mypy was complaining. I'll have another look

pyteal/ast/__init__.py
```diff
@@ -120,141 +120,193 @@
...
+    "TxnGroup",
+    "TxnObject",
+    "TxnType",
+    "UnaryExpr",
     "While",
-    "For",
-    "Break",
-    "Continue",
+    "WideRatio",
+    "WideRatio",
```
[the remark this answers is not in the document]
tzaffi: fixed

pyteal/ast/scratch.py
```diff
@@ -59,40 +59,89 @@ def load(self, type: TealType = TealType.anytype) -> "ScratchLoad":
...
     def __teal__(self, options: "CompileOptions"):
         from ..ir import TealOp, Op, TealBlock
 
+        if self.index_expression is not None:
+            op = TealOp(self, Op.loads)
+            return TealBlock.FromOp(options, op, self.index_expression)
+
+        if not isinstance(self.slot, ScratchSlot):
+            raise TealInternalError(
+                "cannot handle slot of type {}".format(type(self.slot))
```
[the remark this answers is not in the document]
tzaffi: This is a second check. See line 123 above

pyteal/ast/gtxn_test.py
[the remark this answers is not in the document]
tzaffi: still have the same problem

pyteal/ast/subroutine.py
```diff
@@ -247,10 +301,60 @@ def __call__(self, fnImplementation: Callable[..., Expr]) -> SubroutineFnWrapper
...
+
+    def var_n_loaded(param):
+        if param in subroutine.by_ref_args:
+            argVar = ScratchVar(TealType.uint64)
+            loaded = DynamicScratchVar(TealType.anytype, indexer=argVar)
+        else:
+            argVar = ScratchVar(TealType.anytype)
+            loaded = argVar.load()
+
+        return argVar, loaded
```
jasonpaulos, line 352: I'd prefer if we could avoid having the `indexer` parameter. I think it should be possible if you do something like this instead:

```python
    def var_n_loaded(param):
        if param in subroutine.by_ref_args:
            argVar = DynamicScratchVar(TealType.anytype)
            loaded = argVar
        else:
            argVar = ScratchVar(TealType.anytype)
            loaded = argVar.load()

        return argVar.slot, loaded

argSlots, loadedArgs = zip(...)
...
bodyOps = [slot.store() for slot in argSlots[::-1]]
```

pyteal/ast/scratchvar.py
```diff
@@ -47,5 +54,110 @@ def load(self) -> ScratchLoad:
...
         return self.slot.load(self.type)
 
+    def index(self) -> Expr:
+        return self.slot.index()
+
 
 ScratchVar.__module__ = "pyteal"
+
+
+class DynamicScratchVar:
```
jasonpaulos, line 64: If you get rid of the `indexer` parameter, then I think `DynamicScratchVar` could inherit from `ScratchVar`.

This would be preferable because right now, the subroutine invocation code is "lying" to user-defined subroutine bodies, since they expect to be called with an instance of `ScratchVar` but are actually given a `DynamicScratchVar`. Making `DynamicScratchVar` a subclass of `ScratchVar` would fix this oddity.

pyteal/ast/subroutine.py
```diff
@@ -48,16 +55,29 @@ def __init__(
                         name
                     )
                 )
+            if name in implementation.__annotations__:
+                self.expected_arg_types.append(ScratchVar)
```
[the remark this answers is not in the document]
tzaffi: good catch

pyteal/ast/subroutine.py
```diff
@@ -78,21 +100,26 @@ def name(self) -> str:
...
 
-        for i, arg in enumerate(args):
-            if not isinstance(arg, Expr):
-                raise TealInputError(
-                    "Argument at index {} of subroutine call is not a PyTeal expression: {}".format(
-                        i, arg
+        if self.compiler_version is not None and self.compiler_version >= 6:
+            for i, arg in enumerate(args):
+                atype = self.expected_arg_types[i]
+                if not isinstance(arg, atype):
```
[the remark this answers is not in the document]
tzaffi: was failing some older compiler tests but maybe it's more because of the issue you joined pointed out... in general, if people annotated their subroutines in the past our new annotation restrictions break their old code.

pyteal/ast/scratchvar.py
tzaffi: I'm going to push back a bit on this one, but I'm not dead-set against. I think the main issue is that I want to exclude using DynamicScratchVar's every where ScratchVar's are used. The best example is the one you just mentioned whereby I'm "lying". If a DynamicScratchVar is a ScratchVar, then it seems reasaonable to expect the users to write the subroutines with `: DynamicScratchVar`'s annotations instead. This is a can of worms I didn't want to pursue.

jasonpaulos: We should be as truthful as possible with our type interfaces. If we say your subroutine will receive a `ScratchVar`, you should expect `isinstance(arg, ScratchVar)` to be true.

As for allowing `DynamicScratchVar` as a valid annotation on subroutines, I don't think we should allow this right now because I don't see any reason to do it. As far as pass-by-reference is concerned, `DynamicScratchVar` is just an implementation detail that's required to give subroutines a `ScratchVar`.

tzaffi: As `DynamicScratchVar`'s are also intended for use directly (e.g. the Wilt Chamberlain snippet above) we can expect this to be used for more than just an implementation detail. I'll experiment with sub-classing and assess the impact.

pyteal/ast/subroutine.py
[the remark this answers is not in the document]
tzaffi: I've removed all the compiler version gating

pyteal/ast/scratchvar.py
```diff
@@ -47,5 +54,79 @@ def load(self) -> ScratchLoad:
...
         return self.slot.load(self.type)
 
+    def index(self) -> Expr:
+        return self.slot.index()
+
 
 ScratchVar.__module__ = "pyteal"
+
+
+class DynamicScratchVar(ScratchVar):
```
tzaffi, line 64: `DynamicCratchVar` is now a `ScratchVar`

pyteal/ast/scratchvar.py
```diff
@@ -47,5 +54,86 @@ def load(self) -> ScratchLoad:
...
+        self.type = ttype
+        self._indexer = ScratchVar(TealType.uint64)
+        self.slot = self._indexer.slot
+
+    def set_index(self, index_var: ScratchVar) -> Expr:
+        """Set this DynamicScratchVar to reference the provided `indexVar`.
+        Followup `store`, `load` and `index` operations will use the provided `indexVar` until
+        `set_index()` is called again to reset the referenced ScratchVar.
+        """
+        if type(index_var) is not ScratchVar:
```
tzaffi, line 106: new assertion

pyteal/ast/subroutine_test.py
```diff
@@ -78,6 +82,100 @@ def fnWithPartialExprAnnotations(a, b: Expr) -> Expr:
         assert invocation.args == args
 
 
+def test_subroutine_invocation_param_types():
+    def fnWithNoAnnotations(a, b):
```
tzaffi, line 86: more fine grained invocation tests.

pyteal/ast/subroutine_test.py
```diff
@@ -103,27 +204,31 @@ def fnWithNonExprParamAnnotation(a, b: TealType.uint64):
...
     )
 
-    for case, msg in cases:
+    for fn, msg in cases:
         with pytest.raises(TealInputError) as e:
-            SubroutineDefinition(case, TealType.none)
+            SubroutineDefinition(fn, TealType.none)
 
-            assert msg in str(e), "failed for case [{}]".format(case)
+        assert msg in str(e), "failed for case [{}]".format(fn.__name__)
```
tzaffi, line 231: oops, test wasn't actually asserting anything specific previously

ahangsu (commented):
Some comments for now on `DynamicScratchVar`, need to think through on `Subroutine` later

pyteal/ast/scratchvar.py
```diff
@@ -47,5 +54,86 @@ def load(self) -> ScratchLoad:
...
+        Args:
+            ttype (optional): The type that this variable can hold. Defaults to TealType.anytype.
+        """
+        self.type = ttype
+        self._indexer = ScratchVar(TealType.uint64)
+        self.slot = self._indexer.slot
+
+    def set_index(self, index_var: ScratchVar) -> Expr:
+        """Set this DynamicScratchVar to reference the provided `indexVar`.
+        Followup `store`, `load` and `index` operations will use the provided `indexVar` until
```
michaeldiamant, line 103: @tzaffi Minor:  I think there's a mismatch when referring to the method parameters.

```suggestion
        """Set this DynamicScratchVar to reference the provided `index_var`.
        Followup `store`, `load` and `index` operations will use the provided `index_var` until
```

pyteal/ast/scratch.py
```diff
@@ -156,7 +226,7 @@ def __init__(self, slot: ScratchSlot):
         self.slot = slot
 
     def __str__(self):
-        return "(StackStore {})".format(self.slot)
+        return "(ScratchStackStore {})".format(self.slot)
```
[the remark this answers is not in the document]
jasonpaulos: `ScratchStackStore` is unique and potentially destructive, so it should be easy to tell it apart from a normal store in my opinion. If it where merged with `ScratchStore`, it might be too easy to accidentally enable it when you didn't mean to.

(comment about the actual line change): If the other `__str__` methods are `Store` and `Load`, I think this should remain `StackStore`

pyteal/ast/scratch_test.py
```diff
@@ -43,6 +43,19 @@ def test_scratch_load_default():
     assert actual == expected
 
 
+def test_scratch_load_index_expression():
+    expr = ScratchLoad(slot=None, index_expression=Int(1337))
+    assert expr.type_of() == TealType.anytype
+
+    expected = TealSimpleBlock([TealOp(Int(1337), Op.int, 1337)])
```
jasonpaulos, line 50: nit: if using `TealComponent.Context.ignoreExprEquality()`, use `None` for the first argument to `TealOp` for clarity

pyteal/ast/scratch_test.py
```diff
@@ -76,6 +89,28 @@ def test_scratch_store():
...
+    for value in (
+        Int(1),
+        Bytes("test"),
+        App.globalGet(Bytes("key")),
+        If(Int(1), Int(2), Int(3)),
+    ):
+        expr = ScratchStore(slot=None, value=value, index_expression=Int(1337))
+        assert expr.type_of() == TealType.none
+
+        expected = TealSimpleBlock([TealOp(Int(1337), Op.int, 1337)])
```
jasonpaulos, line 102: same thing here

pyteal/ast/scratch_test.py
```diff
@@ -76,6 +89,28 @@ def test_scratch_store():
...
+        valueStart, valueEnd = value.__teal__(options)
+        expected.nextBlock = valueStart
```
jasonpaulos, line 104: prefer using `setNextBlock`

pyteal/ast/scratchvar.py
```diff
@@ -1,6 +1,12 @@
+from operator import index
+from typing import cast
+
+from ..errors import TealInputError, TealInternalError
 from ..types import TealType, require_type
+
 from .expr import Expr
-from .scratch import ScratchSlot, ScratchLoad
+from .int import Int
```
jasonpaulos, line 8: `index`, `cast`, `prefer using setNextBlock`, `Int` aren't used

pyteal/ast/scratchvar_test.py
```diff
@@ -104,3 +117,101 @@ def test_scratchvar_assign_load():
...
+    assert myvar_int.store(Int(1)).type_of() == TealType.none
+    assert myvar_int.load().type_of() == TealType.uint64
+
+    with pytest.raises(TealTypeError):
+        myvar_int.store(Bytes("value"))
+
+    with pytest.raises(TealTypeError):
+        myvar_int.store(Pop(Int(1)))
+
+    myvar_bytes = ScratchVar(TealType.bytes)
```
jasonpaulos, line 142: `myvar_int` and `myvar_bytes` aren't `DynamicScratchVar`s

pyteal/ast/scratchvar_test.py
```diff
@@ -104,3 +117,101 @@ def test_scratchvar_assign_load():
...
+    actual, _ = expr.__teal__(options)
+    actual.addIncoming()
+    actual = TealBlock.NormalizeBlocks(actual)
+
+    assert actual == expected
+
+
+def test_dynamic_scratchvar_cannot_set_index_to_another_dynamic():
+    myvar = DynamicScratchVar()
+    sl1 = myvar.load()
```
jasonpaulos, line 209: nit: `sl1` isn't used

tests/teal/fac_by_ref_expected.teal
```diff
@@ -0,0 +1,41 @@
+#pragma version 6          Case n = 1   Case n = 2
+int 42                      >1          >2
+store 0                     0:1         0:2
+int 0                       >0          >0
+callsub factorial_0         CALL        CALL
+load 0                      >1          >2   
+return                      RET(1!==1)  RET(2!==1*2)
+
+// factorial
+factorial_0:               *>0         *>0          *>2,0,1,2
```
jasonpaulos, line 10: I'd prefer if the commentary to the right is behind `//` so this file is still syntacticly valid

pyteal/ast/scratch.py
michaeldiamant: I held off commenting because I thought a backwards compatibility concern may explain the construction.

* Since there's latitude for API changes and because `slot` and `index_expression` are mutually exclusive, what do folks think of using `Union` here?
  * For example:  `slot: ScratchSlot | Expr`
* Aside:   Is it considered atypical to define a new `Expr` implementation specific to `Op.stores`?
  * Doing removes branching within `ScratchLoad`.
  * I don't have enough insight to understand if adding a new type relating to the scratch slot concept adds complexity elsewhere (e.g. IR).
  * In my _fuzzy_ mental model, I see a further extension.
    * _Disclaimer_:  These thoughts aren't fully formed.  And I don't think I understand the full implications.  I'm sharing to evidence my thinking.  It's _not_ a change request. 
    *  Define an `Expr` like `LeafExpr` that acts as the top-level type for any kind of scratch load.
    *  Define implementations specific to stack vs argument provided index values.
    * The approach enables abstracting over scratch loads independent of invocation scheme.
    * I can envision using pattern matching (as shown in https://docs.python.org/3/whatsnew/3.10.html#pep-634-structural-pattern-matching) to help group/isolate program behaviors.

pyteal/ast/scratchvar.py
jasonpaulos: Link to the prior discussion: https://github.com/algorand/pyteal/pull/198#discussion_r810271366

pyteal/ast/scratch.py
```diff
@@ -59,41 +59,92 @@ def load(self, type: TealType = TealType.anytype) -> "ScratchLoad":
...
+
     def __hash__(self):
-        return hash(self.id)
+        return hash(self.slot.id)
 
+    def type_of(self):
+        return TealType.uint64
 
-ScratchSlot.__module__ = "pyteal"
+    def has_return(self):
```
michaeldiamant, line 89: @tzaffi Clarification question:  How have folks been treating membership for `LeafExpr`?

* I see `LeafExpr` exists to delineate `Expr`s that do _not_ always return from the current subroutine or program.
* Based on `ScratchIndex.has_return` implementation, it seems `ScratchIndex` is eligible.
  * Aside:  The same observation holds for other classes in `scratch.py` (e.g. `ScratchLoad`).
* I don't see references to `LeafExpr` for controlling program behavior.  
  * I'm a little surprised because I assumed the type marker can be used to control behavior.
  * Nevertheless, I'm _imagining_ `LeafExpr` as a member of an algebraic data type (ADT).  And in my experience, ADTs help maintainers gauge the shape of supported behaviors.
* In summary - _If_ `ScratchIndex` is eligible to be a `LeafExpr`, I don't feel strongly today about making a change.  I imagine having a stronger opinion when I develop deeper PyTeal intuitions.

pyteal/ast/scratchvar.py
michaeldiamant: @jasonpaulos Thanks for the pointer.  

With the disclaimer that I still have gaps in my intuition, here's my thinking:
* Is it accurate to say that `DynamicScratchVar` is _only_ intended for internal-to-the-compiler usage rather than also being used by PyTeal developers?
  * If also intended for PyTeal developers, I'd be interested in understanding the use cases. 
  * If intended for internal usage only, I'm wondering:  How can we more clearly delineate public vs private APIs?   
* With the assumption that `DynamicScratchVar` usage is _internal_, I feel like I'm _not_ convinced `DynamicScratchVar` _needs_ to extend `ScratchVar`.
  * Take my skepticism with a grain of salt.  I'm probably skeptical because there's gaps in my intuition.  I'd at least like to keep considering the topic.
  * As written, I see `evaluateSubroutine` in  `subroutine.py` depends on methods with the same signatures to invoke the subroutine.
    * I guess duck typing is at play here.  And it explains why _no_ tests fail when I removed the inheritance locally.
  * The skepticism arises for reasons similar to what @tzaffi raises. And my prior experiences bias me towards composition over inheritance.

pyteal/ast/scratchvar.py
```diff
@@ -47,5 +54,86 @@ def load(self) -> ScratchLoad:
...
+    """
+
+    def __init__(self, ttype: TealType = TealType.anytype):
+        """Create a new DynamicScratchVar which references other ScratchVar's
+
+        Args:
+            ttype (optional): The type that this variable can hold. Defaults to TealType.anytype.
+        """
+        self.type = ttype
+        self._indexer = ScratchVar(TealType.uint64)
```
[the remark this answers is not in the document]
michaeldiamant: _Assuming_ `DynamicScratchVar` continues extending `ScratchVar`, the proposal makes sense to me.

* I had planned to ask if it's intentional that `super().__init__` is _not_ called.  The proposal eliminates the ambiguity.  And I think the proposal provides equivalent functionality.
* I say _assuming_ because I'm still evaluating the class hierarchy.  I think I understand how https://github.com/algorand/pyteal/pull/198#discussion_r810271366 motivates inheritance.  But, it's unclear to me if that's the only viable path.  
  * Don't put much weight on my skepticism.  It _probably_ is the appropriate choice.  I have gaps in my intuition.

pyteal/ast/scratchvar.py
```diff
@@ -47,5 +54,86 @@ def load(self) -> ScratchLoad:
...
+        require_type(value, self.type)
+        index = ScratchLoad(self._indexer.slot, TealType.uint64)
+        return ScratchStore(slot=None, value=value, index_expression=index)
+
+    def load(self) -> ScratchLoad:
+        """Load the current value from the referenced ScratchVar."""
+        index = ScratchLoad(self._indexer.slot, TealType.uint64)
+        return ScratchLoad(slot=None, type=self.type, index_expression=index)
+
+    def index(self) -> Expr:
```
michaeldiamant, line 130: @tzaffi Clarification question:  What motivates the difference in return type for these methods?
* `ScratchVar` returns `ScratchIndex`.
* `DynamicScratchVar.index` returns `ScratchLoad`.

Context:
* As I reviewed, I was surprised to see `index` returns a general type (`Expr`).
* I'm probably missing intuition.  Naively, I'm imagining `ScratchIndex` is the expected return type in all cases.
* As written, I realize `ScratchIndex` requires `ScratchSlot`.  Though it's unclear to me _why_ the `ScratchSlot` dependency exists.
  * There's portions of _IR_ + `ScratchSlot` interaction I still don't follow.  Possibly, the missing intuition lies within _IR_.
  * I _suspect_ the side effects in `ScratchSlot` constructor may also be at play.
  * In my imagination, it's possible to accept either (a) `Int` or (b) union of additional data types.

pyteal/ast/subroutine.py
```diff
@@ -26,12 +27,17 @@ def __init__(
...
+
+        self.expected_arg_types: List[type] = []
+
         if not callable(implementation):
             raise TealInputError("Input to SubroutineDefinition is not callable")
 
         sig = signature(implementation)
 
         for name, param in sig.parameters.items():
+            param = sig.parameters[name]
```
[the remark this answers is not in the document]
michaeldiamant: I have the same question.  I'm imagining it's incidental and can be removed.

pyteal/ast/subroutine.py
```diff
@@ -49,15 +55,32 @@ def __init__(
                     )
                 )
 
+            expected_arg_type: type = Expr
```
michaeldiamant, line 58: @tzaffi Can you help me understand how it's safe to assume the type is `Expr` here?

* It appears a check for `Expr` happens _after_ building `expected_arg_types` on line 76.
* My interpretation is:  It's _not_ safe to assume the data type is `Expr` here.  Though any violating data types will be caught on line 76.
* If my understanding is accurate, then:
  * My concern is someone will later change ordering/validation logic that unexpectedly retains an undesired data type.
  * I'd try to perform validation _before_ (or at the same time as) building `expected_arg_types`.

pyteal/ast/subroutine.py
```diff
@@ -49,15 +55,32 @@ def __init__(
...
+            if name in implementation.__annotations__:
```
michaeldiamant, line 59: @tzaffi To test my understanding:  Is there a situation where `name` won't be in `__annotations__`?

If `name` is always present, I'm then wondering what requires using `signature` and `__annotations__`?

pyteal/ast/scratch.py
```diff
@@ -59,41 +59,92 @@ def load(self, type: TealType = TealType.anytype) -> "ScratchLoad":
...
+            raise TealInputError(
+                "cannot handle slot of type {}".format(type(self.slot))
+            )
+
         self.slot = slot
         self.type = type
+        self.index_expression = index_expression
 
     def __str__(self):
         return "(Load {})".format(self.slot)
```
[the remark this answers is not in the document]
tzaffi: 👍

pyteal/ast/scratch.py
```diff
@@ -59,41 +59,92 @@ def load(self, type: TealType = TealType.anytype) -> "ScratchLoad":
...
 
     def __teal__(self, options: "CompileOptions"):
         from ..ir import TealOp, Op, TealBlock
 
-        op = TealOp(self, Op.load, self.slot)
+        if self.index_expression is not None:
+            op = TealOp(self, Op.loads)
+            return TealBlock.FromOp(options, op, self.index_expression)
```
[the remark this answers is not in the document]
tzaffi: 👍

pyteal/ast/scratch.py
```diff
@@ -109,23 +160,42 @@ def has_return(self):
...
+            raise TealInternalError(
+                "Exactly one of slot or index_expressions must be provided"
+            )
+
         self.slot = slot
         self.value = value
+        self.index_expression = index_expression
 
     def __str__(self):
         return "(Store {} {})".format(self.slot, self.value)
```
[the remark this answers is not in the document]
tzaffi: 👍

pyteal/ast/scratch.py
```diff
@@ -109,23 +160,42 @@ def has_return(self):
...
 
     def __teal__(self, options: "CompileOptions"):
         from ..ir import TealOp, Op, TealBlock
 
+        if self.index_expression is not None:
```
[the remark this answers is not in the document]
tzaffi: 👍

pyteal/ast/scratch.py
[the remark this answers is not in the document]
tzaffi: It's still not entirely clear to me what our philosophy for `__str__()` is in PyTeal. I found these methods useful when debugging, and therefore it was useful for me to be able to distinguish between a `StackStore` and `ScratchStackStore` before it was headed for TEAL assembly. However, there might be a more important purpose whereby it's important to keep the printouts the same. I'll add this to our talking points today.

pyteal/ast/scratchvar.py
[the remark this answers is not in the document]
tzaffi: I intentionally did not call `__super()__` as I wanted to maintain the compositional aspect of `DynamicScratchVar`. Am not sure yet what the implications are of Jason's proposal is. This is another good talking point.

pyteal/ast/subroutine.py
```diff
@@ -1,18 +1,19 @@
...
 from .seq import Seq
-from .scratchvar import ScratchVar
+from .scratchvar import DynamicScratchVar, ScratchVar
 
 if TYPE_CHECKING:
     from ..compiler import CompileOptions
 
 
 class SubroutineDefinition:
+    PARAM_TYPES = (Expr, ScratchVar)
```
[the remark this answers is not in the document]
tzaffi: Would it make more sense to keep as is and then just have PR #154 make the appropriate changes? Alternatively, I can attempt to make this more future proof -another good discssion point.

pyteal/ast/subroutine_test.py
```diff
@@ -1,4 +1,5 @@
 from typing import List
+from xml.dom import InvalidModificationErr
```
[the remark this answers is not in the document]
tzaffi: 👍

pyteal/ast/subroutine.py
```diff
@@ -49,15 +55,32 @@ def __init__(
...
 
+            expected_arg_type: type = Expr
+            if name in implementation.__annotations__:
+                ptype = implementation.__annotations__[name]
+                if isclass(ptype) and issubclass(ptype, ScratchVar):
+                    expected_arg_type = ScratchVar
+
+            self.expected_arg_types.append(expected_arg_type)
+
         for var, var_type in implementation.__annotations__.items():
```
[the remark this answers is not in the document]
tzaffi: 👍

pyteal/ast/subroutine_test.py
```diff
@@ -78,6 +82,100 @@ def fnWithPartialExprAnnotations(a, b: Expr) -> Expr:
...
+
+    def fnWithMixedAnns4AndIntReturn(a: Expr, b: ScratchVar) -> Bytes:
+        return Bytes("helo")
+
+    sv = ScratchVar()
+    x = Int(42)
+    s = Bytes("hello")
+    cases = [
+        ("vannila 1", fnWithNoAnnotations, [x, s], None),
+        ("vannila 2", fnWithNoAnnotations, [x, x], None),
```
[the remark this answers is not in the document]
tzaffi: 👍

pyteal/ast/subroutine.py
```diff
@@ -49,15 +55,32 @@ def __init__(
...
-            if var_type is not Expr:
-                stub = "Return" if var == "return" else ("parameter " + var)
+            if var == "return" and not (
+                isclass(var_type) and issubclass(var_type, Expr)
```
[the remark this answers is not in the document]
tzaffi: 👍

pyteal/ast/subroutine.py
```diff
@@ -26,12 +27,17 @@ def __init__(
         self.id = SubroutineDefinition.nextSubroutineId
         SubroutineDefinition.nextSubroutineId += 1
 
+        self.by_ref_args: Set[str] = set()
+
+        self.expected_arg_types: List[type] = []
```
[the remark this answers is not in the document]
tzaffi: 👍

pyteal/ast/subroutine.py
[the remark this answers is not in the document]
ahangsu: I agree to keep as is, this is just note taking for myself for further implementation. I will try my best to keep the structure in following PRs.

pyteal/ast/subroutine.py
```diff
@@ -248,9 +302,55 @@ def __call__(self, fnImplementation: Callable[..., Expr]) -> SubroutineFnWrapper
...
+        if param in subroutine.by_ref_args:
+            argVar = DynamicScratchVar(TealType.anytype)
+            loaded = argVar
+        else:
+            argVar = ScratchVar(TealType.anytype)
+            loaded = argVar.load()
+
+        return argVar, loaded
+
+    args = subroutine.arguments()
```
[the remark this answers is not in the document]
tzaffi: I believe that one gets back on OrderedDict so that we should be safe here, but the concerns you raise are worthy of further investigation on my part.

pyteal/ast/subroutine.py
```diff
@@ -248,9 +302,55 @@ def __call__(self, fnImplementation: Callable[..., Expr]) -> SubroutineFnWrapper
...
+    Usage (B) for compile-time: "loadedArgs"
+        These are expressions supplied to the user-defined PyTEAL function.
+        The loadedArgs are invoked to by the subroutine to create a self-contained AST which will translate into a TEAL subroutine.
+
+    In both usage cases, we need to handle
+
+    2 Argument Types
+    - -------- -----
+    Type 1 (by-value): these have python type Expr
+    Type 2 (by-reference): these hae python type ScratchVar
```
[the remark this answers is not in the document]
tzaffi: 👍

pyteal/ast/subroutine.py
```diff
@@ -248,9 +302,55 @@ def __call__(self, fnImplementation: Callable[..., Expr]) -> SubroutineFnWrapper
...
+
+    Usage (A) "argumentVars" - Storing pre-placed stack variables into local scratch space:
+        Type 1. (by-value) use ScratchVar.store() to pick the actual value into a local scratch space
+        Type 2. (by-reference) ALSO use ScratchVar.store() to pick up from the stack
+            NOTE: SubroutineCall.__teal__() has placed the _SLOT INDEX_ on the stack so this is stored into the local scratch space
+
+    Usage (B) "loadedArgs" - Passing through to an invoked PyTEAL subroutine AST:
+        Type 1. (by-value) use ScratchVar.load() to have an Expr that can be compiled in python by the PyTEAL subroutine
+        Type 2. (by-reference) use a DynamicScratchVar as the user will have written the PyTEAL in a way that satifies
```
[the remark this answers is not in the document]
tzaffi: 👍

pyteal/ast/subroutine_test.py
```diff
@@ -78,6 +82,100 @@ def fnWithPartialExprAnnotations(a, b: Expr) -> Expr:
...
+                assert invocation.subroutine is definition, case_name
+                assert invocation.args == args, case_name
+                assert invocation.has_return() is False, case_name
+
+            except Exception as e:
+                if isinstance(e, AssertionError):
+                    raise
+                assert (
+                    not e
+                ), f"EXPECTED SUCCESS. encountered unexpected error during invocation case <{case_name}>: {e}"
```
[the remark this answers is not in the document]
tzaffi: I'll try and simplify. I followed this approach because of being stuck inside the purest.raises context and wanting to force my way out but there is probably a leaner approach along the lines you are suggesting.

pyteal/ast/__init__.py
```diff
@@ -120,141 +120,150 @@
...
+    ScratchStore,
+)
+from .scratchvar import DynamicScratchVar, ScratchVar
 from .maybe import MaybeValue
 from .multi import MultiValue
 
 __all__ = [
-    "Expr",
-    "LeafExpr",
+    "AccountParam",
```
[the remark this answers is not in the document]
tzaffi: I should NOT be modifying an autogenerated file. Should have realized that this was. Will modify the generator instead and make sure there is a clear way for others not to fall into the same trap as I did.

pyteal/ast/scratch.py
```diff
@@ -59,41 +59,92 @@ def load(self, type: TealType = TealType.anytype) -> "ScratchLoad":
...
+    def __init__(self, slot: ScratchSlot):
+        super().__init__()
+        self.slot = slot
+
+    def __str__(self):
+        return "(ScratchIndex {})".format(self.slot)
+
     def __hash__(self):
-        return hash(self.id)
+        return hash(self.slot.id)
```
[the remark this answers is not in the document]
tzaffi: Another discussion point

pyteal/ast/scratch.py
tzaffi: Good discussion point

tzaffi: My initial thoughts are to stick with @jasonpaulos suggestions to maintain consistency and simplicity.

pyteal/ast/scratch_test.py
tzaffi: 👍

pyteal/ast/scratch_test.py
tzaffi: 👍

pyteal/ast/scratch_test.py
tzaffi: 👍

pyteal/ast/scratchvar.py
tzaffi: 👍

pyteal/ast/scratchvar_test.py
tzaffi: 👍

pyteal/ast/scratchvar_test.py
tzaffi: 👍

tests/teal/fac_by_ref_expected.teal
tzaffi: 👍

pyteal/ast/scratch.py
tzaffi: Worthy of discussion

pyteal/ast/scratchvar.py
tzaffi: ScratchIndex is a brand new type so I was not super careful here. I think the discrepancy you point out is additional ammunition for @jasonpaulos 's idea to tighten the class hierarchy between ScratchVar and DynamicScratchCar

pyteal/ast/subroutine.py
tzaffi: 👍

pyteal/ast/subroutine.py
tzaffi: Name won't show up in __annotations__ in the case that defined subroutine didn't have an explicit annotation, and in that case we default to Expr

pyteal/ast/scratchvar.py
michaeldiamant: Summarizing verbal discussion below.

@tzaffi I'm good to resolve the thread on my side.  However, since I didn't open the thread, I leave it open in case you wish to add something.  Feel welcomed to resolve if you're good here too.

> Is it accurate to say that DynamicScratchVar is only intended for internal-to-the-compiler usage rather than also being used by PyTeal developers?

* `DynamicScratchVar` is intended for internal _and_ PyTeal developer usage.
* An example use case involves allowing a PyTeal developer to reserve an array of slots.  The cursor provided by `DynamicScratchVar` is a step towards providing such a facility.
* Given these points, I don't have further reason to suggest avoiding inheritance here.

pyteal/ast/scratchvar.py
michaeldiamant: Summarizing a verbal below.  Resolving because the verbal discussion addressed my questions.

* The return types vary because it's not possible to align the types in a way that generates correct TEAL programs.  
* In general, it's a by-design decision to bias towards `Expr` rather than a more specific type to minimize public API changes.

pyteal/ast/scratch.py
michaeldiamant: I don't think we covered the topic verbally.

* Since I don't feel strongly and I don't want to hold up the PR, I'm on-board with @jasonpaulos's suggestion.
* If there's interest in the discussion point my prior remark raises, I'm happy to discuss live out-of-band to the PR.

pyteal/ast/scratch.py
michaeldiamant: I don't think we directly discussed the topic verbally.

* Although I'm interested in a discussion eventually, I'm not looking to hold the PR.  So, I'll resolve the thread.
* The verbal discussion highlighted a bias towards `Expr` usage rather than specific `Expr` subtypes.  Indirectly, I imagine that bias may explain why `LeafExpr` is not used more pervasively.

pyteal/ast/subroutine_test.py
```diff
@@ -40,6 +41,9 @@ def fnWithOnlyArgExprAnnotations(a: Expr, b: Expr):
     def fnWithPartialExprAnnotations(a, b: Expr) -> Expr:
         return Return()
 
+    def fnWithMixedAnnotations(a: ScratchVar, b: Expr) -> Return:
```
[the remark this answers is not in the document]
tzaffi: I've renamed this test case `fnReturningExprSubclass` and moved it into the `test_subroutine_definition_invalid()` test

pyteal/ast/subroutine_test.py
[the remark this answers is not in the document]
tzaffi: Just got rid of the extra handler... we lose the case info, but gain simplicity.

pyteal/ast/subroutine.py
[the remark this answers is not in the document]
tzaffi: Results of further investigation:
We can see that in practice (running in Python 3.9) that this code converts `odict_keys` to `list`, so we're safe here:
```python
>>> from inspect import signature
>>> def foo(x, y: int, z: str):
...     pass
... 
>>> sig = signature(foo)
>>> params = sig.parameters
>>> params
mappingproxy(OrderedDict([('x', <Parameter "x">), ('y', <Parameter "y: int">), ('z', <Parameter "z: str">)]))
>>> params.keys()
odict_keys(['x', 'y', 'z'])
```

According to [python's official docs](https://docs.python.org/3/library/inspect.html#inspect.Signature.parameters):
> Parameters appear in strict definition order, including keyword-only parameters.

But they add a troubling caveat, that looks more fundamental _regardless_ of approach taken here:
> Changed in version 3.7: Python only explicitly guaranteed that it preserved the declaration order of keyword-only parameters as of version 3.7, although in practice this order had always been preserved in Python 3.

pyteal/ast/scratch.py
tzaffi: Gonna push back on this one, because optional arguments cannot come before required parameters so setting default `= None` would break the signature. So one has to explicitly provide `None` in the slot when an `index_expression` is provided.

pyteal/ast/scratch.py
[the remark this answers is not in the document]
tzaffi: 👍

pyteal/ast/scratchvar.py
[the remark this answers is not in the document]
tzaffi: 👍

scripts/generate_init.py
```diff
@@ -37,7 +37,7 @@ def generate_init_pyi() -> str:
     start_idx = init_contents.index(begin_flag)
     end_idx = init_contents.index(end_flag)
 
-    all_imports = ",\n    ".join(['"{}"'.format(s) for s in static_all])
+    all_imports = ",\n    ".join(['"{}"'.format(s) for s in sorted(set(static_all))])
```
tzaffi, line 40: sorting and uniquifying

pyteal/ast/subroutine_test.py
```diff
@@ -78,6 +78,76 @@ def fnWithPartialExprAnnotations(a, b: Expr) -> Expr:
...
+    sv = ScratchVar()
+    x = Int(42)
+    s = Bytes("hello")
+    cases = [
+        ("vanilla 1", fnWithNoAnnotations, [x, s], None),
+        ("vanilla 2", fnWithNoAnnotations, [x, x], None),
+        ("vanilla no sv's allowed 1", fnWithNoAnnotations, [x, sv], TealInputError),
+        ("exprs 1", fnWithExprAnnotations, [x, s], None),
+        ("exprs 2", fnWithExprAnnotations, [x, x], None),
+        ("exprs no sv's asllowed 1", fnWithExprAnnotations, [x, sv], TealInputError),
```
michaeldiamant, line 109: @tzaffi Friendly pointer to minor typo.

```suggestion
        ("exprs no sv's allowed 1", fnWithExprAnnotations, [x, sv], TealInputError),
```

pyteal/ast/subroutine.py
michaeldiamant: @tzaffi To fill in the gap in my intuition:  What prevents (or makes undesirable) deferring the validation until `evaluateSubroutine`?

* As I understand it, the `SubrouteDefinition` constructor is forced to assume a parameter type in order to build `expected_arg_types` and `by_ref_args`.  Obviously, without a type parameter hint, a _guess_ is made:  `Expr`.
* Then, during `invoke` the _guess_ is checked against the provided argument and aborts evaluation when the _guess_ is inaccurate.
* I started down the path because I didn't immediately grok _why_ `Expr` is chosen as the default.  With the above understanding in mind, I'm wondering if there's more intuition I missed that explains why validation happens in the constructor with partial information.
  * My _assumption_ is:  Enforcing `Expr` as the only supported type without annotation simplifies life for PyTeal maintainers.  If that's the case, it works for me though I want to verify.

tzaffi: > @tzaffi To fill in the gap in my intuition: What prevents (or makes undesirable) deferring the validation until `evaluateSubroutine`?
In general, it's good to "fail fast" so there would have to be a strong argument (e.g. an object is invalid because it can be made valid later) to defer validation to a later stage.
> 
> * As I understand it, the `SubrouteDefinition` constructor is forced to assume a parameter type in order to build `expected_arg_types` and `by_ref_args`.  Obviously, without a type parameter hint, a _guess_ is made:  `Expr`.
It's more than a "guess". We have just introduced subroutine annotations in #182, so by default parameter types should be backwards compatible, and therefore an `Expr`.


> * Then, during `invoke` the _guess_ is checked against the provided argument and aborts evaluation when the _guess_ is inaccurate.
If we view this not as a "guess" but as an "implicit declaration" does aborting evaluation sound less onerous?

> * I started down the path because I didn't immediately grok _why_ `Expr` is chosen as the default.  With the above understanding in mind, I'm wondering if there's more intuition I missed that explains why validation happens in the constructor with partial information.
The reason is "backwards compatibility".

>   
>   * My _assumption_ is:  Enforcing `Expr` as the only supported type without annotation simplifies life for PyTeal maintainers.  If that's the case, it works for me though I want to verify.
alas, that simplified life as about to get more complicated with the upcoming introduction of ABI type params

pyteal/ast/scratch.py
```diff
@@ -1,8 +1,8 @@
-from typing import TYPE_CHECKING
+from typing import cast, TYPE_CHECKING, Optional
```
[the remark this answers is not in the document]
tzaffi: I've created issue #215 to address this concern.

pyteal/ast/subroutine_test.py
```diff
@@ -128,20 +121,6 @@ def fnWithMixedAnns4AndIntReturn(a: Expr, b: ScratchVar) -> Bytes:
         ("mixed3 copacetic", fnWithMixedAnns3, [s, sv], None),
         ("mixed3 flipped", fnWithMixedAnns3, [sv, x], TealInputError),
         ("mixed3 missing the sv", fnWithMixedAnns3, [x, s], TealInputError),
-        (
-            "mixed3 missing the non-sv",
-            fnWithMixedAnns4AndIntReturn,
-            [sv, sv],
-            TealInputError,
-        ),
```
jasonpaulos: Why get rid of this test?

pyteal/ast/subroutine_test.py
```diff
@@ -212,20 +190,29 @@ def fnWithScratchVarSubclass(a, b: DynamicScratchVar):
...
+        ),
+        (
+            fnWithMixedAnns4AndBytesReturn,
+            "Function has return of disallowed type <class 'pyteal.Bytes'>",
+        ),
     )
 
     for fn, msg in cases:
         with pytest.raises(TealInputError) as e:
+            print(f"case=[{msg}]")
```
jasonpaulos, line 215: Is this print statement intentional?

jasonpaulos (commented):
Implementation looks good, I just have lingering questions

pyteal/ast/scratch.py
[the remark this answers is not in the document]
michaeldiamant: @tzaffi Thanks for opening https://github.com/algorand/pyteal/issues/215 - resolving.

pyteal/ast/subroutine.py
michaeldiamant: @tzaffi Thanks for the explanation - makes sense to me.  

Since I spent some time understanding the topic, I thought an inline comment _may_ be beneficial.  I opened an optional child PR:  https://github.com/algorand/pyteal/pull/217.

If you don't feel the PR is an improvement worth taking in, I'll happily close it.  Either way, I learned a bit more about PyTeal.

scripts/generate_init.py
tzaffi: Looking closer, it turns out no de-duping actually occurs. So I'll change the code to make an assertion instead.

pyteal/ast/subroutine_test.py
tzaffi: I moved thiis test case to `test_subroutine_definition_invalid()` because we are no longer allowing `Expr` subclasses for annotation return types.

pyteal/ast/subroutine_test.py
tzaffi: Yes. Instead of catching the exception and re-raising with the `case` as I was doing before, now I'm just printing out each case. So if you run `pytest -sv` you can see the last `case=XYZ` that got printed and infer the broken case.

pyteal/ast/subroutine_test.py
tzaffi: I also renamed the function `fnWithMixedAnns4AndBytesReturn` (it was wrongly named before)

jasonpaulos (approved):
LGTM

PR #212, opened 2022-02-28
Test on Python 3.10

jasonpaulos:
Test with Python 3.10 as well in our CI process.


michaeldiamant (approved):
@jasonpaulos Thanks for adding 3.10 into the test mix.  ☕

PR #220, opened 2022-03-02
WIP: PyTEAL Developer Docs

tzaffi:
# Just a stub at this point...

following up on [retro promise](https://algorand.atlassian.net/wiki/spaces/SCY/pages/2347892761/Scytale+Retro+02+16+2022)


docs/repository.rst
```diff
@@ -0,0 +1,21 @@
+PyTeal Repository Structure
```
michaeldiamant, line 1: I think a question the PR prompts is:  Do we want to apply the same docs model for contributors as users?

* The user guide is generated from rst files and lands on https://pyteal.readthedocs.io/en/stable/.  Existing contributor docs are markdown local to the repo (e.g. https://github.com/algorand/pyteal/blob/master/CONTRIBUTING.md).
* I have an opinion, but I don't feel strongly and am interested in discussion:
  * Particularly because PyTeal is an open source project, it seems _reasonable_ to me that contributor docs ought to follow the user guide model.
  * While I think consistency is worth considering, I also don't believe _every_ internal doc ought to become an rst.  For example, I favor markdown files local to specific directories for providing context.

docs/repository.rst
```diff
@@ -0,0 +1,21 @@
...
+
+pyteal/compiler
+~~~~~~~~~~~~~~~
+
+The basic compilation algorithms.
+
+pyteal/ir
+~~~~~~~~~
+
+Code for PyTeal's Intermediate Representation (IR) which the compiler generates midway from the AST before finally converting to TEAL.
```
michaeldiamant, line 21: Optional and mostly unrelated to the line:  

I think reviewing merged PRs for a particular class of problem is an effective way (for me) to accelerate learning.

In PyTeal's case, I think linking to a merged opcode addition PR is a great way to help contributors understand the space.
* Obviously, some opcodes are more involved than others.
* I think we'd want to pick an example that illustrates a complete definition of done (e.g. includes user guide doc updates).

If we think it's a productive idea and we don't have a candidate PR, we can always use the next opcode addition as the candidate.

docs/repository.rst
tzaffi: Since we already have the bulk of our documents in RST, it made sense to me to continue building these out. But like yourself, I'm not super opinionated about it. I do dread the thought of having two different documentation systems with two different standards in the same repo. But a one-off doc here and there is certainly fine.

docs/repository.rst
tzaffi: 👍 - good addition

PR #221, opened 2022-03-02
Adjust ABI test relative imports to acquiesce mypy and IDE

michaeldiamant:
Adjust relative imports in ABI tests to address mypy and IDE (PyCharm) errors.

I'm not sure if the change I'm suggesting makes sense.  Here's the path I took:
* I began reviewing _feature/abi_ to understand ABI support in PyTeal and quickly observed IDE errors in tests.  Which is to say, I know little about historical changes and ongoing issues.
* Seeing IDE errors, I ran `mypy` (Python 3.10.2) and discovered mypy sees a subset of IDE errors:  [mypy-pyteal.log](https://github.com/algorand/pyteal/files/8172298/mypy-pyteal.log).
* Studying the relative imports and `pyteal/ast/__init__.py`, I see `abi` types _should_ be resolved with `from ... import *`.
* PR changes:
  * Since I don't understand the root issue and I saw mypy errors, I supplied additional relative imports.
  * Secondarily, I _think_ `from ... import CompileOptions` is duplicative.  Since mypy does _not_ raise an error, I propose removing the statement.

### DIFF
--- a/pyteal/ast/abi/array_test.py
+++ b/pyteal/ast/abi/array_test.py
@@ -1,14 +1,13 @@
 from typing import List, cast
 import pytest
 
+from .. import *
 from ... import *
+
 from .type import substringForDecoding
 from .tuple import encodeTuple
 from .bool import boolSequenceLength
 from .array import ArrayElement
-
-# this is not necessary but mypy complains if it's not included
-from ... import CompileOptions
 
 options = CompileOptions(version=5)
 
--- a/pyteal/ast/abi/bool_test.py
+++ b/pyteal/ast/abi/bool_test.py
@@ -1,6 +1,6 @@
 from typing import NamedTuple, List
-import pytest
 
+from .. import *
 from ... import *
 from .bool import (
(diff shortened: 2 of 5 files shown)


ahangsu (approved):
Feels good to me! I was wondering what is wrong with testing since pytest is passing.

jasonpaulos (commented):
I believe the problem is actually because "abi" is missing from `pyteal/__init__.pyi`.

I think it got lost in 7a66db7ffab07fe7ec400f78d9f2453e3f9c64c8 which merged master

If I rerun `python scripts/generate_init.py` locally, the warnings go away for me

edit: root cause explained in https://github.com/algorand/pyteal/pull/223

michaeldiamant:
@jasonpaulos Thanks for catching the root issue cause and opening a PR to address it:  https://github.com/algorand/pyteal/pull/224.  ☕ 

Closing the PR because it's _not_ the appropriate fix.

PR #223, opened 2022-03-02
Fix build script invocation

jasonpaulos:
Currently the build step `python3 -c "import pyteal" scripts/generate_init.py --check` does not actually invoke `scripts/generate_init.py`, it just runs `import pyteal` then exits.

This PR fixes that.

You can verify by seeing that the build now includes a line that says `No changes in __init__.py`, which is printed from `scripts/generate_init.py`.

In order to make the script work in CI, I also added a line to install the current version of pyteal in editable state.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,12 +20,14 @@ jobs:
         with:
           fetch-depth: 0
       - name: Install pip dependencies
-        run: pip install -r requirements.txt
+        run: |
+          pip install -r requirements.txt
+          pip install -e .
       - name: Build and Test
         run: |
           pytest
           mypy pyteal
-          python3 -c "import pyteal" scripts/generate_init.py --check
+          python scripts/generate_init.py --check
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -20,12 +20,14 @@ jobs:
         with:
           fetch-depth: 0
       - name: Install pip dependencies
-        run: pip install -r requirements.txt
+        run: |
+          pip install -r requirements.txt
+          pip install -e .
```
tzaffi, line 25: Future proofing?

jasonpaulos: In theory this is necessary for the `from pyteal import __all__ as static_all` line to work in `scripts/generate_init.py`, but I haven't tested without it

tzaffi: 👍

tzaffi (approved):
LGTM

PR #224, opened 2022-03-03
Merge with master and regenerate `__index__.pyi`

jasonpaulos:
Contains fix from #223 and regenerates `__index__.pyi` to fix missing `abi`

Issue #218, opened 2022-03-01
Integrate linter:  flake8 


michaeldiamant:
## Problem

A linter like [flake8](https://flake8.pycqa.org/en/latest/) catches issues like unused imports, and lack of adherence to language standards.  It's ideal to catch these issues during the CI process because it provides a consistent approach to writing software in all parts of a repo. 

## Solution

While multiple linters exist, I propose starting with flake8 and later swapping out and/or adding others.  https://github.com/algorand/pyteal/pull/213 illustrates partial integration and can be extended.

A full solution includes:
* CI process runs flake8.
* Ensure CI and local building share the same steps.
  * Presumably, a _script_ can be written to accomplish the goal.  I'm being vague to allow freedom for interpretation. 
* Upon merge, flake8 does _not_ find errors.  Implies fixing errors for the to-be-agreed upon flake8 configuration.
* Add pyteal repo docs explaining flake8 usage.

## Dependencies

N/A

michaeldiamant:
Consider https://github.com/bndr/pipreqs to isolate unused dependencies.  Thanks to @tzaffi for suggesting in https://github.com/algorand/pyteal/pull/249#discussion_r832680338.

tzaffi:
That's a fine idea. Upon further investigation, one contender to `pipreqs` is [pigar](https://github.com/damnever/pigar). It seems more active than pipreqs, if not as polished.

michaeldiamant:
FYI - I opened an issue to tracking monitoring unused dependencies because we didn't tackle it here:  https://github.com/algorand/pyteal/issues/298.

PR #225, opened 2022-03-03
Reorder CI build steps to change source files before running tests

michaeldiamant:
Optionally extends #224 to resorder CI build steps.  PR proposes the following ordering constraints:
* Finish all source code modifications before running tests.  Rationale:  A source code modification _can_ cause test behavior to change.
* Run type checking before tests.  Rationale:  Type checking issues happen before runtime (e.g. test runs).

If the PR feels like it's opening up a debate, I'll happily close the PR.  
* I'm _not_ looking to spend much time here. I only provided the PR because #224 is in the same neighborhood, and Github won't allow a multi-line _suggestion_.
* I'm not attached to the PR because I think #218 offers another opportunity to circle back.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,10 +25,10 @@ jobs:
           pip install -e .
       - name: Build and Test
         run: |
-          pytest
-          mypy pyteal
           python scripts/generate_init.py --check
           black --check .
+          mypy pyteal
+          pytest
   build-docset:
     runs-on: ubuntu-20.04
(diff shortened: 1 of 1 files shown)


jasonpaulos (approved):
Reordering makes sense, but note that the CI does not change any source files with these steps.

Instead, it checks that the source files would _not_ be changed if our formatter or generator script runs. This protects us from merging PRs with bad source states and ensures whatever's on the branch is correct.

PR #234, opened 2022-03-05
Fix subroutine mutual recursion with different argument counts bug

### LINKED ISSUE #233: Incorrect TEAL code produced for mutually recursive subroutines with different argument counts
## Summary

There is a bug in the implementation of [`spillLocalSlotsDuringRecursion`](https://github.com/algorand/pyteal/blob/f2598da3dec6041051df4442cd9cc7e3c373674d/pyteal/compiler/subroutines.py#L65-L204) that causes incorrect TEAL code to be generated for mutually recursive subroutines with different argument counts.

The problem is that the variable `numArgs` is the number of arguments of the calling subroutine, **NOT** the number of arguments in the called subroutine. This will cause incorrect slot spilling and restoring code to be generated, in some cases producing an infinite recursive loop.

## Example

Below is an example of a PyTeal that's affected by this bug:

```python
from pyteal import *

@Subroutine(TealType.uint64)
def factorial(i: Expr) -> Expr:
    return If(i <= Int(1), Int(1), factorial_intermediate(i - Int(1), Bytes("inconsequential")) * i)

@Subroutine(TealType.uint64)
def factorial_intermediate(i: Expr, j: Expr) -> Expr:
    return Seq(Log(j), factorial(i))

program = Return(factorial(Int(4)) == Int(24))
```

It currently produces the following **incorrect** TEAL code (with PyTeal v0.10.0), with my commentary:
```
#pragma version 5
int 4
callsub factorial_0
int 24
==
return

// factorial
factorial_0:
store 0
load 0
int 1
<=
bnz factorial_0_l2
load 0
int 1
-
byte "inconsequential"
load 0
swap    // the compiler thinks factorialintermediate_1 only takes 1 argument, but it takes 2. Only one value is uncovered
callsub factorialintermediate_1
swap
store 0
load 0
*
b factorial_0_l3
factorial_0_l2:
int 1
factorial_0_l3:
retsub

// factorial_intermediate
factorialintermediate_1:
store 2
store 1
load 2
log
load 1
load 1
load 2
uncover 3
uncover 3    // the compiler thinks factorial_0 takes 2 arguments, but it only takes 1. More values are uncovered than needed
callsub factorial_0
cover 2
store 2
store 1
retsub
```

jasonpaulos:
This PR fixes the bug described by issue #233. Now, the `numArgs` variable is properly set to the number of arguments in the called subroutine, not the caller subroutine.

Unit tests have been added to make sure this change works in TEAL versions 4 and 5 (before and after the `cover` and `uncover` opcodes were introduced).

Given this affected PyTeal program:

```python
from pyteal import *

@Subroutine(TealType.uint64)
def factorial(i: Expr) -> Expr:
    return If(i <= Int(1), Int(1), factorial_intermediate(i - Int(1), Bytes("inconsequential")) * i)

@Subroutine(TealType.uint64)
def factorial_intermediate(i: Expr, j: Expr) -> Expr:
    return Seq(Log(j), factorial(i))

program = Return(factorial(Int(4)) == Int(24))
```

Here's a diff of the generated TEAL before and after this PR:

```diff
 #pragma version 5
 int 4
 callsub factorial_0
 int 24
 ==
 return
 
 // factorial
 factorial_0:
 store 0
 load 0
 int 1
 <=
 bnz factorial_0_l2
 load 0
 int 1
 -
 byte "inconsequential"
 load 0
-swap
+cover 2
 callsub factorialintermediate_1
 swap
 store 0
 load 0
 *
 b factorial_0_l3
 factorial_0_l2:
 int 1
 factorial_0_l3:
 retsub
 
 // factorial_intermediate
 factorialintermediate_1:
 store 2
 store 1
 load 2
 log
 load 1
 load 1
 load 2
-uncover 3
-uncover 3
+uncover 2
 callsub factorial_0
 cover 2
 store 2
 store 1
 retsub
```

Closes #233

### DIFF
--- a/pyteal/compiler/subroutines.py
+++ b/pyteal/compiler/subroutines.py
@@ -106,7 +106,14 @@ def spillLocalSlotsDuringRecursion(
             before: List[TealComponent] = []
             after: List[TealComponent] = []
 
-            reentrySubroutineCalls = reentryPoints.intersection(stmt.getSubroutines())
+            calledSubroutines = stmt.getSubroutines()
+            # the only opcode that references subroutines is callsub, and it should only ever
+            # reference one subroutine at a time
+            assert (
+                len(calledSubroutines) <= 1
+            ), "Multiple subroutines are called from the same TealComponent"
+
+            reentrySubroutineCalls = list(reentryPoints.intersection(calledSubroutines))
             if len(reentrySubroutineCalls) != 0:
                 # A subroutine is being called which may reenter the current subroutine, so insert
                 # ops to spill local slots to the stack before calling the subroutine and also to
@@ -114,12 +121,9 @@ def spillLocalSlotsDuringRecursion(
                 # reentry into the current subroutine from modifying variables we are currently
                 # using.
 
-                # the only opcode that references subroutines is callsub, and it should only ever
-                # reference one subroutine at a time
-                assert (
-                    len(reentrySubroutineCalls) == 1
-                ), "Multiple subroutines are called from the same op"
-                reentrySubroutineCall = reentrySubroutineCalls.pop()
+                # reentrySubroutineCalls should have a length of 1, since calledSubroutines has a
+                # maximum length of 1
+                reentrySubroutineCall = reentrySubroutineCalls[0]
                 numArgs = reentrySubroutineCall.argumentCount()
 
                 digArgs = True
--- a/pyteal/compiler/subroutines.py
+++ b/pyteal/compiler/subroutines.py
@@ -94,7 +94,6 @@ def spillLocalSlotsDuringRecursion(
 
     for subroutine, reentryPoints in recursivePoints.items():
         slots = list(sorted(slot for slot in localSlots[subroutine]))
-        numArgs = subroutine.argumentCount()
 
         if len(reentryPoints) == 0 or len(slots) == 0:
             # no need to spill slots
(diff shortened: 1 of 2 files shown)


tzaffi (approved):
Good catch and LGTM.

pyteal/compiler/subroutines.py
```diff
@@ -107,13 +106,22 @@ def spillLocalSlotsDuringRecursion(
...
                 # restore the local slots after returning from the subroutine. This prevents a
                 # reentry into the current subroutine from modifying variables we are currently
                 # using.
 
+                # the only opcode that references subroutines is callsub, and it should only ever
+                # reference one subroutine at a time
+                assert (
+                    len(reentrySubroutineCalls) == 1
+                ), "Multiple subroutines are called from the same op"
+                reentrySubroutineCall = reentrySubroutineCalls.pop()
```
michaeldiamant, line 122: @jasonpaulos For my edification:  Have you considered the following changes to catch the highlighted problem earlier?
* In `TealOp` constructor, validate maximum of 1 `SubroutineDefinition` provided or raise error.
* Change `TealOp.getSubroutines` to something like `Teal.subroutine() -> Optional[SubroutineDefinition]`.

I'm trying to gauge:
* Does the proposed change presents a technical barrier?  From my cursory review, it seems like _no_.
* If no technical barrier, I'm wondering if you had considered the change, but elected for a more smaller scoped fix?  I'm on-board with resolving the bug ASAP though I'm also interested in potential follow on PRs to any agreed upon improvements.

----

Additionally, presuming we keep as shown:  Can you consider the following concern and let me know if you agree with the proposed change?
* PyCharm shows `reentrySubroutineCalls` is a `set`.  https://docs.python.org/3.6/library/stdtypes.html#frozenset.pop shows`set.pop()` removes an _arbitrary_ element.
* While asserting `len(reentrySubroutineCalls) == 1`, `pop()` usage is _safe_. 
* My concern is:
  * I think the approach can be made more robust with moderate effort.  I think it's unlikely we'll recall the Set semantics if/when the assertion is relaxed.
  * There are probably multiple ways to improve robustness and I'm flexible here.  One approach is to convert `reentrySubroutineCalls` to a `list`.

michaeldiamant: @jasonpaulos Thanks for the change in https://github.com/algorand/pyteal/pull/234/commits/07e48b6f5fe4d296b2979ec157028ea00907302f.  

The commit solves my immediate concern and I'll approve the PR.  Though I hold the thread open in case you'd like to add commentary about asserting earlier.

michaeldiamant (approved):
@jasonpaulos Thanks for the effort to document and resolve a sneaky problem!

pyteal/compiler/subroutines.py
michaeldiamant: @jasonpaulos Gotcha - Thanks for the added context.  I have additional questions that I'll can pursue out-of-bands to the PR.  Resolving.

PR #236, opened 2022-03-07
Wide Arithmetic Support (128bit uint)

### LINKED ISSUE #101: Support for wide (128-bit) arithmetic
## Summary
TEAL has the opcodes `mulw`, `addw`, and `divmodw`, which can be used to multiply, add, and divide 64-bit unsigned integers without worrying about possible overflow. They work by using two 64-bit integers to represent a single 128-bit integer on the stack.

PyTeal could expose these opcodes in order to create convenience methods for math operations whose input and output are 64-bit integers, but whose intermediate values might be as large as a 128-bit integer. (By this I mean we could make it easier to express `A * B / C` where `A * B` might overflow a 64-bit int, but the final expression will not overflow. If you're looking for ways to directly handle integers larger than 64-bits, consider [byteslice arithmetic](https://pyteal.readthedocs.io/en/latest/arithmetic_expression.html#byteslice-arithmetic) instead.)

## Scope
Create a set of functions which internally use `mulw`, `addw`, and `divmodw` to perform multi-step arithmetic expressions on 64-bit unsigned integers. These could be:
* `WideRatio(numeratorFactors: List[Expr], denominatorFactors: List[Expr]) -> Expr`: calculate the product of `numeratorFactors` divided by the product of `denominatorFactors`, e.g. `(N_1 * N_2 * N_3 * ...) / (D_1 * D_2 * D_3 * ...)`
* `WideSummation(positiveTerms: List[Expr], negativeTerms: List[Expr]) -> Expr`: calculate the summation of all `positiveTerms` minus the summation of all `negativeTerms`, e.g. `(P_1 + P_2 + P_3 + ...) - (N_1 + N_2 + N_3 + ...)`

Maybe there's a need for expressions mixing addition and division too, possibly of the form `(A + B) / C`?

## Priority

Not urgent because it's possible to use byteslice arithmetic in place of this (e.g. `Btoi(BytesDiv(BytesMul(Itob(Int(1)), Itob(Int(2))), Itob(Int(3))))` expresses `1 * 2 / 3` in a way that's safe from overflow). However, exposing support for `divmodw` and friends would be more efficient in terms of opcode cost.

ahangsu:
## Summary

Provides a set of uint128 (essentially two uint64 values on stack) arithmetic support from 5 wide arithmetic opcodes:
- `addw`
- `mulw`
- `divw`
- `divmodw`
- `expw`

Closes #101 

## Testing

Pending, currently only local testing, waiting on e2e semantics test.


pyteal/ast/widemath.py
```diff
@@ -157,3 +214,45 @@ def has_return(self):
...
+        return WideUint128(
+            Op.addw, [TealType.uint64, TealType.uint64], args=list(terms)
+        )
+
+    @staticmethod
+    def mulw(*factors: Expr):
+        if len(factors) < 2:
+            pass
+        return WideUint128(
+            Op.addw, [TealType.uint64, TealType.uint64], args=list(factors)
```
michaeldiamant, line 233: @ahangsu Friendly pointer to what I believe to be a typo.

```suggestion
            Op.mulw, [TealType.uint64, TealType.uint64], args=list(factors)
```

pyteal/ast/widemath.py
```diff
@@ -157,3 +214,45 @@ def has_return(self):
...
+class WideUint128(MultiValue):
+    @staticmethod
+    def addw(*terms: Expr):
+        if len(terms) < 2:
+            pass
+        return WideUint128(
+            Op.addw, [TealType.uint64, TealType.uint64], args=list(terms)
+        )
+
+    @staticmethod
```
michaeldiamant, line 228: @ahangsu For my edification as a new Python programmer:  It seems like you're providing factory methods where annotating with `@staticmethod`.  So that I know how style logic I write, is `@staticmethod` a de facto way to declare factories?

Brief googling pointed me towards `@classmethod` usage:
* https://realpython.com/python-multiple-constructors/#providing-multiple-constructors-with-classmethod-in-python discusses the idea.
* I also found https://stackoverflow.com/questions/136097/difference-between-staticmethod-and-classmethod, which points to https://stackoverflow.com/a/1950927.
* From the reading, it seems to me the primary difference with `@classmethod` is to support factory method construction in subclasses.  Whether it's advantageous to provide subclass factories probably depends on the use case.

pyteal/ast/widemath.py
```diff
@@ -157,3 +214,45 @@ def has_return(self):
...
+    def expw(base: Expr, _pow: Expr):
+        return WideUint128(
+            Op.expw, [TealType.uint64, TealType.uint64], args=[base, _pow]
+        )
+
+    def __init__(self, op: Op, types: List[TealType], *, args: List[Expr]):
+        super().__init__(op, types, args=args)
+
+    def __add__(self, other: Expr):
```
michaeldiamant, line 245: @ahangsu Clarification question:  How did you decide which intrinsic operators from https://docs.python.org/3/library/operator.html to override here?

I'm trying to gauge if there's a rationale for the subset overridden.  Or if possibly the implementation is partial because the PR is WIP.

pyteal/ast/widemath.py
```diff
@@ -157,3 +439,515 @@ def has_return(self):
...
+                return "(divW {} {})".format(self.lhs, self.term)
+
+        return WideUint128DivU128(self, other)
+
+    def divU64(self, other: Expr):
+        require_type(other, TealType.uint64)
+
+        class WideUint128DivU64(WideUint128):
+            def __init__(self, lhs: WideUint128, arg: Expr):
+                WideUint128.__init__(self, 1)
```
michaeldiamant, line 635: @ahangsu Request for clarification:  In the following scenarios, I _think_ programs generated via PyTeal do _not_ produce the intended on-chain results.  It's quite possible I have gaps in my understanding.  Can you double check me?
* Use `WideUint128DivU64` for subsequent `WideUint128` arithmetic (e.g. `+`).
  * Concern:  The result is 1 uint64.  Subsequent arithmetic operations assume a `WideUint128` produces 2 outputs (2 uint64).
* Produce `WideUint128` with `outputNum=2`, and supply the value as argument to `@Subroutine`.
  * Concern:  I'm concerned only 1 of 2 stack values will be passed into the subroutine.  I didn't fully vet the concern before raising.  Trying to gauge if you've considered it before I go deeper.

After addressing the above concerns, taking a step back:  
* Is the intent to provide a general purpose wide arithmetic API or to _only_ provide the operations needed to use wide arithmetic operations for intermediate steps? 
* It feels to me like the PR attempts to provide a general purpose wide arithmetic API.  Though it's unclear to me that the goal is a general purpose API.

pyteal/ast/widemath.py
```diff
@@ -1,20 +1,262 @@
...
+            # stack: [..., lowword(B + D), A, C, highword(B + D)]
+            TealOp(expr, Op.cover, 3),
+            # stack: [..., lowword(B + D), highword(B + D), A, C]
+            TealOp(expr, Op.cover, 2),
+            # stack: [..., lowword(B + D), highword(B + D), A + C]
+            TealOp(expr, Op.add),
+            # stack: [..., lowword(B + D), highword(B + D) + A + C]
+            TealOp(expr, Op.add),
+            # stack: [..., highword(B + D) + A + C, lowword(B + D)]
+            TealOp(expr, Op.swap),
```
PabloLION, line 100: It seems line 88 and line 94 are not needed here.

PR #238, opened 2022-03-08
Refactor Uint8.set to make data size a first class concept

michaeldiamant:
Optionally proposes a refactoring of `Uint8.set` to isolate the storage data size as a first class concept.

Rationale:
* `Uint8` is my introduction to PyTeal ABI support.  I found myself spending _some_ time understanding `Uint8.set` implementation and incrementally refactored for my comprehension.  It's quite possible that I'm _just_ unfamiliar with the space.  
* In case folks feel it's a sufficient improvement, I thought I'd present the PR while the content is fresh in my mind.  If we'd prefer to keep as is, I'll happily close the PR.

Notes:
* I didn't extend the change to other `Uint` subclasses because it's unclear we want the change.  _If_ we want the change, I can extend the concept in a follow on PR.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,3 +3,4 @@ mypy==0.931
 pytest
 pytest-timeout
 py-algorand-sdk
+types-dataclasses
--- a/pyteal/ast/abi/sized.py
+++ b/pyteal/ast/abi/sized.py
@@ -1,0 +1,24 @@
+from dataclasses import dataclass
+from typing import Union
+from .. import Expr
+from ..assert_ import Assert
+from pyteal.ast import Int, Seq, ScratchVar
+
+
+@dataclass(frozen=True)
+class SizedExpr:
+    """Provides a wrapper class to tag `Expr`s with known ABI data sizes.
+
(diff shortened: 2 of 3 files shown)


requirements.txt
```diff
@@ -3,3 +3,4 @@ mypy==0.931
 pytest
 pytest-timeout
 py-algorand-sdk
+types-dataclasses
```
michaeldiamant, line 6: Note for reviewers:  Without installing `types-dataclasses`, Python 3.6 lacks `@dataclass` support.

Here's the error prior to installing the dependency:  https://github.com/algorand/pyteal/runs/5471035129?check_suite_focus=true.

ahangsu (commented):
I am leaning towards not to add this, here are the reasons:
- Seems Jason has a PR after the first version of ABI, this PR might induce merge conflict.
- The introduction of `SizedExpr` is only for `Uint`, which is not a helpful abstraction can be extended to other implementation, and it somewhat obfuscates `Expr` and `SizedExpr`.
- This leads to another pip package to be imported for backwards-compatability.

Issue #244, opened 2022-03-11
Upgrade minimum Python version to 3.10


michaeldiamant:
## Problem

Though PyTeal minimally supports Python 3.6,  on-going effort (https://github.com/algorand/pyteal/pull/222) requires access to Python 3.8.  

Since _a_ version bump is required, we'd like to use the opportunity to move to the newest minor version:  3.10 (released 10/2021).

* Rationale:  Newer language facilities simplify maintaining PyTeal.
* Underlying assumption:  It's straightforward for PyTeal developers to configure a functioning PyTeal environment by configuring a project-specific Python version. 

## Solution

We'll need to do the following:
* Provide advance community notice about the upcoming change.
* Update the build to reflect the change in minimum Python version.
* For PyTeal developer convenience, document how to configure a project-specific Python version.

## Dependencies

Provide _sufficient_ notice for the community to hear about the pending change.

## Urgency

Minimally bumping to Python 3.8 is needed to release https://github.com/algorand/pyteal/issues/185.

algochoi:
We should also try to update the `black` version to `22.3.0` - it is the latest stable version (we pinned it down to a beta version before this) and should reformat our code with it.

PR #247, opened 2022-03-16
Tools : PyTeal optimization utility

algoidurovic:
This PR implements the first iteration of the PyTeal optimization utility. The first optimization to be included is the removal of contiguous store/load opcodes that are deemed redundant.

**Background**:
PyTeal often uses ScratchSlots implicitly to bind sections of the source code together. This pattern is best illustrated by the implementations of PyTeal subroutines and MultiValue expressions. The issue arises at the TEAL level when ScratchSlots are utilized naively, leading to redundant store/load opcodes (ie the ops cancel each other out) that can be removed without affecting the behavior of the program.

The conditions for applying this optimization are very simple -- and perhaps unnecessarily narrow:
1. candidate for removal must match the pattern `[store i, load i]` where i is the index of a scratch slot.
2. the same scratch slot cannot be loaded from anywhere else in the code.

**Testing**:
The unit tests cover both the optimization of hardcoded low level teal and the optimization of teal generated by the compiler from a PyTeal program. Tests cover subroutines, MultiValue expressions, ScratchVars, and more.

In order to gauge the practical utility of this optimization, I applied it to the [Algorand Auction Demo](https://github.com/algorand/auction-demo). Unfortunately, the scope of the optimization was too narrow to apply to the demo. Further investigation (market research?) is needed here to determine a fruitful direction to take the optimizer in the future.


pyteal/compiler/optimizer/optimizer.py
```diff
@@ -0,0 +1,73 @@
...
+        if not has_load_dependencies(teal, cast(int, op.args[0]), i + 1):
+            remove[i] = remove[i + 1] = True
+
+    return [op for i, op in enumerate(teal) if not remove[i]]
+
+
+def apply_local_optimizations(
+    teal: List[TealComponent], options: OptimizeOptions
+) -> List[TealComponent]:
+    while True:
```
michaeldiamant, line 59: @algoidurovic Though it feels unlikely, presumably it's possible optimizations will loop without end.  I'm wondering if you considered setting an upper bound for the number of optimization passes to attempt?

One possibility is to change `iterate_optimizations` from boolean to the maximum pass count.

pyteal/compiler/optimizer/optimizer_test.py
```diff
@@ -0,0 +1,363 @@
...
+        [
+            If(Txn.sender() == Global.creator_address()).Then(Pop(add(Int(1), Int(2)))),
+            Approve(),
+        ]
+    )
+
+    optimize_options = OptimizeOptions()
+
+    # unoptimized
+    expected = """#pragma version 4
```
michaeldiamant, line 106: @algoidurovic Have you considered parameterizing the tests by TEAL version?

For the PR's optimization, it seems unlikely that different TEAL versions produce different outputs.  Nevertheless, it feels to me that we ought to be able to certify optimizations for all supported versions.

pyteal/compiler/optimizer/optimizer.py
```diff
@@ -0,0 +1,73 @@
+from typing import List, cast
```
michaeldiamant, line 1: @algoidurovic Minor and looking to test if there's a convention we wish to uphold:  When defining packages, do we try to avoid circular dependencies?

* I'm noting `pyteal/compiler/compiler.py` refers to a _deeper_ package:  `pyteal/compiler/optimizer/optimizer.py`.
* Typically, to minimize the chance for a circular dependency forming, I'd aim to keep communication flowing from less-nested-to-more-deeply-nested packages.
* In this case, _if_ we wish to enforce a dependency flow, possible actions include removing the `optimizer` package or moving `compiler` to a subpackage.
* I'll follow the group's convention.  I raise the question because I've had experiences where later refactorings can be complicated by circular dependencies.

pyteal/compiler/optimizer/optimizer.py
```diff
@@ -0,0 +1,73 @@
...
+        if i == pos:
+            continue
+
+        if type(op) == TealOp and op.op == Op.load and op.args[0] == slot:
+            return True
+
+    return False
+
+
+def apply_slot_to_stack(teal: List[TealComponent]) -> List[TealComponent]:
```
michaeldiamant, line 37: @algoidurovic Optional - Feel welcomed to resolve as is:  I'd optionally consider prefixing `apply_slot_to_stack` and `has_load_dependencies` with `_` to denote private methods.

From what I can tell, these methods are _not_ intended to be generally available.

pyteal/compiler/optimizer/optimizer.py
```diff
@@ -0,0 +1,73 @@
...
+        if op.args[0] != next_op.args[0]:
+            continue
+
+        if not has_load_dependencies(teal, cast(int, op.args[0]), i + 1):
+            remove[i] = remove[i + 1] = True
+
+    return [op for i, op in enumerate(teal) if not remove[i]]
+
+
+def apply_local_optimizations(
```
michaeldiamant, line 56: @algoidurovic Optional and minor:  I'm wondering if the method name can be made more specific?

* I think _local optimization_ can be read ambiguously (e.g. _Local to what?_).
* I propose either renaming the method or folding into `apply_optimizations`.   Since only one kind of optimization exists, it might be easier to defer adding more structure today.

pyteal/compiler/optimizer/optimizer.py
```diff
@@ -0,0 +1,73 @@
...
+from pyteal.ir.ops import Op
+
+
+class OptimizeOptions:
+    """An object which specifies optimizations to be performed."""
+
+    use_slot_to_stack: bool
+    iterate_optimizations: bool
+
+    def __init__(
```
michaeldiamant, line 14: @algoidurovic Optional and minor:  If you'd prefer to save the boilerplate of initializing `OptimizeOptions`, I think we can begin using https://docs.python.org/3/library/dataclasses.html.

We've already discussed picking up the 3.10 version bump (https://github.com/algorand/pyteal/issues/244).  So, if you prefer it, I don't see harm in using a 3.7 feature.

pyteal/compiler/optimizer/optimizer.py
algoidurovic: The loop stops iterating when the optimization no longer changes the code, so it's guaranteed to terminate eventually.

michaeldiamant: Leaving a note to document a verbal discussion with @algoidurovic:
* We discussed that a number of possible outcomes may cause infinite looping.  For example, a mistake in optimization algorithm, TEAL programs written in unexpected ways reaching compilation optimization, etc.
* Rather than use a fixed ceiling to bound iterations, we agreed to bound iteration count = program length.  The suggestion avoids a static, value fixed for all programs while also precluding infinite loops.

pyteal/compiler/compiler.py
```diff
@@ -1,4 +1,7 @@
 from typing import List, Tuple, Set, Dict, Optional, cast
+from pyteal.compiler import optimizer
```
michaeldiamant, line 2: @algoidurovic Unrelated to the line:  Since we're mostly agreed on the optimization to-be-provided, can you update the user guide?

Nominally, https://pyteal.readthedocs.io/ ought to explain how users can opt-in for optimization.

pyteal/compiler/compiler.py
```diff
@@ -181,6 +187,7 @@ def compileTeal(
             constants will be assembled in the most space-efficient way, so enabling this may reduce
             the compiled program's size. Enabling this option requires a minimum TEAL version of 3.
             Defaults to false.
+        optimize (optional): todo
```
michaeldiamant, line 190: @algoidurovic Friendly visual reminder to fill in the TODO.

pyteal/compiler/optimizer/optimizer.py
```diff
@@ -0,0 +1,71 @@
...
+from pyteal.ir.ops import Op
+
+
+@dataclass
+class OptimizeOptions:
+    """An object which specifies the optimizations to be performed.
+    Args:
+        useSlotToStack (optional): cancel contiguous store/load operations
+            that have no load dependencies elsewhere.
+        iterateOptimizations (optional): repeat the optimizations until the
```
michaeldiamant, line 15: @algoidurovic Minor:  Propose making docs match casing of the source code.

```suggestion
        iterate_optimizations (optional): repeat the optimizations until the
```

pyteal/compiler/optimizer/optimizer.py
```diff
@@ -0,0 +1,71 @@
...
+from pyteal.ir.tealcomponent import TealComponent
+from pyteal.ir.tealop import TealOp
+from pyteal.ir.ops import Op
+
+
+@dataclass
+class OptimizeOptions:
+    """An object which specifies the optimizations to be performed.
+    Args:
+        useSlotToStack (optional): cancel contiguous store/load operations
```
michaeldiamant, line 13: @algoidurovic Minor:  Propose making docs match casing of the source code.

```suggestion

        use_slot_to_stack (optional): cancel contiguous store/load operations
```

pyteal/compiler/optimizer/optimizer_test.py
michaeldiamant: @algoidurovic That's a reasonable response.  I'm on-board to keep as is without exercising a matrix of TEAL versions.

pyteal/compiler/optimizer/optimizer.py
michaeldiamant: @algoidurovic Thanks for bringing in the feedback via https://github.com/algorand/pyteal/pull/247/commits/a9ba3efd9fea25cab76e96c7d100fe61d120f578.

pyteal/compiler/optimizer/optimizer.py
algoidurovic: Quoting Jason:
"The approach PyTeal currently uses is below. Here a module means a directory of source code where all objects meant for external use are exported by the __init__.py source file in that directory.
When importing an object from the same module or a parent module, import the full path of the source file that defines the object.
When importing an object from a child module, import the entire module folder and access the desired object from there. Do not directly import the file that defines the object.
With this approach, circular dependencies can happen in two ways:
**Intra-module circular dependencies**, e.g. m1/a.py imports m1/b.py which imports m1/a.py. When this happens, normally one of the files only needs to import the other to implement something, so the import statements can be moved from the top of the file to the body of the function that needs them. This breaks the import cycle.
**Inter-module circular dependencies**, e.g. m1/a.py imports m1/m2/__init__.py which imports m1/m2/x.py which imports m1/a.py. To avoid this, make sure that any objects/files in m1 that need to be exposed to deeper modules do not rely on any objects from that deeper module. If that isn’t possible, then perhaps m1/a.py belongs in the m1/m2/ module"

pyteal/compiler/compiler.py
algoidurovic: good catch, fixed.

pyteal/compiler/compiler.py
```diff
@@ -1,4 +1,7 @@
 from typing import List, Tuple, Set, Dict, Optional, cast
+from pyteal.compiler import optimizer
+
+from pyteal.compiler.optimizer import OptimizeOptions, apply_optimizations
```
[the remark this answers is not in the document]
algoidurovic: fixed

pyteal/compiler/optimizer/optimizer.py
```diff
@@ -0,0 +1,72 @@
...
+class OptimizeOptions:
+    """An object which specifies the optimizations to be performed.
+    Args:
+
+        use_slot_to_stack (optional): cancel contiguous store/load operations
+            that have no load dependencies elsewhere.
+        iterate_optimizations (optional): repeat the optimizations until the
+            code is no longer changed."""
+
+    use_slot_to_stack: bool = False
```
jasonpaulos, line 19: I have two suggestions:
* Can you give `use_slot_to_stack` a more general name? In the future it might not only move variables from scratch slots to the stack, it could also reduce the number of loads and stores or do something more advanced. Maybe just `scratch_slots`? How optimizations will evolve is a bit of an open question, but right now I think `OptimizeOptions` should only have fields about general categories of optimizations, and in the future when we have more specific options for each category, we might want to change the value from a boolean to _another_ options objects just for that category of optimizations.

* Why should `iterate_optimizations` be exposed? Is there a scenario where you would want to disable this?

pyteal/compiler/optimizer/optimizer.py
```diff
@@ -0,0 +1,72 @@
...
+        if type(next_op) != TealOp or next_op.op != Op.load:
+            continue
+
+        if op.args[0] != next_op.args[0]:
+            continue
+
+        if not _has_load_dependencies(teal, cast(int, op.args[0]), i + 1):
+            remove[i] = remove[i + 1] = True
+
+    return [op for i, op in enumerate(teal) if not remove[i]]
```
jasonpaulos, line 50: nit: it would be more space-efficient to make `remove` a set of indexes rather than a list that's always as long as the program.

pyteal/compiler/optimizer/optimizer.py
```diff
@@ -0,0 +1,72 @@
...
+            teal = _apply_slot_to_stack(teal)
+
+        if not options.iterate_optimizations or prev_teal == teal:
+            break
+
+    return teal
+
+
+# Currently, only local optimizations are available so this is equivalent to
+# calling applyLocalOptimizations().
```
jasonpaulos, line 70: nit: outdated comment

pyteal/compiler/optimizer/optimizer_test.py
```diff
@@ -0,0 +1,363 @@
+from pickletools import optimize
```
jasonpaulos, line 1: nit: unnecessary import

jasonpaulos (commented):
I have one concern that must be addressed before releasing this:
1. The optimization looks correct except for the case where a user makes a `ScratchVar` with a specifically requested slot ID. In that case, we need to respect that slot ID and not optimize the scratch slot away, so this optimization should likely not touch any `ScratchSlot` with a user-defined ID. Based on the current code, this is not checked.
    * Item 1 from the next list would be one way to solve this, since if you operate on `ScratchSlot` objects you can tell if they're user-defined or not.

And a few that need not be addressed now, but should be before more optimization work takes place. We can talk about these in more detail next week if you'd like.
1. This optimization happens _after_ `ScratchSlot` objects have been assigned integer IDs. This step of the compilation will fail if more than 256 scratch slots are in use, so it would be more useful if this optimization can be applied before that happens. I realize the flow of the compiler as it is currently doesn't make this easy, but it should be possible to reorder compilation events to make this happen.
3. Currently this optimization happens on the linearized list of `TealComponent`s. For this simple optimization, that's acceptable, but anything more complex will need access to more control flow information than is available at this stage and should probably operate with the control flow graph. For that reason I'd recommend changing this optimization to also operate on `TealBlock`s before we implement more advanced features.

pyteal/compiler/optimizer/optimizer.py
algoidurovic: I agree with the first point. In regards to the second, this was mainly a conservative decision to allow for fine grained control in potential cases where iteration would be undesirable, although I don't know of any at the moment. I realize that trying to plan for all contingencies isn't the best argument so I don't mind removing the flag.

pyteal/compiler/compiler.py
```diff
@@ -167,6 +172,7 @@ def compileTeal(
     *,
     version: int = DEFAULT_TEAL_VERSION,
     assembleConstants: bool = False,
+    optimize: OptimizeOptions = OptimizeOptions(),
```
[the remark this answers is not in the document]
algoidurovic: Good point, I didn't know this about python. I'll change it to None.

pyteal/compiler/optimizer/optimizer.py
michaeldiamant: @algoidurovic Unresolving for a clarification:  

From the above note, I understood `pyteal/compiler/compiler.py` is liable to create _inter-module circular dependencies_ with `pyteal/compiler/optimizer/optimizer.py`.   Does that match your understanding?

If so, are we leaving the package structure unchanged because no issue exists today or was the thread unintentionally resolved?

pyteal/compiler/compiler.py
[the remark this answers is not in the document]
michaeldiamant: Leaving as a reference, not a change request:  It _is_ possible to make the object immutable because it's a `dataclass`.  From https://docs.python.org/3/library/dataclasses.html:

> frozen: If true (the default is False), assigning to fields will generate an exception. This emulates read-only frozen instances. If __setattr__() or __delattr__() is defined in the class, then [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError) is raised. See the discussion below.

pyteal/compiler/optimizer/optimizer.py
michaeldiamant: Summarizing a verbal with @algoidurovic that resolves the thread:  

We agreed there's no strong enforcement for circular dependencies today.  Since no circular dependency exists in the PR, we agreed to leave the structure as is.

pyteal/compiler/scratchslots.py
```diff
@@ -35,7 +35,7 @@ def collectScratchSlots(
 def assignScratchSlotsToSubroutines(
     subroutineMapping: Dict[Optional[SubroutineDefinition], List[TealComponent]],
     subroutineBlocks: Dict[Optional[SubroutineDefinition], TealBlock],
-) -> Dict[Optional[SubroutineDefinition], Set[int]]:
+) -> Tuple[Dict[Optional[SubroutineDefinition], Set[int]], Set[int]]:
     """Assign scratch slot values for an entire program.
```
michaeldiamant, line 39: @algoidurovic Minor:  Can you update the docstring to reflect the return type change?

pyteal/compiler/scratchslots_test.py
```diff
@@ -386,7 +386,7 @@ def sub3Impl(a1, a2, a3):
 
     # mainSlot2 and subroutine2Slot1 request the same ID, 100
     with pytest.raises(TealInternalError):
-        actual = assignScratchSlotsToSubroutines(subroutineMapping, subroutineBlocks)
+        actual, _ = assignScratchSlotsToSubroutines(subroutineMapping, subroutineBlocks)
```
michaeldiamant, line 389: @algoidurovic I didn't closely review the existing tests, but the diff suggests reserved IDs aren't tested (usage of `_`).  Can you modify/create a test to help build confidence the functionality works as intended?

pyteal/compiler/scratchslots.py
```diff
@@ -107,6 +108,7 @@ def assignScratchSlotsToSubroutines(
                 "Slot ID {} has been assigned multiple times".format(slot.id)
             )
         slotIds.add(slot.id)
+        reserved_ids.add(slot.id)
```
michaeldiamant, line 111: @algoidurovic Optional - I spent a few min refactoring here because I got tripped up on the usage of `continue`.  Here's a version that passes all tests with the added benefit of being able to list _all_ duplicate slot IDs (though I didn't update the error message).

```
    def find_reserved_slot_ids(slots: typing.Set[ScratchSlot]) -> typing.Set[int]:
        ids = [slot.id for slot in slots if slot.isReservedSlot]
        duplicates = [k for k, xs in itertools.groupby(sorted(ids)) if len(list(xs)) > 1]

        if len(duplicates) > 0:
            raise TealInternalError("Duplicate IDs assigned")

        return set(ids)

    reserved_ids = find_reserved_slot_ids(allSlots)
    slotIds = reserved_ids.copy()
```

Requires a few imports:

```
import typing
import itertools
```

If you feel it's worth taking in, you're welcome to use it.  Otherwise let's resolve the thread.

algoidurovic: Resolving because my recent commit changed this part of the code and I don't think the feedback is applicable anymore.

pyteal/compiler/optimizer/optimizer.py
```diff
@@ -0,0 +1,89 @@
...
+def apply_global_optimizations(start: TealBlock, options: OptimizeOptions) -> TealBlock:
+    # limit number of iterations to length of teal program to avoid potential
+    # infinite loops.
+    for block in TealBlock.Iterate(start):
+        prev_ops = block.ops.copy()
+        for _ in range(len(block.ops)):
+            if options.scratch_slots:
+                _apply_slot_to_stack(block, start, options.skip_ids)
+
+            if prev_ops == block.ops:
```
jasonpaulos, line 86: If this branch isn't true, I think you'll want to do `prev_ops = block.ops.copy()` again, otherwise you're not measuring the difference between iterations, you're measuring the difference from the beginning.

pyteal/compiler/optimizer/optimizer.py
```diff
@@ -0,0 +1,89 @@
...
+                the current application. For example, the 'gloads' opcode can
+                access the slots of other applications in the tx group.
+                2. global slots because they're outside the scope of global
+                optimizations, which only apply to the control flow graph of
+                a single subroutine.
+                3. slots used with dynamic scratch vars. These slots use
+                indirection by means of the 'stores' opcode and dependencies
+                can only be determined at runtime."""
+
+    def __init__(self, *, scratch_slots: bool = False, skip_ids: Set[int] = None):
```
jasonpaulos, line 26: My understanding of `skip_ids` is that it's an implementation detail that's not meant to be user facing. If that's the case, I think it should not be possible to set in the constructor, and it should probably be prefixed with `_` so users know they shouldn't use it directly.

I'm suggesting this because I think in the long term, it _should_ be possible to pass information like this to optimizations in a different way, and if we make that change it would be nice if we could remove the field from here.

pyteal/compiler/optimizer/optimizer.py
```diff
@@ -0,0 +1,89 @@
...
+            continue
+
+        next_op = cur_block.ops[i + 1]
+        if type(next_op) != TealOp or next_op.op != Op.load:
+            continue
+
+        if op.args[0] != next_op.args[0]:
+            continue
+
+        if not _has_load_dependencies(cur_block, start, cast(int, op.args[0]), i + 1):
```
jasonpaulos, line 71: Since this code has been moved before scratch slots are assigned IDs, instead of the argument being `int`, it will be a `ScratchSlot`. I don't think that changes the code much, but it definitely needs to chnage the type annotations.

pyteal/compiler/scratchslots.py
```diff
@@ -2,45 +2,99 @@
...
+    """
+
+    unoptimized_slots: Set[int] = set()
+
+    def collectSlotsFromBlock(block: TealBlock):
+        for op in block.ops:
+            for slot in op.getSlots():
+                # dynamic slot or reserved slot
+                if op.op == Op.int or slot.isReservedSlot:
+                    unoptimized_slots.add(slot.id)
```
jasonpaulos, line 32: I think it would be better to return `Set[ScratchSlot]` instead of `Set[int]`, since the integer ID is really an implementation detail of how slot equality is determined.

pyteal/compiler/optimizer/optimizer_test.py
```diff
@@ -0,0 +1,417 @@
+from .optimizer import OptimizeOptions
+
+from pyteal.ir.ops import Op
+
+from pyteal import *
```
jasonpaulos, line 5: nit: use relative import `.` instead of `pyteal`. Also, `Op` should already be included in the star import, no need to explicitly import it as well

pyteal/compiler/optimizer/optimizer_test.py
```diff
@@ -0,0 +1,417 @@
...
+
+import pytest
+
+
+def test_optimize_subroutine():
```
jasonpaulos, line 10: Could you also add a version of this test that uses a subroutine with a `ScratchVar` argument? I think there will still be things to optimize in that case.

pyteal/compiler/optimizer/optimizer.py
```diff
@@ -0,0 +1,89 @@
...
+
+            if type(op) == TealOp and op.op == Op.load and op.args[0] == slot:
+                return True
+
+    return False
+
+
+def _apply_slot_to_stack(cur_block: TealBlock, start: TealBlock, skip_ids: Set[int]):
+    slots_to_remove = set()
+    for i, op in enumerate(cur_block.ops[:-1]):
```
jasonpaulos, line 56: It's possible for some blocks to have 0 ops (an example is the program `If(Int(1)).Then(Approve()).Else(Reject())`, since the branch converges on an empty block), and if that's the case I think this would error.

A very simple unit test that an empty block won't cause this to fail would be nice to have.

pyteal/compiler/compiler.py
```diff
@@ -109,9 +117,9 @@ def verifyOpsForMode(teal: List[TealComponent], mode: Mode):
 def compileSubroutine(
     ast: Expr,
     options: CompileOptions,
-    subroutineMapping: Dict[Optional[SubroutineDefinition], List[TealComponent]],
     subroutineGraph: Dict[SubroutineDefinition, Set[SubroutineDefinition]],
     subroutineBlocks: Dict[Optional[SubroutineDefinition], TealBlock],
+    subroutineEndBlocks: Dict[Optional[SubroutineDefinition], TealBlock],
```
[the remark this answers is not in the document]
algoidurovic: I was going to do this originally but it requires changing a lot of code and rewriting tests in non trivial ways. If you think it's worth the effort then can you elaborate a little on why?

pyteal/compiler/optimizer/optimizer.py
algoidurovic: Totally agree.

pyteal/compiler/optimizer/optimizer.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def _has_load_dependencies(cur_block: TealBlock, start: TealBlock, slot: int, pos: int):
+    for block in TealBlock.Iterate(start):
+        for i, op in enumerate(block.ops):
+            if block == cur_block and i == pos:
+                continue
+
+            if type(op) == TealOp and op.op == Op.load and op.args[0] == slot:
+                return True
```
ahangsu, line 49: If I understand it correctly, this function checks the existence of load from `slot` other than specific block and operation position, right? 

We do not distinguish before/after execution of the `load` we are ignoring, am I correctly understanding it?

pyteal/compiler/compiler.py
```diff
@@ -1,5 +1,8 @@
 from typing import List, Tuple, Set, Dict, Optional, cast
 
+from pyteal.compiler.optimizer import OptimizeOptions
```
ahangsu, line 3: nit: relative import `from .optimizer import optimizer, OptimizeOptions`

pyteal/compiler/compiler.py
```diff
@@ -150,23 +158,38 @@ def compileSubroutine(
...
             subroutine.getDeclaration(),
             options,
-            subroutineMapping,
             subroutineGraph,
             subroutineBlocks,
+            subroutineEndBlocks,
         )
 
 
+def createTealMapping(
```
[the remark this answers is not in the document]
ahangsu: ^

pyteal/compiler/compiler.py
```diff
@@ -109,9 +117,9 @@ def verifyOpsForMode(teal: List[TealComponent], mode: Mode):
 def compileSubroutine(
     ast: Expr,
     options: CompileOptions,
-    subroutineMapping: Dict[Optional[SubroutineDefinition], List[TealComponent]],
     subroutineGraph: Dict[SubroutineDefinition, Set[SubroutineDefinition]],
     subroutineBlocks: Dict[Optional[SubroutineDefinition], TealBlock],
```
ahangsu, line 121: nit: would be great if you rename it `subroutine_start_blocks` (if I understand it correctly, this maps `SubroutineDefinition` to the `TealBlock` it starts?)

pyteal/compiler/compiler.py
```diff
@@ -199,20 +223,31 @@ def compileTeal(
...
-        subroutineMapping, subroutineBlocks
-    )
+    # note: optimizations are off by default, in which case, apply_global_optimizations
+    # won't make any changes. Because the optimizer is invoked on a subroutine's
+    # control flow graph, the optimizer requires context across block boundaries. This
+    # is necessary for the dependency checking of local slots. Global slots, slots
+    # used by DynamicScratchVar, and reserved slots are not optimized.
+    if options.optimize.scratch_slots:
+        options.optimize.skip_ids = collectUnoptimizedSlotIDs(subroutineBlocks)
+        print(str(options.optimize.skip_ids))
```
ahangsu, line 242: nit: do we need this line?

ahangsu (commented):
Just some minor comments here and there after a second pass on implementation

pyteal/compiler/optimizer/optimizer.py
algoidurovic: I think the inner for loop in `apply_global_optimizations` prevents calling `_apply_slot_to_stack` with an empty block. Surprisingly, even if an empty block were passed directly to `_apply_slot_to_stack`, it wouldn't cause an error because it's completely legal in python for slice indexes to be outside the bounds of the underlying list. I added tests to confirm this behavior.

pyteal/compiler/optimizer/optimizer.py
algoidurovic: Right, this is basically dumb, overly eager dependency checking. I'll add a comment to the function for clarification.

pyteal/compiler/optimizer/optimizer.py
jasonpaulos: Interesting, thanks for looking into this!

jasonpaulos (approved):
Looks good to me, thanks for addressing all concerns!

pyteal/compiler/optimizer/optimizer.py
```diff
@@ -0,0 +1,79 @@
...
+
+def _apply_slot_to_stack(
+    teal: List[TealComponent], reserved_ids: Set[int]
+) -> List[TealComponent]:
+    remove = set()
+    for i, op in enumerate(teal[:-1]):
+        if type(op) != TealOp or op.op != Op.store:
+            continue
+
+        # do not optimize away reserved slots
```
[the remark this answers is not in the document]
algoidurovic: This should be addressed now in the docstring for OptimizeOptions.

pyteal/compiler/scratchslots_test.py
algoidurovic: I think this is no longer a concern with the recent changes I've made. Resolving for now but feel free to unresolve.

docs/compiler_optimization.rst
```diff
@@ -0,0 +1,30 @@
...
+============================== ================================================================================ ===========================
+Optimization Flag              Description                                                                      Default
+============================== ================================================================================ ===========================
+:code:`scratch_slots`          A boolean describing whether or not scratch slot optimization should be applied. :code:`False`
+============================== ================================================================================ ===========================
+
+.. code-block:: python
+
+    optimize_options = OptimizeOptions(scratch_slots=True)
+    compileTeal(approval_program(), mode=Mode.Application, version=4, optimizer=optimize_options)
```
jasonpaulos, line 30: I think the keyword argument should be `optimize` not `optimizer`?

docs/compiler_optimization.rst
```diff
@@ -0,0 +1,30 @@
...
+to: compiled code size, scratch slot usage, opcode cost, etc. 
+
+Optimizer Usage
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The compiler determines which optimizations to apply based on the provided :any:`OptimizeOptions` object as
+shown in the code block below. The :any:`OptimizeOptions` constructor receives a set of keyword arguments 
+representing flags corresponding to particular optimizations. If arguments are provided to the
+constructor or no :any:`OptimizeOptions` object is passed to :any:`compileTeal` then the default behavior is
+that no optimizations are applied.
```
jasonpaulos, line 19: I think you mean if arguments are _not_ provided to the constructor?

Issue #214, opened 2022-03-01
Semantic E2E Tests


tzaffi:
## Problem

When creating new PyTeal constructs (e.g. #198, #196), or modifying existing ones, there are currently no tests to validate that resulting constructs are _semantically_ correct. We have end to end tests that validate that generated TEAL is as expected, but it would be even better to validate that the generated TEAL has the intended meaning.

## Solution

Introduce Dry-run capabilities into the e2e tests so that generated TEAL can be run and results be validated.

## Dependencies

None

## Urgency

Medium. I think this fills in a gap in the current PyTEAL development cycle.

tzaffi:
Added #179 as a blocker, as the current PR #249 as it currently stands, requires a change to `@Subroutine`'s interface, and this could conflict with what ABI ends up doing (and which has a higher priority)

Issue #178, opened 2022-02-01
Support `TealType` subroutine annotations


jasonpaulos:
## Summary

Some developers use `TealType` annotations in subroutines to declare the expected type of an argument or return value (uint64, bytes, or any). For example:

```python
from pyteal import *

@Subroutine(TealType.uint64)
def mySubroutine(a: TealType.uint64, b: TealType.uint64) -> TealType.uint64:
    return a + b
```

Unfortunately these types are not truthful -- the actual arguments and return values must be `Expr`s. So it would be beneficial to be able to express the same constraints, but in a way that preserves the actual type information.

## Scope
Somehow allow `TealType` declarations in a way that does not break the actual Python types.

* One possible solution to this would be to make `Expr` a generic type that takes a `TealType` as a parameter, such as `Expr[TealType.uint64]`. (Note that this would likely require changes to all subclasses of `Expr` so that they can properly handle the fact that it is now a generic type.) Then the example could turn into:
```python
@Subroutine(TealType.uint64)
def mySubroutine(a: Expr[TealType.uint64], b: Expr[TealType.uint64]) -> Expr[TealType.uint64]:
    return a + b
```

Regardless of how the `TealType` of arguments and return values are declared, the compiler should also perform type checking every time a subroutine is invoked to verify that the argument being passed to it conform to declared types.

tzaffi:
<img width="744" alt="image" src="https://user-images.githubusercontent.com/291133/159812986-b70b0efa-60d4-47c9-93b3-f40a433948cf.png">

Minor point: with this new expression available, I think we should deprecate providing the return type in the decorator's parameter, as it could conflict with the return type, and makes for harder to understand subroutines definition. I.e., we should be promoting the following decorator usage (notice the more python'ic un-capped usage as well):

```python
@subroutine
def mySubroutine(a: Expr[TealType.uint64], b: Expr[TealType.uint64]) -> Expr[TealType.uint64]:
    return a + b
```

PR #249, opened 2022-03-18
Semantic Blackbox Testing

tzaffi:
# Subroutine Blackbox - E2E Testing using the TEAL Blackbox Toolkit

[Graviton](https://github.com/algorand/graviton) introduces the TEAL Blackbox Toolkit.

This PR extends the functionality to PyTEAL. In particular:

1. Start with a pure `@Subroutine` and stack a `@Blackbox` decorator on top to specifying its `input_types` 
2. Generate an approval program for that subroutine using `blackbox_pyteal()` that is ready to compile to either a logic sig or an app
3. Use the TEAL Blackbox Toolkit to make semantic assertions and generate reports

See [graviton_test examples 1-3](https://github.com/algorand/pyteal/blob/853d7c09b01d1b637a70334691b40abb11bf8133/tests/integration/graviton_test.py#L581) for detailed usage examples.

## Summary of Changes
* `.github/workflows/build.yml` - introduce [MakerXStudio sandbox dev repo's](https://github.com/MakerXStudio/algorand-sandbox-dev) prebuilt [docker image](https://hub.docker.com/r/makerxau/algorand-sandbox-dev) to allow for integration tests in github actions (and this is _blazingly fast_ - 2 minutes with caching)
* `Makefile` - adding:
  * `make integration-test`
  * `local-gh-simulate`
  * `sandbox-dev-up/down`
* `README.md` - Instructions for running integration test. Additionally, explain how to use the dockerhub `algod` image.
* new file `docker-compose.yml` (derived from [MakerX](https://github.com/MakerXStudio/algorand-sandbox-dev)) that lets us quickly stand up an `algod` for integration tests on github
* `examples/signature/factorizer_game.py` - pyteal parametrized family of logic sig contract accounts. I used this in my last demo which created and executed more than 10K dry runs on 100+ logic sigs in around 3 minutes.
* `pyteal/ast/subroutine.py` - adding the ability to specify the `TealType` of `@Subroutine` parameters
* `setup.py` - requiring the following for development: 
  * graviton@🦙 - for blackbox testing
  * pytest-xdist - for asynchronous testing capability
  * tabulate - required by `graviton` when generating reports
* `tests/`
  * blackbox.py` - utility dry-running `@Subroutines`. In particular, an `algod_client()` is provided together with `blackbox_pyteal()` which takes as an input a `@Subroutine` + a `Mode` and outputs compilable `PyTeal` ready to dryrun, inspect and assert against.
  * `integration/` - introducing the integrations tests directory
    * `graviton_test.py` - **New integration test that shows how to graviton can be leveraged to gain confidence in pyteal correctness** 
    * `user_guide_test.py` - tests for examples that end up in the user guide including 3 graviton examples. This also shows how to leverage graviton, but using **a more straightforward approach than graviton_test.py**
  * `unit/` - new folder which contains the pure unit tests `compile_test.py`, `module_test.py`, `pre_v6_test.py`, and `pass_by_ref_test.py`
    * `blackbox_test.py` - companion _unit_ test for `blackbox.py`
    * `pre_v6_test.py` - new e2e test that freezes some pre-v6 pyteal's and asserts that their generated TEAL doesn't change
    * `pass_by_ref_test.py` - add more test asserting that illegal programs don't compile
  
### ANSWER: NO! (_in this case_)

While I have your attention, compare what happens 

### when miss the docker cache for the sandbox:
image

### compared to what happens when we don't:
image

## Action Items

- [x] `tests/blackbox.py::blackboxk_pyteal()`: function that converts a `@Subroutine` to an executable logic-sig or app program for blackbox testing
- [x] standing up an `algod` on github using new `docker-compose.yml`
- [x] add integration tests under `tests` and separate out unit tests as well
- [x] add dependency to `graviton` in `setup.py`
- [x] add `make` commands for blackbox testing and use these for github actions
- [x] bring doc tests back in sync with actual docs
- [x] decide what to name the new top-level directory `utils` &rarr; this is now folded under `tests`
- [x] decide whether to continue with the current approach for specifying `@Subroutine` input types, or go with another - we're going with a new `@Blackbox` per child PR #297 
- [x] decide if / how to generate docs for dryrun testing pyteal - we're _not_ going to generate these for now, and we won't even have an executable doc comment for `tests.blackbox.blackbox_pyteal()`
- [x] decide whether to cache Docker layers. Decision: NO 
- [x] add `SubroutineDefinititon.validate()` unit test in `subroutine_test.py`


.github/workflows/build.yml
```diff
@@ -19,16 +19,15 @@ jobs:
         uses: actions/checkout@v2
         with:
           fetch-depth: 0
-      - name: Install pip dependencies
```
tzaffi, line 22: refactor into Makefile for easier pre-commit checks

.gitignore
```diff
@@ -132,3 +133,6 @@ dmypy.json
 
 # IDE
 .idea
+
+# Comma Separated Value report files
```
tzaffi, line 137: blackbox generates CSV reports

pyteal/ast/scratchvar.py
```diff
@@ -97,6 +97,8 @@ def set_index(self, index_var: ScratchVar) -> Expr:
         Followup `store`, `load` and `index` operations will use the provided `index_var` until
         `set_index()` is called again to reset the referenced ScratchVar.
         """
+        # TODO: this type check can probably be changed to isinstance(..., ScratchVar) (cf. issues #232, #242)
+        # if not isinstance(index_var, ScratchVar): # requires further investigation
```
tzaffi, line 101:
```suggestion
```

pyteal/ast/subroutine.py
```diff
@@ -22,11 +22,12 @@ def __init__(
         implementation: Callable[..., Expr],
         returnType: TealType,
         nameStr: str = None,
+        input_types: List[TealType] = None,
```
tzaffi, line 25: Adding optional input types. I needed this when generating a pyteal program from a subroutine. This is because `Btoi()` needs to be applied to external `args` which must be converted to the appropriate type of the subroutine parameter.

tests/pre_v6_test.py
```diff
@@ -0,0 +1,88 @@
+from pyteal import *
```
tzaffi, line 1: Are these tests passing? If so, something is up with my other tests which generate more random slot assignments.

tests/semantic_asserts.py
```diff
@@ -0,0 +1,147 @@
...
+
+def get_blackbox_scenario_components(
+    scenario: Dict[blackbox.DryRunAssertionType, dict], mode: Mode
+):
+    return blackbox.get_blackbox_scenario_components(
+        scenario, mode_to_execution_mode(mode)
+    )
+
+
+def e2e_pyteal(subr: SubroutineFnWrapper, mode: Mode) -> Callable[..., Expr]:
```
tzaffi, line 61: Add an explanatory comment.

tests/semantic_test.py
```diff
@@ -0,0 +1,561 @@
...
+    dryrun_assert,
+    lightly_encode_args,
+    lightly_encode_output,
+    scratch_encode,
+)
+
+from pyteal import *
+
+# TODO: get tests working on Circle and set this to True
+SEMANTIC_TESTING = False
```
tzaffi, line 29: Currently not actually running the dry-runs on github. Worth discussing how to move forward o this.

tests/semantic_test.py
```diff
@@ -0,0 +1,561 @@
...
+    csv_from_dryruns,
+    dryrun_assert,
+    lightly_encode_args,
+    lightly_encode_output,
+    scratch_encode,
+)
+
+from pyteal import *
+
+# TODO: get tests working on Circle and set this to True
```
tzaffi, line 28:
```suggestion
# TODO: get tests working on github and set this to True
```

Makefile
```diff
@@ -0,0 +1,28 @@
...
+	pytest
+
+
+# Extras:
+
+coverage:
+	pytest --cov-report html --cov=pyteal
+
+blackbox:
+	HAS_ALGOD=TRUE pytest tests/semantic_test.py
```
tzaffi, line 28:
```suggestion
	HAS_ALGOD=TRUE pytest tests/semantic_test.py

```

requirements.txt
```diff
@@ -1,5 +1,7 @@
 black==21.7b0
 mypy==0.931
+git+https://github.com/algorand/py-algorand-sdk@teal-blackbox
```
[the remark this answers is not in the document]
tzaffi: 👍

tests/semantic_asserts.py
```diff
@@ -0,0 +1,299 @@
...
+            @Subroutine(TealType.uint64, input_types=[TealType.uint64, TealType.uint64])
+            def euclid(x, y):
+                a = ScratchVar(TealType.uint64)
+                b = ScratchVar(TealType.uint64)
+                tmp = ScratchVar(TealType.uint64)
+                start = If(x < y, Seq(a.store(y), b.store(x)), Seq(a.store(x), b.store(y)))
+                cond = b.load() > Int(0)
+                step = Seq(
+                    tmp.store(b.load()), b.store(Mod(a.load(), b.load())), a.store(tmp.load())
+                )
```
tzaffi, line 160: @ahangsu @jasonpaulos - is there a better way to write this for loop? I had to provide `Do()` with a dummy instruction `Pop(Int(1))` but I'd rather avoid this if possible.

ahangsu: Probably an empty `Seq()`? Since `Do` wants an expression that has none type, then empty seq of instructions might be an option?

tzaffi: 👍

requirements.txt
```diff
@@ -1,5 +1,7 @@
...
 pytest
+pytest-cov
 pytest-timeout
-py-algorand-sdk
+tabulate==0.8.9
```
michaeldiamant, line 7: @tzaffi For my understanding:  It seems pyteal does _not_ use _tablulate_.  Though I know py-algorand-sdk uses the dependency.  I'm wondering if the dependency is declared to explicitly state the transitive dependency or if there's another reason?

I'm asking the question as a way to learn a bit more about pip conventions/norms.

pyteal/ast/subroutine.py
michaeldiamant: @tzaffi Optional - I think the comment you've left is a worthwhile inline addition within `SubroutineDefinition`.  I can see it being helpful for future maintainers to understand _why_ `input_types` was added here.

I don't feel _too_ strongly, so feel welcomed to disregard.

requirements.txt
tzaffi: I believe this is a temporary artifact of me `pip install`ing our **PySDK** from github instead of from PyPI, so `pip` isn't smart enough to compute the transitive dependency and I had to add it here. But once my PR gets merged in the PySDK repo, we should double check it.

Thinking more along these lines, I realized that there is a more general problem? How do you tell that all of our requirements are actually necessary?

Here's a possible way:

1. We should add autoflake (as you suggested) to our CI and use it to enforce having no unused imports
2. We should then be able to use `pipreqs` to [validate that all the items](https://python-bloggers.com/2021/10/how-to-generate-the-requirements-of-your-python-project-based-on-your-imports/) in `requirements.txt` are actually needed

Ironically, it's exactly this situation that would be a false positive...

pyteal/ast/subroutine.py
tzaffi: 👍

tests/semantic_asserts.py
```diff
@@ -0,0 +1,299 @@
...
+def _indexer_client(
+    indexer_address="http://localhost:8980", indexer_token="a" * 64
+) -> indexer.IndexerClient:
+    """Instantiate and return Indexer client object."""
+    return indexer.IndexerClient(indexer_token, indexer_address)
+
+
+def algod_with_assertion():
+    algod = _algod_client()
+    assert algod.status(), "alod.status() did not produce any results"
```
michaeldiamant, line 36: @tzaffi Friendly pointer to minor typo.

```suggestion
    assert algod.status(), "algod.status() did not produce any results"
```

Makefile
```diff
@@ -0,0 +1,29 @@
...
+test:
+	pytest
+
+
+# Extras:
+
+coverage:
+	pytest --cov-report html --cov=pyteal
+
+blackbox:
```
michaeldiamant, line 27: @tzaffi `make blackbox` is failing due to a change in py-algorand-sdk.  I opened https://github.com/algorand/pyteal/pull/258 to fix the failures.  Can you review?

Let's resolve the thread once the child PR is either merged or closed.

requirements.txt
michaeldiamant: @tzaffi Got it - thanks for the response.

* Considering the situation more broadly, I've also wondered how we'd separate src vs test dependencies.  I imagine that requires crossing into a more sophisticated build tool.  
* _pipreqs_ looks interesting.  Thanks for raising.  I added a note into https://github.com/algorand/pyteal/issues/218.
* If it's OK with you, let's leave the thread open until py-algorand-sdk is merged?  It can serve as a visual reminder to double-check the dependencies.

Makefile
michaeldiamant: Resolving - #258 is merged.

Makefile
```diff
@@ -0,0 +1,29 @@
...
+	HAS_ALGOD=TRUE pytest tests/semantic_test.py
```
michaeldiamant, line 28: @tzaffi I feel like there's a chance to unify naming between this PR and https://github.com/algorand/py-algorand-sdk/pull/298.  To illustrate what I'm seeing, I opened an optional PR for your review: https://github.com/algorand/pyteal/pull/259.

Let's resolve the thread once the child PR is either closed or merged.

requirements.txt
tzaffi: 👍

Makefile
michaeldiamant: @tzaffi Thanks for being willing to take #259 in - Resolving.  ☕

tzaffi: 🙏

tests/semantic_asserts.py
tzaffi: It worked! Thank you @ahangsu

tests/blackbox_test.py
```diff
@@ -0,0 +1,629 @@
...
+    DryRunEncoder as Encoder,
+    DryRunExecutor,
+    DryRunProperty as DRProp,
+    DryRunTransactionResult,
+    SequenceAssertion,
+)
+
+from pyteal import *
+
+# TODO: get tests working on github and set this to True
```
michaeldiamant, line 26: @tzaffi FYI - I'm _trying_ to work on the TODO for integration tests on CI.

I'm working out of https://github.com/algorand/pyteal/pull/260.  If it proves too challenging, I may suggest opening a separate issue.

tzaffi: Sync'ed up offline but let's summarize my wish list regarding testing (with the caveat that clearly "testing" is better than "no testing").

From most desired to least desired (bug again, even least desired is really great):
1. have a dry run test that doesn't require running a node (a la `goal clerk dryrun` without the `remote` option)
2. have the dry run test that uses a quickly downloadable pre-built docker image (a la `docker pull algorand/stable`)
3. have a dry run that stands up the minimum necessary sandbox-like environment
4. have a dry run that stands up a "heavier" sandbox environment similar to the one that the SDK's do

tests/blackbox_asserts.py
```diff
@@ -0,0 +1,298 @@
...
+        def approval():
+            if subr.subroutine.returnType == TealType.none:
+                result = ScratchVar(TealType.uint64)
+                part1 = [subr_caller(), result.store(Int(1337))]
+            else:
+                result = ScratchVar(subr.subroutine.returnType)
+                part1 = [result.store(subr_caller())]
+
+            part2 = [make_log(result.load()), make_return(result.load())]
+            return Seq(*(part1 + part2))
```
jasonpaulos, line 295: Why log or return something on the top of the stack if the return value is already stored in `result`? It seems like you could directly check the contents of that scratch slot without the caveats of the other two methods

tests/blackbox_asserts.py
```diff
@@ -0,0 +1,298 @@
...
+        if e.type_of() == TealType.uint64:
+            return Log(Itob(e))
+        if e.type_of() == TealType.bytes:
+            return Log(e)
+        return Log(Bytes("nada"))
+
+    if mode == Mode.Signature:
+
+        def approval():
+            return make_return(subr_caller())
```
jasonpaulos, line 282: Shouldn't this also have a case for the return type being `TealType.none`?

tests/blackbox_asserts.py
```diff
@@ -0,0 +1,298 @@
...
+    def make_return(e):
+        if e.type_of() == TealType.uint64:
+            return e
+        if e.type_of() == TealType.bytes:
+            return Len(e)
+        if e.type_of() == TealType.anytype:
+            x = ScratchVar(TealType.anytype)
+            return Seq(x.store(e), Int(1337))
+        # TealType.anytype:
+        return Seq(e, Int(1337))
```
jasonpaulos, line 270: dead code

tests/blackbox_asserts.py
```diff
@@ -0,0 +1,298 @@
...
+        * arg of TealType.uint64:
+            convert arg to int using Btoi() when received
+    * output conversion:
+        * TealType.uint64:
+            provide subroutine's result to the top of the stack when exiting program
+        * TealType.bytes:
+            convert subroutine's result to the top of the stack to its length and then exit
+        * TealType.none or TealType.anytype:
+            push Int(1337) to the stack as it is either impossible (TealType.none),
+            or unknown at compile time (TealType.any) to convert to an Int
```
jasonpaulos, line 88: What is the point of doing this?

pyteal/ast/subroutine.py
```diff
@@ -15,18 +15,37 @@
...
     ) -> None:
+        """
+        Args:
+            implementation: The python function that was used to define the subroutine
+            returnType: the TealType the the subroutine should return
+            nameStr (optional): the name that is used to identify the subroutine.
+                If omitted, the name defaults to the implementation's __name__ attribute
+            input_types (optional): list of TealTypes of the arguments of the subroutine.
```
jasonpaulos, line 35: My initial impression is that, while this addition makes sense in the context of this PR, it will quickly become outdated when ABI types are introduced, which don't map have an explicit `TealType`.

A longer term solution would be to look into a generic `Expr` class like this issue describes: https://github.com/algorand/pyteal/issues/178

.github/workflows/build.yml
```diff
@@ -19,16 +19,15 @@ jobs:
...
+      - name: Install required os level applications
         run: |
-          pip install -r requirements.txt
-          pip install -e .
+          apt update -y
+          apt install -y git make
+          git --version
```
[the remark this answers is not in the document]
tzaffi: trying and will see what happens

tests/blackbox_asserts.py
tzaffi: I'm not sure I'm following your suggestion. Are you proposing not to log `result` or not to save at the top of the stack, or did you find a more efficient way to go about this?

I want to log in addition to leaving at the top of the stack to provide maximum flexibility to the tester and for compatibility with ABI, in case the subroutine encodes as ABI and the reader uses the ABI decoding functionality of an SDK.

.github/workflows/build.yml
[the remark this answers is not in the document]
tzaffi: Your idea works! Here's the [proof](https://github.com/algorand/pyteal/runs/5668121732?check_suite_focus=true#step:5:8).

tests/blackbox_asserts.py
tzaffi: This case - `returnType == TealType.none` for `mode.Signature` is covered by [this unit test](https://github.com/algorand/pyteal/blob/005f72e671eecd82cc151179b051d2a370704d53/tests/blackbox_test.py#L137)

Also the semantic tests for [@Subroutine square_byref()](https://github.com/algorand/pyteal/blob/005f72e671eecd82cc151179b051d2a370704d53/tests/blackbox_test.py#L91) which are running in PySDK prove that it works as expected.

tests/blackbox_asserts.py
tzaffi: Just a bad comment (thanks for catching!) but not dead code. I'm changing the comment to:
```python
# TealType.none:
```

tests/blackbox_asserts.py
tzaffi: I want to have a way for the user to check that TEAL that was created for a subroutine is functioning as expected, even if the subroutine doesn't return anything (`TealType.none`), or the its type is something I can't actually return automatically (`TealType.anytype`). So I'm providing the sentinel 1337 as a signal that the PyTEAL generator works fine, and we -in fact- got no output. It's almost impossible to prove that something _doesn't exist_ but this tries as hard to do that as it can.

pyteal/ast/subroutine.py
tzaffi: This is an excellent point and a much better user experience/API then the quick and dirty way I'm proposing. Given today's PySDK developments, I think I will have plenty of other prep-work to pave the way for this PR so I propose tabling this issue until next week. Then we can have a powwow and hammer out a unified approach forward.

pyteal/compiler/subroutines.py
```diff
@@ -11,7 +11,7 @@
 Node = TypeVar("Node")
 
 
-def depthFirstSearch(graph: Dict[Node, Set[Node]], start: Node, end: Node) -> bool:
+def graph_search(graph: Dict[Node, Set[Node]], start: Node, end: Node) -> bool:
```
tzaffi, line 14: seems to be a combo of dfs + bfs,

tests/blackbox_test.py
tzaffi: resolving as this is a separate issue discussed on #260

tests/user_guide_test.py
```diff
@@ -22,6 +45,130 @@ def user_guide_snippet_dynamic_scratch_var() -> Expr:
     )
 
 
+# RECURSIVE VERSION HAS AN INFINITE LOOP!!!!
+@Subroutine(TealType.uint64, input_types=[TealType.uint64, TealType.uint64])
```
tzaffi, line 49: Is this a real bug?

algoidurovic: Shouldn't `buggy_euclid(x, Mod(y, x))` be `buggy_euclid(y, Mod(x, y))`?

tzaffi: No it's not a real bug. Upshot, the python equivlent had the same bug. The fix is the following
(basically swap x and y in the `Else`):


```python
@Subroutine(TealType.uint64, input_types=[TealType.uint64, TealType.uint64])
def buggy_euclid(x, y):
    return (
        If(x < y)
        .Then(buggy_euclid(y, x))
        .Else(If(y == Int(0)).Then(x).Else(buggy_euclid(y, Mod(x, y))))
    )
```

# TODO: run the blackbox agsint this new "correction"

## Python model to assert as well:
```python
def be(x, y):
    return be(y, x) if x < y else (x if y == 0 else be(y, x % y))
```

requirements.txt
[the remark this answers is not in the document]
tzaffi: I removed `requirements.txt` altogether (from the top level but `docs/requirements.txt` is still present

tzaffi:
Defining Issue: #214

pyteal/ast/gitxn.py
```diff
@@ -1,11 +1,10 @@
 from typing import TYPE_CHECKING, cast, Union
 
+from pyteal.ast.expr import Expr
```
tzaffi, line 3: sorting imports

pyteal/ast/if_.py
```diff
@@ -1,9 +1,9 @@
 from typing import TYPE_CHECKING
 
+from pyteal.ast.expr import Expr
```
tzaffi, line 3: sorting imports

pyteal/ast/maybe.py
```diff
@@ -1,11 +1,10 @@
 from typing import List, Union
```
tzaffi, line 1: sorting imports

pyteal/ast/methodsig.py
```diff
@@ -1,10 +1,9 @@
 from typing import TYPE_CHECKING
```
tzaffi, line 2: sorting imports

Makefile
```diff
@@ -40,10 +40,28 @@ mypy:
...
+NUM_PROCS = 1 
+integration-run:
+	pytest -n $(NUM_PROCS) --durations=10 -sv tests/integration
+
+integration-test: integration-run # graviton-demo-run
+
 # Extras:
 coverage:
 	pytest --cov-report html --cov=pyteal
+
```
tzaffi, line 57: TODO: probably best to carve out the following for a follow-up PR against graviton@🐗

examples/signature/factorizer_game.py
```diff
@@ -0,0 +1,86 @@
+# WARNING: this logic sig is for demo purposes only
```
tzaffi, line 1: This example features in [my fork of the pyteal course](https://github.com/tzaffi/pyteal-course/blob/main/CONTRACT_PAYOUT.md) and will be useful to demonstrate graviton@🐗

tests/user_guide_test.py
```diff
@@ -1,84 +0,0 @@
-import pytest
```
tzaffi, line 1: expanded and moved into `tests/integration`

tests/unit/pass_by_ref_test.py
```diff
@@ -2,7 +2,7 @@
```
tzaffi, line 2: TODO: major cleanup

tests/unit/compile_test.py
```diff
@@ -1,4 +1,4 @@
-import os
+from pathlib import Path
```
tzaffi, line 1: Using `pathlib` instead of `os.path` in this test

tests/integration/user_guide_test.py
```diff
@@ -0,0 +1,294 @@
...
+
+import pyteal as pt
+from ..compile_asserts import assert_new_v_old
+
+
+def user_guide_snippet_dynamic_scratch_var() -> pt.Expr:
+    """
+    The user guide docs use the test to illustrate `DynamicScratchVar` usage.  If the test breaks, then the user guide docs must be updated along with the test.
+    """
+    from pyteal import Assert, Int, DynamicScratchVar, ScratchVar, Seq, TealType
```
tzaffi, line 11: TODO: update the corresponding docs with the new non wildcard imports

tests/integration/blackbox_test.py
```diff
@@ -0,0 +1,656 @@
...
+)
+
+from tests.compile_asserts import assert_teal_as_expected
+from tests.blackbox_asserts import (
+    algod_with_assertion,
+    blackbox_pyteal,
+    mode_to_execution_mode,
+)
+
+# TODO: remove this comment before merging
```
tzaffi, line 31:
```suggestion
# TODO: remove this comment before merging
```

TODO: heads up!

setup.py
```diff
@@ -22,9 +22,14 @@
             "flake8==4.0.1",
             "flake8-tidy-imports==4.6.0",
             "mypy==0.942",
+            "py-algorand-sdk",
             "pytest==7.1.1",
             "pytest-cov==3.0.0",
             "pytest-timeout==2.1.0",
+            "graviton@git+https://github.com/algorand/graviton@🦙",
+            # "graviton@git+https://github.com/algorand/graviton@fuzz",
```
tzaffi, line 30:
```suggestion
```

delete before merging

pyteal/compiler/scratchslots.py
```diff
@@ -1,9 +1,9 @@
 from typing import Tuple, Set, Dict, Optional, cast
```
tzaffi, line 1: sorting imports

pyteal/compiler/optimizer/optimizer.py
```diff
@@ -1,7 +1,8 @@
 from typing import Set
+
 from pyteal.ast import ScratchSlot
```
tzaffi, line 3: sorting imports

pyteal/ast/widemath.py
```diff
@@ -1,9 +1,9 @@
 from typing import List, Tuple, TYPE_CHECKING
 
-from pyteal.types import TealType
```
tzaffi, line 3: sorting imports

pyteal/ast/while_.py
```diff
@@ -1,9 +1,9 @@
 from typing import TYPE_CHECKING, Optional
 
+from pyteal.ast.expr import Expr
```
tzaffi, line 3: sorting

pyteal/ast/tmpl.py
```diff
@@ -1,8 +1,8 @@
 from typing import TYPE_CHECKING
 
-from pyteal.types import TealType, valid_tmpl
```
tzaffi, line 3: sorting

pyteal/ast/ternaryexpr.py
```diff
@@ -1,9 +1,9 @@
 from typing import Tuple, TYPE_CHECKING
 
-from pyteal.types import TealType, require_type
 from pyteal.errors import verifyTealVersion
 from pyteal.ir import TealOp, Op, TealBlock
 from pyteal.ast.expr import Expr
```
tzaffi, line 5:
```suggestion
from pyteal.ast.expr import Expr
from pyteal.errors import verifyTealVersion
from pyteal.ir import TealOp, Op, TealBlock
```

sort

pyteal/ast/subroutine.py
tzaffi: I see that in the ABI work proposed version of [ast/subroutine.py](https://github.com/algorand/pyteal/blob/3fb73746bf6d0a25aad63ddb162bf8c62c279e07/pyteal/ast/subroutine.py#L383) is creating a new class `ABIReturnSubroutineFnWrapper` and hence the changes proposed here do not strictly conflict with our ABI work. So I think the way I proposed could technically work. However, I can see that the it is not the most elegant design, or that we might not want to expose it to developers at large, except for testing purposes. So I will propose several ways of going forward and welcome others well:

1. keep as is (new optional parameter `input_types: Optional[List[Tealtype]]`)
2. allow generics to specify the input types (though I haven't tested to see if it works with `mypy`). I.e. by endowing `Expr` and `ScratchVar` as generics we could then define something like:
```python
@Subroutine(TealType.uint64)
def foo(x: ScratchVar[TealType.bytes], y: Expr[TealType.uint64]):
    ...
```
and then infer the generic type `T_x` of `x` for example via: `T_x = inspect.get_annotations(foo)['x'].__args__[0]`
3. Instead of making `Expr` and `ScratchVar` generic, we could create new types `ExprParam(Generic[T])` and `ScratchVarParam(Generic[T])` which are also allowed as parameter types of `@Subroutines` and use an approach similar to no. 2 
4. If we want to hide this for general purpose we could have a new decorator `@SubroutineTester` which inherits from `@Subroutine` and allows doing either no.'s 1 or 3 without touching `@Subroutine`'s functionality

tzaffi: I see that in the ABI work proposed version of [ast/subroutine.py](https://github.com/algorand/pyteal/blob/3fb73746bf6d0a25aad63ddb162bf8c62c279e07/pyteal/ast/subroutine.py#L383) is creating a new class `ABIReturnSubroutineFnWrapper` and hence the changes proposed here do not strictly conflict with our ABI work. So I think the way I proposed could technically work. However, I can see that the it is not the most elegant design, or that we might not want to expose it to developers at large, except for testing purposes. So I will propose several ways of going forward and welcome others well:

1. keep as is (new optional parameter `input_types: Optional[List[Tealtype]]`)
2. allow generics to specify the input types (though I haven't tested to see if it works with `mypy`). I.e. by endowing `Expr` and `ScratchVar` as generics we could then define something like:
```python
@Subroutine(TealType.uint64)
def foo(x: ScratchVar[TealType.bytes], y: Expr[TealType.uint64]):
    ...
```
and then infer the generic type `T_x` of `x` for example via: `T_x = inspect.get_annotations(foo)['x'].__args__[0]`
3. Instead of making `Expr` and `ScratchVar` generic, we could create a new types `SubrParam(Generic[
4. If we want to hide this for general purpose we could have a new decorator `@SubroutineTester` which inherits from `@Subroutine` and allows doing either 1 or 3 without touching `@Subroutine`'s functionality

_BTW, my preference is no. 1, because it's already done and the others seem very complicated_

Makefile
```diff
@@ -40,10 +40,30 @@ mypy:
...
+integration-test: integration-run # graviton-demo-run
+
+all-tests: build-and-test integration-test
+
 # Extras:
 coverage:
 	pytest --cov-report html --cov=pyteal
+
+# The following depends on [Graviton PR #9](https://github.com/algorand/graviton/pull/9):
```
tzaffi, line 60:
```suggestion

# The following depends on [Graviton PR #9](https://github.com/algorand/graviton/pull/9):
```

Makefile
```diff
@@ -40,10 +40,31 @@ mypy:
...
+
+all-tests: build-and-test integration-test
+
 # Extras:
 coverage:
 	pytest --cov-report html --cov=pyteal
+
+
+# The following depends on [Graviton PR #9](https://github.com/algorand/graviton/pull/9):
+#
```
tzaffi, line 62:
```suggestion
# TODO: Delete the below before merging:
```

.gitignore
```diff
@@ -133,5 +133,9 @@ dmypy.json
 .idea
 .vscode
 
+# Comma Separated Value report files
```
tzaffi, line 136:
```suggestion
# comma seperated vals report files
```

setup.py
```diff
@@ -22,9 +22,13 @@
             "flake8==4.0.1",
             "flake8-tidy-imports==4.6.0",
             "mypy==0.942",
+            "py-algorand-sdk",
```
tzaffi, line 25:
```suggestion
```

it's already available

Makefile
```diff
@@ -40,10 +40,40 @@ mypy:
...
+	docker-compose stop algod
 
 build-and-test: check-generate-init lint test-unit
 
+# set NUM_PROCS = auto when the following issue has been fixed https://github.com/algorand/pyteal/issues/199
+NUM_PROCS = 1 
+integration-run:
+	pytest -n $(NUM_PROCS) --durations=10 -sv tests/integration
+
+integration-test: integration-run # graviton-demo-run
```
tzaffi, line 58: TODO: clean up (save in followup) before committing

Makefile
```diff
@@ -40,10 +40,29 @@ mypy:
...
+	docker-compose stop algod
 
 build-and-test: check-generate-init lint test-unit
 
+# TODO: set NUM_PROCS = auto when the following issue has been fixed https://github.com/algorand/pyteal/issues/199
+NUM_PROCS = 1 
+integration-run:
+	pytest -n $(NUM_PROCS) --durations=10 -sv tests/integration
+
+integration-test: integration-run
```
tzaffi, line 58: Followup #294 also runs the demo here, hence the seemingly redundant build target

tests/blackbox_asserts.py
tzaffi: FYI - this is now in `utils/blackbox.py L345`

utils/blackbox.py
```diff
@@ -0,0 +1,348 @@
...
+        * TealType.uint64:
+            convert subroutine's output using Itob() and log the result
+        * TealType.bytes:
+            log the subroutine's result
+        * TealType.none or TealType.anytype:
+            log Itob(Int(1337)) as it is either impossible (TealType.none),
+            or unknown at compile time (TealType.anytype) how to convert to Bytes
+
+
+    Example 1: Using blackbox_pyteal for a simple test of both an app and logic sig:
```
tzaffi, line 106: @jasonpaulos @michaeldiamant - where should this doc live? I could see a new section `PyTeal Dryrun Testing` coming after `Compiler Options`. Also, this doesn't seem to properly belong under `API` and I would argue that we should leave that section untouched.

requirements.txt
tzaffi: Now we need `tabulate` because of `graviton`

pyteal/ast/subroutine.py
```diff
@@ -15,37 +15,74 @@
...
         self,
         implementation: Callable[..., Expr],
         returnType: TealType,
         nameStr: str = None,
+        input_types: List[TealType] = None,
     ) -> None:
+        """
+        Args:
+            implementation: The python function defining the subroutine
+            returnType: the TealType to be returned by teh subroutine
```
michaeldiamant, line 32:
```suggestion
            returnType: the TealType to be returned by the subroutine
```

tzaffi: good catch 🙏

.github/workflows/build.yml
```diff
@@ -25,6 +29,57 @@ jobs:
...
+           sudo curl -L "https://github.com/docker/compose/releases/download/1.29.2/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
+           sudo chmod +x /usr/local/bin/docker-compose
+           sudo ln -s /usr/local/bin/docker-compose /usr/bin/docker-compose
+           docker-compose --version
+      - name: Check cache for Docker layers
+        uses: satackey/action-docker-layer-caching@v0.0.11
+        # Ignore the failure of a step and avoid terminating the job:
+        continue-on-error: true
+        with:
+          key: docker-layer-caching-${{ github.workflow }}-${{ hashFiles(env.SANDBOX_CONFIG, env.SANDBOX_GENESIS) }}-{hash}
```
michaeldiamant, line 72: @tzaffi A few questions here:

* Is the provided key still accurate?
  * https://github.com/algorand/pyteal/runs/6164428544?check_suite_focus=true suggests to me that _only_ `makerxau/algorand-sandbox-dev:latest` is cached.  So, it feels to me like the following changes can be made:
    * Remove `env.SANDBOX_CONFIG` and `env.SANDBOX_GENESIS`.
    * Remove `key` override to use the default name.
* Since the PR refers to a `latest` image, does github caching provide the expected behavior?  
  * For `makerxau/algorand-sandbox-dev`, when a new image version appears as `latest`, will CI pick up the change?
    * I don't know off-hand, but I'm guessing _no_. 
  * I see https://github.com/MakerXStudio/algorand-sandbox-dev does _not_ provide tags.  
  * If we expect new image versions to be picked up and provided the build time remains tolerable, then I propose removing the caching.

.github/workflows/build.yml
```diff
@@ -25,6 +29,57 @@ jobs:
...
+      - uses: actions/setup-python@v3
+        with:
+          python-version: "${{ matrix.python }}"
+      - name: Test Python version
+        run: |
+          installed="$(python --version)"
+          expected="${{ matrix.python }}"
+          echo $installed
+          [[ $installed =~ "Python ${expected}" ]] && echo "Configured Python" || (echo "Failed to configure Python" && exit 1)
+      - name: Install required os level applications
```
michaeldiamant, line 51: @tzaffi There's various setup/dependency downloading happening here + subsequent `run` stanzas.  Is it all still needed now that the PR uses a pre-built sandbox image?

tests/integration/user_guide_test.py
```diff
@@ -0,0 +1,293 @@
...
+    assert app_result.stack_top() == x**2, app_result.report(
+        args, "stack_top() gave unexpected results for app"
+    )
+    assert lsig_result.stack_top() == x**2, lsig_result.report(
+        args, "stack_top() gave unexpected results for lsig"
+    )
+
+    # check to see that itob of x^2 has been logged (only for the app case)
+    assert app_result.last_log() == DryRunEncoder.hex(x**2), app_result.report(
+        args, "last_log() gave unexprected results from app"
```
michaeldiamant, line 134:
```suggestion
        args, "last_log() gave unexpected results from app"
```

utils/blackbox.py
```diff
@@ -0,0 +1,363 @@
...
+            assert app_result.stack_top() == x**2, app_result.report(
+                args, "stack_top() gave unexpected results for app"
+            )
+            assert lsig_result.stack_top() == x**2, lsig_result.report(
+                args, "stack_top() gave unexpected results for lsig"
+            )
+
+            # check to see that itob of x^2 has been logged (only for the app case)
+            assert app_result.last_log() == DryRunEncoder.hex(x**2), app_result.report(
+                args, "last_log() gave unexprected results from app"
```
michaeldiamant, line 144:
```suggestion
                args, "last_log() gave unexpected results from app"
```

utils/blackbox.py
```diff
@@ -0,0 +1,363 @@
...
+    Args:
+        subr: annotated subroutine to wrap inside program.
+            Note: the `input_types` parameters should be supplied to @Subroutine() annotation
+        mode: type of program to produce: logic sig (Mode.Signature) or app (Mode.Application)
+
+    Returns:
+        a function that called with no parameters -e.g. result()-
+        returns a PyTeal expression compiling to a ready-to-test TEAL program.
+
+    The return type is callable in order to adhere to the API of end-to-end unit tests.
```
michaeldiamant, line 64: @tzaffi Optional - If possible, I prefer to avoid the _end-to-end_ terminology because we've otherwise adopted _blackbox_.  Feel welcomed to resolve if you prefer as is.

```suggestion
    The return type is callable in order to adhere to the API of blackbox tests.
```

utils/blackbox.py
```diff
@@ -0,0 +1,363 @@
...
+            # GCD via the Euclidean Algorithm (recursive version):
+            @Subroutine(TealType.uint64, input_types=[TealType.uint64, TealType.uint64])
+            def euclid(x, y):
+                return (
+                    If(x < y)
+                    .Then(euclid(y, x))
+                    .Else(If(y == Int(0)).Then(x).Else(euclid(y, Mod(x, y))))
+                )
+
+            # Generate PyTeal and TEAL for the recurive Euclidean algorithm:
```
michaeldiamant, line 285:
```suggestion
            # Generate PyTeal and TEAL for the recursive Euclidean algorithm:
```

tests/integration/user_guide_test.py
```diff
@@ -0,0 +1,293 @@
...
+    # GCD via the Euclidean Algorithm (recursive version):
+    @Subroutine(TealType.uint64, input_types=[TealType.uint64, TealType.uint64])
+    def euclid(x, y):
+        return (
+            If(x < y)
+            .Then(euclid(y, x))
+            .Else(If(y == Int(0)).Then(x).Else(euclid(y, Mod(x, y))))
+        )
+
+    # Generate PyTeal and TEAL for the recurive Euclidean algorithm:
```
michaeldiamant, line 275:
```suggestion
    # Generate PyTeal and TEAL for the recursive Euclidean algorithm:
```

.github/workflows/build.yml
tzaffi: I'll investigate and see which are actually still necessary.

utils/blackbox.py
michaeldiamant: @tzaffi A few thoughts:
* I agree with a new section under _Compiler Options_.  I'd name it _Blackbox testing_ or similar with _blackbox_ to stick with the naming applied throughout the repo.
* Additional thoughts:
  * Are we also saying the inline doc examples highlighted by the thread will be removed?
    * Open to discussion:  My preference is _yes_ because the examples aren't compiled and will go stale.  As an example, I've already opened several, minor spelling fix _suggestions_ in 2 places.
   * Open to discussion:  I'll follow the group's preference, but want to note that I'd be on-board to agree on _where_ user guides will be documented here.  And then supply the docs in a new PR because there's already a lot happening here. 
* Aside:  The question makes me wonder if a broader user guide re-organization makes sense.  For the moment, I'm thinking we opt for a change specific to the problem at-hand.

tests/integration/user_guide_test.py
```diff
@@ -0,0 +1,293 @@
...
+        DRProp.maxStackHeight: (
+            lambda args, actual: 2
+            <= actual
+            <= 3 * math.ceil(math.log2(max(args + (1,))))
+        ),
+        # the program PASS'es exactly for non-0 math.gcd (3 variants):
+        DRProp.status: lambda args: "PASS" if math.gcd(*args) else "REJECT",
+        DRProp.passed: lambda args: bool(math.gcd(*args)),
+        DRProp.rejected: lambda args: not bool(math.gcd(*args)),
+        # the program never erors:
```
michaeldiamant, line 253:
```suggestion
        # the program never errors:
```

tests/integration/user_guide_test.py
```diff
@@ -0,0 +1,293 @@
...
+
+    # Generate PyTeal and TEAL for the recurive Euclidean algorithm:
+    euclid_app = blackbox_pyteal(euclid, Mode.Application)
+    euclid_app_teal = compileTeal(euclid_app(), Mode.Application, version=6)
+
+    # Execute on the input sequence to get a dry-run inspectors:
+    algod = algod_with_assertion()
+    inspectors = DryRunExecutor.dryrun_app_on_sequence(algod, euclid_app_teal, inputs)
+
+    # Assert that each invarient holds on the sequences of inputs and dry-runs:
```
michaeldiamant, line 283:
```suggestion
    # Assert that each invariant holds on the sequences of inputs and dry-runs:
```

utils/blackbox.py
```diff
@@ -0,0 +1,363 @@
...
+    Seq,
+    SubroutineFnWrapper,
+    TealType,
+    Txn,
+)
+
+# ---- Clients ---- #
+
+
+def algod_client(
```
michaeldiamant, line 27: @tzaffi Minor:  What use cases do you have in mind where it'd be preferable to create a client _without_ asserting it's connected?

* I'm trying to gauge if there's a need to have users disambiguate `algod_client` from `algod_client_with_assertion`?
  * On the surface, I'd imagining using `with_assertion` all the time.
* _If_ we retain both methods, then I think `algod_with_assertion` ought to take the same parameters + defaults as `algod_client`.

Makefile
```diff
@@ -20,30 +24,58 @@ bundle-docs: bundle-docs-clean
 	doc2dash --name pyteal --index-page index.html --online-redirect-url https://pyteal.readthedocs.io/en/ _build/html && \
 	tar -czvf pyteal.docset.tar.gz pyteal.docset
 
+# ---- Code Quality ---- #
+
 generate-init:
```
michaeldiamant, line 29: @tzaffi Minor and optional:  Does `generate-init` fall under the _code quality_ or _setup_ umbrella?  I thought it's the latter because `build-and-test` fails otherwise.

I don't feel strongly.  If you prefer as is, feel welcomed to resolve.

utils/blackbox_test.py
```diff
@@ -0,0 +1,96 @@
...
+@pytest.mark.parametrize("subr, mode", product(UNITS, pt.Mode))
+def test_blackbox_pyteal(subr, mode):
+    """
+    TODO: here's an example of issue #199 at play - (the thread-safety aspect):
+    compare the following!
+    % pytest -n 2 tests/integration/graviton_test.py::test_blackbox_pyteal
+    vs
+    % pytest -n 1 tests/integration/graviton_test.py::test_blackbox_pyteal
+    """
+    path = Path.cwd() / "tests" / "teal" / "blackbox" / "unit"
```
michaeldiamant, line 87: @tzaffi What do you think of the following thought?

* Looking at the dir structure `tests/teal`, I feel it's difficult to discern where these files are used.
* Proposal:  Move `tests/teal/blackbox/unit` to `utils/resources` or a similarly named dir under `utils`.

.github/workflows/build.yml
tzaffi: Last I checked, they only put one docker image up. One hesitation to this approach is the possibility of support being dropped and the docker image going stale. I have [forked the repo](https://github.com/tzaffi/algorand-sandbox-dev) and it has an MIT license which means we can build the image against a newer `go-algorand` and push to our own dockerhub. I've verified that rebuilding the image works as expected and I was able to pass PyTEAL integration test on it. I've even created and uploaded that image (`tzaffi/algod:24apr2022`)

Since caching isn't really any faster than not, I'll remove it altogether. If build times become less tolerable, we can re-introduce.

Action items:
- [ ]  remove `env.SANDBOX_*`
- [ ]  remove caching
- [ ]  test against `tzaffi/algod:24apr2022`

utils/blackbox.py
tzaffi: 👍 

`algod_client()` doesn't call out so when accessing an `algod` in a context that doesn't already have it you can save a round trip. At some point in the evolution of this PR, this was actually useful.  However, there isn't any such usage now so I'm going to renaming it to `_algod_client()` indicating that this is private and only used internally (in fact, only by `algod_with_assertion()`. We can always make `_algod_client()` public if this need arises again.

Makefile
tzaffi: 👍

utils/blackbox_test.py
tzaffi: 👍

.github/workflows/build.yml
michaeldiamant: @tzaffi Few thoughts:
* Thanks for removing caching via https://github.com/algorand/pyteal/pull/249/commits/d56772f3d330dcf6f7df9a276c5cd3ccb8b5242d.
* For `tzaffi/algod:24apr2022` (from https://github.com/algorand/pyteal/commit/50f59c93bc1b16b32f6cfa718e573dd0046e61e2), how will other maintainers trace the image's lineage back to https://github.com/tzaffi/algorand-sandbox-dev?
  * I agree forking the repo to maintain our own image is a good move.  Though I'm wondering if we can stay with the prior image and ticket the desire to move to a self-managed image.

utils/blackbox.py
michaeldiamant: @tzaffi Gotcha - Thanks for the change in https://github.com/algorand/pyteal/pull/249/commits/01f681b0ab23abbeb63ac0bb75fb2b87120236a6.   Resolving.

utils/blackbox_test.py
michaeldiamant: @tzaffi Thanks for the move in https://github.com/algorand/pyteal/pull/249/commits/5520d710950f5281e8ed2a9312c366c809001310.  Resolving.

.gitignore
```diff
@@ -133,5 +133,9 @@ dmypy.json
 .idea
 .vscode
 
+# comma seperated vals report files
+*.csv
+!tests/**/*_example.csv
+
```
michaeldiamant, line 139: @tzaffi Friendly pointer:  For tests moved to `utils/test/teal/app_utest_noop_args.teal`, I think there's a missing a `.gitignore` entry.  The previous entry (`tests/**/*.teal`) no longer covers all scenarios.

utils/blackbox_test.py
```diff
@@ -0,0 +1,97 @@
...
+    compare the following!
+    % pytest -n 2 tests/integration/graviton_test.py::test_blackbox_pyteal
+    vs
+    % pytest -n 1 tests/integration/graviton_test.py::test_blackbox_pyteal
+    """
+    is_app = mode == pt.Mode.Application
+    name = f"{'app' if is_app else 'lsig'}_{subr.name()}"
+
+    compiled = pt.compileTeal(blackbox_pyteal(subr, mode)(), mode, version=6)
+    save_to = TEAL_PATH / (name + ".teal")
```
michaeldiamant, line 93: @tzaffi Several thoughts here around _hopefully_ helping maintainers understand repo structure.

If agreed to, I am on-board to make the changes in a follow-on PR if preferred.

* Can `tests/teal` be moved to `tests/unit/teal` to help demarcate its usage?  On the surface, it's unclear what will generate output to the dir.
* So that we can treat `teal/` dirs as generated-only dirs, are you opposed to keeping the `_expected.teal` in a different dir than `tests/teal`?
  * Rationale: The dir is cluttered with generated files.  It's hard to discern which files _should_ be retained.  Currently, most `_expected.teal` files are no longer referenced and can be deleted.
  * For example:  To locate closer to usage, can be moved to `tests/integration/resources` with an analogous generated dir `tests/integration/teal`.

pyteal/ast/subroutine.py
jasonpaulos: Thanks for expanding on all the choices here. My current thoughts are that 2 is most preferred, but 3 and 4 would be good solutions as well.

Though _if_ we decide this interface should be internal-only, then I'd say 4 is the most compelling, since it requires the least additional work. We can then change our approach over time as we see fit without endangering backwards compatibility of user code.

(I'm still reviewing the rest of this PR, but I thought you'd appreciate this suggestion ASAP)

.github/workflows/build.yml
tzaffi: That approach works too. I'll create a ticket for that.

.github/workflows/build.yml
tzaffi: I whittled down the github container's bootstrapping as far as I could go. I was able to get rid of `git` and `nodejs` but needed everything else. The dependency is as follows:

`sandbox` &rarr; `docker-compose` &rarr; `docker-*` &rarr; `gpg` &rarr; `curl`

pyteal/ast/subroutine.py
tzaffi: I'm going to experiment with @michaeldiamant 's #5, but just create the `@Blackbox` decorator in `utils/` that way we don't even touch `pyteal/`. Will keep you posted!

utils/blackbox.py
michaeldiamant: Coming back to the thread with fresher eyes, I'm changing my position a bit.  Happy to discuss live if easier.

I think my rubric is to favor:
* Accelerate merging the PR.
* Retain control to make backwards _incompatible_ changes to blackbox infrastructure.  At least until we gain a bit more experience.

My view is that blackbox testing is _definitely_ useful for PyTeal developers.  A primary question in my mind is:  What level of blackbox testing exposure (if any) do we provide to developers?

Given that rubric, I'm open to a few options for the PR's purpose:
* Opt to make blackbox testing a tool internal to PyTeal maintainers - If so, I'd _probably_ lean towards _no_ user guide docs because the user guide doesn't cover maintainers today (possibly, a separate concern).  I think we can provide inline pointers to examples and/or other lightweight references.
* Opt to make blackbox testing available to PyTeal developers - If so:
  * Declare at least in comments that the tooling is liable to change.  
  * I'd be open to keeping the docs as discussed in the prior note.

.github/workflows/build.yml
michaeldiamant: @tzaffi Thanks for making a pass here.  I experimented further by partitioning act-only steps in https://github.com/algorand/pyteal/pull/299.

Can you have a look to see if you'd feel comfortable bringing in the change?  Otherwise, I can hold until we merge here.

pyteal/ast/subroutine.py
tzaffi: Please refer to #297 which implements this approach

.github/workflows/build.yml
tzaffi: Here's the ticket in `algorand-internal` https://github.com/algorand/go-algorand-internal/issues/2004

utils/blackbox.py
tzaffi: committed to link in comments to where tests actually run.

tzaffi: Here is the new version. I've refactored the `user_guide_test` and removed the long executable portion of the `blackbox_pyteal()`'s comment:
https://github.com/algorand/pyteal/blob/6afe1af7400cc12a7aeaaf70cc6faaa5009d8db9/utils/blackbox.py#L105

.github/workflows/build.yml
michaeldiamant: @tzaffi Thanks for taking the time to ticket https://github.com/algorand/go-algorand-internal/issues/2004.  Resolving.

.gitignore
michaeldiamant: @tzaffi Thanks - #297 addresses the concern.  I'm assuming we'll merge #297 in here.  Resolving.

.github/workflows/build.yml
tzaffi: blocked by #299 discussion

utils/blackbox_test.py
tzaffi: This is now implemented

.github/workflows/build.yml
tzaffi: I've merged #299 with one [outstanding issue](https://github.com/algorand/pyteal/pull/299#issuecomment-1110296347)

pyteal/ast/subroutine.py
```diff
@@ -23,29 +27,63 @@ def __init__(
...
+
+        impl_params, anns, arg_types, byrefs = self.validate()
+        self.implementationParams: MappingProxyType[str, Parameter] = impl_params
+        self.annotations: dict[str, Expr | ScratchVar] = anns
+        self.expected_arg_types: list[type[Expr | ScratchVar]] = arg_types
+        self.by_ref_args: set[str] = byrefs
+
+        self.__name: str = nameStr if nameStr else self.implementation.__name__
+
+    def validate(
```
tzaffi, line 54: This is used also to validate `BlacboxWrapper` (decorator related objects)

utils/blackbox_test.py
michaeldiamant: @tzaffi Thanks for the changes to bring in a `generated` dir and restructure `tests/`.  ☕ 

Looks good to me as of https://github.com/algorand/pyteal/pull/249/commits/8adec2624225788d63763a0259f812bedd6eb0f4.  Resolving.

.github/workflows/build.yml
michaeldiamant: @tzaffi Thanks for taking in #299.  Resolving.

utils/blackbox.py
michaeldiamant: @tzaffi Thanks for the updates - I consider the thread resolved.  Since I didn't open it, I leave it open for you to resolve in case you're holding for other feedback.

Summarizing changes for anyone else looking in:
* We opted for _no_ user guide docs in favor of an inline doc pointer from `blackbox_pyteal` to examples.
* We retained `user_guide_test.py` as a unit test.  It had temporarily been moved under _integration_ prior to last changeset.

pyteal/ast/subroutine.py
```diff
@@ -106,6 +144,19 @@ def _validate_parameter_type(
                     )
                 )
 
+            if (
```
tzaffi, line 147: This is just plain wrong. Gotta fix it.

michaeldiamant (approved):
@tzaffi Excited to bring these testing facilities into use - thanks for the effort here across multiple feedback iterations.  ☕

jasonpaulos:
This got lost from my review comment:

It seems like `utils` is only used for tests, so can it be moved to `test/utils`? As a general rule we should keep the top level directory as clutter free as possible

pyteal/ast/subroutine.py
```diff
@@ -15,6 +15,10 @@
 
 
 class SubroutineDefinition:
+    """Class that leverages TEAL's `callsub` and `retsub` opcode-pair
+    to allow writing functions and subroutines.
```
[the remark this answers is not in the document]
tzaffi: 👍

examples/signature/factorizer_game.py
```diff
@@ -0,0 +1,86 @@
...
+            Txn.close_remainder_to() == Global.zero_address(),
+            X1 != X2,
+            PRIZE.load(),
+            Txn.amount() == PRIZE.load(),
+        ),
+    )
+
+
+def create(a, b, c):
+    return logicsig(*map(lambda x: int(str(x)), (a, b, c)))
```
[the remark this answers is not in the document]
tzaffi: This is a relic of some coding iterations that got the lsig to work for [a tutorial I forked from the pyteal course](https://github.com/tzaffi/pyteal-course/blob/main/CONTRACT_PAYOUT.md). 

Taking `int(str(x))` was probably an attempt to deal with decimals at the same time as system arg's... but that makes no sense (e.g., you can't take `int("1.5")` but `int(1.5)`. I'll just make this `int(x)` for now. A future PR will make the necessary adjustments.

pyteal/ast/subroutine.py
tzaffi: 👍

pyteal/ast/subroutine.py
```diff
@@ -23,29 +27,63 @@ def __init__(
...
+    ]:
+        implementation = self.implementation
         if not callable(implementation):
             raise TealInputError("Input to SubroutineDefinition is not callable")
 
-        sig = signature(implementation)
+        impl_params = signature(implementation).parameters
+        anns = get_annotations(implementation)
+        arg_types = []
+        byrefs = set()
```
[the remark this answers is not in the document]
tzaffi: 👍

pyteal/ast/subroutine_test.py
```diff
@@ -1,8 +1,10 @@
-from typing import List
+from itertools import product
 import pytest
+from typing import List
 
 import pyteal as pt
-from pyteal.ast.subroutine import evaluateSubroutine
+from pyteal.ast.subroutine import SubroutineDefinition, evaluateSubroutine
+from pyteal.errors import TealInputError
```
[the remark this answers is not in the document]
tzaffi: 👍

setup.py
```diff
@@ -21,10 +21,13 @@
             "black==22.3.0",
             "flake8==4.0.1",
             "flake8-tidy-imports==4.6.0",
+            "graviton@git+https://github.com/algorand/graviton@🦙",
             "mypy==0.942",
             "pytest==7.1.1",
             "pytest-cov==3.0.0",
             "pytest-timeout==2.1.0",
+            "pytest-xdist==2.5.0",
+            "tabulate==0.8.9",
```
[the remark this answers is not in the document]
tzaffi: `pytest-xdist` is introduced here so we can [run integration tests in parallel](https://github.com/algorand/pyteal/blob/ae9067a176018964adb5b3c2293f0f27d00be1cb/Makefile#L66). (Notice on [github it's running now with 2 threads](https://github.com/algorand/pyteal/runs/6204514711?check_suite_focus=true#step:8:108) after I turned it on again)

`tabulate` is coming from `graviton`. I could use your advice how I should declare the dependency in `graviton`

setup.py
```diff
@@ -21,10 +21,13 @@
             "black==22.3.0",
             "flake8==4.0.1",
             "flake8-tidy-imports==4.6.0",
+            "graviton@git+https://github.com/algorand/graviton@🦙",
```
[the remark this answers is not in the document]
tzaffi: I don't have a roadmap for this. Probably worth a discussion. I'm not sure how to gauge the importance of this. This is the current popularity of the repo:
image

tests/unit/module_test.py
```diff
@@ -1,6 +1,3 @@
-from pyteal import *
```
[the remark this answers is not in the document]
tzaffi: apologies

tzaffi:
> This got lost from my review comment:
> 
> It seems like `utils` is only used for tests, so can it be moved to `test/utils`? As a general rule we should keep the top level directory as simple as possible

Ok. I've moved `utils` back under `tests`

Makefile
```diff
@@ -20,30 +27,62 @@ bundle-docs: bundle-docs-clean
...
+# TODO: add `tests` and `utils` to $MYPY when graviton respects mypy (version 🐗) 
 MYPY = pyteal scripts
 mypy:
 	mypy $(MYPY)
 
 lint: black flake8 mypy
 
+# ---- Unit Tests (no algod) ---- #
+
+# TODO: add utils to multithreaded tests when following issue has been fixed https://github.com/algorand/pyteal/issues/199
```
tzaffi, line 51:
```suggestion
# TODO: add blackbox_test.py to multithreaded tests when following issue has been fixed https://github.com/algorand/pyteal/issues/199
```

pyteal/ast/subroutine.py
```diff
@@ -260,7 +303,7 @@ def __init__(
             fnImplementation, returnType=returnType, nameStr=name
         )
 
-    def __call__(self, *args: Expr, **kwargs) -> Expr:
+    def __call__(self, *args: Union[Expr, ScratchVar], **kwargs) -> Expr:
```
tzaffi, line 306:
```suggestion
    def __call__(self, *args: Expr | ScratchVar, **kwargs) -> Expr:
```

tests/unit/blackbox_test.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+@pytest.mark.parametrize("subr, mode", product(UNITS, pt.Mode))
+def test_blackbox_pyteal(subr, mode):
+    """
+    TODO: here's an example of issue #199 at play - (the thread-safety aspect):
+    compare the following!
+    % pytest -n 2 tests/integration/graviton_test.py::test_blackbox_pyteal
+    vs
+    % pytest -n 1 tests/integration/graviton_test.py::test_blackbox_pyteal
```
tzaffi, line 85:
```suggestion
    % pytest -n 2 tests/unit/blackbox_test.py::test_blackbox_pyteal
    vs
    % pytest -n 1 tests/unit/blackbox_test.py::test_blackbox_pyteal
```

tzaffi:
@jasonpaulos - I've either implemented your suggestions or answered your questions. FYI, regarding the privatizing of `BlackboxWrapper` and `SubroutineDefinition.validate()`. I did add an underscore prefix, but I wonder if that really makes sense, given that that `tests/blackbox.py` is a consumer.

LMK

setup.py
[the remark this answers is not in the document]
tzaffi: I'll try that

tzaffi:
> @tzaffi thanks for the changes. Since `BlackboxWrapper` is not in user-consumable code, I see no reason for it to have an underscore prefix (not to mention I've never seen it added to a class before). `SubroutineDefinition.validate()` _is_ in user-consumable code, so the underscore communicates that it's not meant to be directly used by consumers and shouldn't be relied upon. However, this is purely informational, so it's perfectly fine if our code continues to use the `_validate()` method.
> 
> Once [#249 (comment)](https://github.com/algorand/pyteal/pull/249#discussion_r861196899) is resolved I'll approve.

Ok, I believe we should be good to go @jasonpaulos . LMK if you spot anything else.

jasonpaulos (approved):

PR #170, opened 2022-01-25
`ABI Router` implementation and test

ahangsu:
# Summary

ABI Router helps create contract approval/clear program much easier.

closes #167 and #239

## Changes

- [x] `onBareAppCall` to handle bare app call
- [x] `onMethodCall` to handle method call (app arg redirect, check method selector)
- [x] App-args de-tuple and (de)serialize for feeding to registered method, and logging return
- [x] creation argument specification (check app id == 0 condition)
- [x] ABI Json generate from router
- [x] test cases

PR #256, opened 2022-03-22
ABI Type Annotation Support for Subroutine

### LINKED ISSUE #179: Support passing ABI types to and from subroutines
## Summary

#154 introduces ABI type support in PyTeal. Following that, it would be incredibly useful for subroutines to be able to take ABI types as arguments and to be able to return an ABI type.

The most obvious way to denote this would be to use the ABI type as an annotation to a subroutine's arguments or return value. 

## Scope

If a subroutine has ABI type arguments, then the compiler should pass a value representing the ABI type onto the stack (this is not necessarily the encoded value, since types like Uint8-64 are more naturally represented as TEAL uint64s)

Due to the fact that `abi.Type`s are not `Expr`s, I suspect that an `abi.ComputedType` will be needed to wrap the return value. For example:

```python
from pyteal import *

@Subroutine() # a return declaration here is not necessary anymore?
def mySubroutine(a: abi.Uint64, b: abi.Uint64) -> abi.ComputedType[abi.Uint64]:
    # TODO: somehow return a + b
```

Alternatively, we could provide a way to store the return value into an argument representing the output. Something like:

```python
@Subroutine() # a return declaration here is not necessary anymore?
def mySubroutine(a: abi.Uint64, b: abi.Uint64, *, output: abi.Uint64) -> Expr:
    return output.set(a.get() + b.get())
```

ahangsu:
## Description

To handle passing (and receiving) ABI typed argument in subroutines, we need to additionally allow following things to be available in subroutine (and related class instances):
- ABI type annotation for arguments in subroutine implementation, which will be passed to `Subroutine` decorator.
- `output` keyword argument, which contains output ABI value.
- Extended type checking for ABI arguments, the `type_spec` should match one specified in type annotations.

## Code Changes

Mostly on `subroutine(_test).py`.

Closes #179

### DIFF
--- a/pyteal/ast/subroutine_test.py
+++ b/pyteal/ast/subroutine_test.py
@@ -9,6 +9,7 @@ from .. import CompileOptions, Return
 
 options = CompileOptions(version=6)
 
+# something here
 
 def test_subroutine_definition():
     def fn0Args():


pyteal/ast/subroutine.py
```diff
@@ -76,37 +76,176 @@ def __init__(
...
                 raise TealInputError(
                     "Function has parameter {} of disallowed type {}. Only the types {} are allowed".format(
-                        parameter_name, ptype, (Expr, ScratchVar)
+                        parameter_name, ptype, (Expr, ScratchVar, "ABI")
                     )
                 )
 
-            return ptype
+    @staticmethod
+    def _validate_full_signature(
```
tzaffi, line 174: I recommend renaming to something that pays homage to the outputs. For example:

```python
def _arg_types_and_by_refs(...)
```
or if you think it's important to emphasize validation
```python
def _arg_types_and_by_refs_with_validation(...)
```

ahangsu: I think the former suggestion makes sense.

Essentially the reason I move the chunk of code outside `__init__` to a separate function is that: the complexity proliferate when we introduce `ABI` types in subroutine.

Then out of the purpose of retrieving all `arg_types` and `args_by_ref`, it fine to rename to the one you suggested.

pyteal/ast/subroutine_test.py
```diff
@@ -88,6 +88,13 @@ def fnWithExprAnnotations(a: Expr, b: Expr) -> Expr:
     def fnWithSVAnnotations(a: ScratchVar, b: ScratchVar):
         return Return()
 
+    def fnWithABIAnnotations(
+        a: abi.Byte,
+        b: abi.StaticArray[abi.Uint32, Literal[10]],
+        c: abi.DynamicArray[abi.Bool],
+    ):
```
tzaffi, line 95:
```python
    def fnWithABIAnnotations(
        a: abi.Byte,
        b: abi.StaticArray[abi.Uint32, Literal[10]],
        c: abi.DynamicArray[abi.Bool],
    ):
```

We have a bit of a race condition between this PR and #260. Here are the issues:
* It seems like Blackbox testing is important for vetting the semantic effects of this work
* But PR #260 supports a different format for specifying the input type, rather than using generics/templats as is done here
* It seems to me that this approach is better than that of #260

I propose that this PR go in first, and that #260 adopt the approach taken here. But that in short order, after #260 is merged a PR whose main goal is to semantically test the ABI router along with other types of subroutines (to ensure nothing broke) be created in short order.

pyteal/ast/subroutine.py
```diff
@@ -320,6 +490,12 @@ def __call__(self, fnImplementation: Callable[..., Expr]) -> SubroutineFnWrapper
 Subroutine.__module__ = "pyteal"
 
 
+def routerSubroutine(fnImplementation: Callable[..., Expr]) -> SubroutineFnWrapper:
```
tzaffi, line 493:
```suggestion
def router_subroutine(fnImplementation: Callable[..., Expr]) -> SubroutineFnWrapper:
```

Making this suggestion as going forward we are trying to be more pythonic in our naming conventions. Of course, this begs the question - what do we do about the fact that we are using different snake conventions in the same module?

ahangsu: 👍

pyteal/ast/subroutine.py
```diff
@@ -21,15 +34,13 @@ def __init__(
...
         returnType: TealType,
-        nameStr: str = None,
+        nameStr: Optional[str] = None,
+        routerRegistrable: bool = False,
     ) -> None:
         super().__init__()
         self.id = SubroutineDefinition.nextSubroutineId
         SubroutineDefinition.nextSubroutineId += 1
-
-        self.by_ref_args: Set[str] = set()
```
tzaffi, line 30: I see this declaration is being removed. I think it's still safe -and preferable- to keep a version of it. In general it's nice to know what all the fields are in a class along with their types as early as possible in the init. If you don't want to initialize the member here you can still do the following:
```python
self.by_ref_args: Set[str]
```

pyteal/ast/subroutine.py
```diff
@@ -40,34 +51,23 @@ def __init__(
 
         if "return" in annotations and annotations["return"] is not Expr:
             raise TealInputError(
-                "Function has return of disallowed type {}. Only Expr is allowed".format(
-                    annotations["return"]
-                )
+                f"Function has return of disallowed type {annotations['return']}. Only Expr is allowed"
```
tzaffi, line 54: 👍 yay for f-strings

pyteal/ast/subroutine.py
ahangsu: yes, I was just having a havoc on the code and was too lazy to annotate them for additional knowledge of type 😅

I will fix that soon.

ahangsu: OK I have them initialized on line 69 and 72!

tzaffi: Similar comment regarding `expected_arg_types`

ahangsu: I guess it's also handled

pyteal/ast/subroutine.py
```diff
@@ -21,15 +34,13 @@ def __init__(
         self,
         implementation: Callable[..., Expr],
         returnType: TealType,
-        nameStr: str = None,
+        nameStr: Optional[str] = None,
+        routerRegistrable: bool = False,
```
tzaffi, line 38:
```suggestion
        router_registrable: bool = False,
```
I feel guilty for even suggesting this...

pyteal/ast/subroutine.py
```diff
@@ -21,15 +34,13 @@ def __init__(
...
     ) -> None:
         super().__init__()
         self.id = SubroutineDefinition.nextSubroutineId
         SubroutineDefinition.nextSubroutineId += 1
-
-        self.by_ref_args: Set[str] = set()
-
-        self.expected_arg_types: List[Type[Union[Expr, ScratchVar]]] = []
+        self.__routerRegistrable = routerRegistrable
```
tzaffi, line 43:
```suggestion
        self.__routerRegistrable = router_registrable
```

tzaffi (commented):
not finished reviewing, but batching two suggestions into one

pyteal/ast/subroutine.py
ahangsu: oh boy, I should go back to pep8 and check the formatting...

pyteal/ast/subroutine.py
```diff
@@ -76,37 +78,176 @@ def __init__(
         self.declaration: Optional["SubroutineDeclaration"] = None
         self.__name = self.implementation.__name__ if nameStr is None else nameStr
 
+    @staticmethod
+    def check_if_ABI_class(arg_type: Type) -> bool:
```
tzaffi, line 82:
```suggestion
    def is_ABI_class(arg_type: Type) -> bool:
```
recommend renaming because `check_` sounds like some assertion is going to be made

pyteal/ast/subroutine.py
```diff
@@ -76,37 +78,176 @@ def __init__(
...
+        return (
+            isclass(arg_type)
+            and arg_type is not abi.BaseType
+            and issubclass(arg_type, abi.BaseType)
+        )
+
+    @staticmethod
+    def check_if_ABI_annotation(abi_anno: Any) -> bool:
```
tzaffi, line 90:
```suggestion
    def is_ABI_annotation(abi_anno: Any) -> bool:
```
similar comment

pyteal/ast/subroutine.py
```diff
@@ -76,37 +78,176 @@ def __init__(
...
+        try:
+            abi.type_spec_from_annotation(abi_anno)
+            return True
```
tzaffi, line 93:
```suggestion
    def is_annotation(obj: Any) -> bool:
        try:
            abi.type_spec_from_annotation(obj)
            return True
```
similar to previous, but also recommend renaming param since we don't really no yet that it's an `abi_anno`

pyteal/ast/subroutine.py
```diff
@@ -76,37 +78,176 @@ def __init__(
...
+        except TypeError:
+            return False
+
+    @staticmethod
+    def _check_ABI_output(annotations: Dict[str, type], sig: Signature) -> None:
```
tzaffi, line 98: Why can't we use the same approach as before to look at the return type and see if it has an annotation present? If present, then use that rather than ask the user to use `output`  as a keyword. Seems like that is a more consistent approach.

For a similar reason, the approach taken in my PR #249 (using `input_types`) is less preferable than the approach taken here.

pyteal/ast/subroutine.py
```diff
@@ -76,37 +78,176 @@ def __init__(
...
+                    outputPType
+                )
+            )
+
     @staticmethod
     def _validate_parameter_type(
-        user_defined_annotations: dict, parameter_name: str
-    ) -> Type[Union[Expr, ScratchVar]]:
+        user_defined_annotations: dict, parameter_name: str, enforceABI: bool
+    ) -> Union[Type[Union[Expr, ScratchVar]], abi.TypeSpec]:
```
tzaffi, line 134:
```suggestion
        user_defined_annotations: dict, parameter_name: str, enforce_ABI: bool
    ) -> Union[Type[Union[Expr, ScratchVar]], abi.TypeSpec]:
```
snake case...

pyteal/ast/subroutine.py
```diff
@@ -76,37 +78,176 @@ def __init__(
...
         ptype = user_defined_annotations.get(parameter_name, None)
+
+        if enforceABI and not SubroutineDefinition.check_if_ABI_annotation(ptype):
```
tzaffi, line 137:
```suggestion
        if enforce_ABI and not SubroutineDefinition.check_if_ABI_annotation(ptype):
```

pyteal/ast/subroutine.py
tzaffi: When would you _not_ want to `enforce_ABI` ?

pyteal/ast/subroutine_test.py
```diff
@@ -97,9 +104,27 @@ def fnWithMixedAnns2(a: ScratchVar, b) -> Expr:
...
+    def fnWithMixedAnns4(a: ScratchVar, b, c: abi.Uint16) -> Expr:
+        return Return()
+
+    def fnWithAllABIAnns(
+        a: abi.Uint16,
+        b: abi.StaticArray[abi.Uint32, Literal[10]],
+        c: abi.DynamicArray[abi.Bool],
+        *,
+        output: abi.Bool,
+    ):
```
tzaffi, line 116: As mentioned elsewhere, if at all possible, could we go with this format instead:
```python
    def fnWithAllABIAnns(
        a: abi.Uint16,
        b: abi.StaticArray[abi.Uint32, Literal[10]],
        c: abi.DynamicArray[abi.Bool],
    ) -> abi.Bool:
```

pyteal/ast/subroutine_test.py
```diff
@@ -171,7 +299,15 @@ def fnReturningExprSubclass(a: ScratchVar, b: Expr) -> Return:
         return Return()
 
     def fnWithMixedAnns4AndBytesReturn(a: Expr, b: ScratchVar) -> Bytes:
-        return Bytes("helo")
+        return Bytes("hello uwu")
+
+    def fnWithMixedAnnsABIRet1(
+        a: Expr, b: ScratchVar, c: abi.Uint16
+    ) -> abi.StaticArray[abi.Uint32, Literal[10]]:
```
tzaffi, line 306: Ok, I see the my preferred API is available, yay! But I'm still lobying to _disallow_ the other approach. I.e., don't allow `output` as a special keyword only variable.

tzaffi: Woops. I spoke too soon. This is a _negative_ example. Using to assert that an error has occurred

tzaffi (commented):
I'm finished my first round of reviews. I guess my main request is to change the way `output`'s are handled. This might very well contradict what you @jasonpaulos and /or @michaeldiamant have already come to an agreement about, in which case I still would like to have a discussion to understand why we prefer this API over what seems more natural to me.

pyteal/ast/subroutine.py
ahangsu: 👍

pyteal/ast/subroutine.py
ahangsu: 👍

pyteal/ast/subroutine.py
ahangsu: 👍

pyteal/ast/subroutine.py
ahangsu: 👍

pyteal/ast/subroutine_test.py
ahangsu: good point. I guess what we can do now is to set a reminder here indicating the situation (like a TODO in code snippet), in case we would forget about sync both implementations.

pyteal/ast/subroutine_test.py
tzaffi: (I meant with generics...)

ahangsu:
The current implementation requires an extra keyword argument ABI value to allow for extra composability for ABI return:
- The initial design was to `pass-by-ref`, let subroutine write to an ABI value (with the help of dynamic scratch var):
  - before such subroutine, we place an return-type ABI value's *scratch slot id* on stack
  - the subroutine internally takes the scratch slot id, and internally uses a dynamic scratch var to point at it, and build an ABI value over such dynamic scratch var.

  welp... sounds horrible enough, and a good (or bad) news is that, the method just described is not working.

  What compiler knows is that:
  - we are passing a scratch slot id to a subroutine, with no previous store operation
  - (something nebulous happening in subroutine)
  - after subroutine, load from the original ABI scratch slot
  - and this triggers `load-before-store` error.

  so what's up next?
- ABI argument is a python keyword argument, but it is *not* a TEAL level argument:
  - in `evaluate_subroutine` we can feed the subroutine such ABI value to keyword argument first, so we can construct the expression returned by subroutine.
  - let the expression returned by original subroutine definition be `expr`, we append `expr` with a return to construct `Seq(expr, Return(kwarg.load()))` to load the stored value of ABI kwarg-return value to stack.
    - NOTE: this requires `expr` is not allowed to have return, or execution cannot reach the return stored value line.
  - we then use `ReturnedValue` to wrap around to allow for `store_into` method to work, to assign subroutine ABI result to another value.

The description above is the thinking and design behind the current implementation (as of 2022/04/27).

> A side note: the current implementation looks conflicting with Zeph's Blackbox implementation: #249 and #297 .
>
> Need to cooperate to resolve and adapt their implementation.

pyteal/ast/subroutine.py
```diff
@@ -62,12 +60,16 @@ def __init__(
...
             abi_args,
-        ) = SubroutineDefinition._arg_types_and_by_refs(sig, annotations)
-        self.expected_arg_types: List[
-            Union[Type[Expr], Type[ScratchVar], abi.TypeSpec]
+            abi_output_kwarg,
+        ) = SubroutineDefinition._arg_types_and_by_refs(
+            sig, annotations, abi_output_arg_name
+        )
+        self.expected_arg_types: list[
+            Type[Expr] | Type[ScratchVar] | abi.TypeSpec
```
tzaffi, line 68:
```suggestion
            type[Expr] | type[ScratchVar] | abi.TypeSpec
```

ahangsu: one less thing to import :)

pyteal/ast/subroutine.py
```diff
@@ -105,9 +107,7 @@ def _validate_parameter_type(
         else:
             if not isclass(ptype) and not SubroutineDefinition.is_abi_annotation(ptype):
```
michaeldiamant, line 108: @ahangsu For my edification:  Do you understand why the following expression evaluates to `False`?

`print(isclass(pt.abi.StaticArray(pt.abi.ByteTypeSpec(), 10)))` 

* It's _slightly_ tangential to the PR.  I reviewed here while inspecting tests.  It took me some time to realize _some_ ABI arguments aren't treated as classes even though they're defined as classes.
* When I run a debugger locally, the debugger does _not_ let me step into `inspect.isinstance`.  So, I have limited visibility.
* The inline docs to `inspect.isclass` suggest classes provide `__doc__` and `__module__`.  The example _does_ provide these attributes.

pyteal/ast/subroutine.py
```diff
@@ -177,36 +184,50 @@ def _arg_types_and_by_refs(
...
             )
-            expected_arg_types.append(expected_arg_type)
+
+            if param.kind is Parameter.KEYWORD_ONLY:
+                if not isinstance(expected_arg_type, abi.TypeSpec):
+                    raise TealInputError(
+                        f"Function keyword parameter {name} has type {expected_arg_type}"
+                    )
+                abi_output_kwarg[name] = expected_arg_type
+                continue
```
michaeldiamant, line 194: @ahangsu As a I started reviewing, I made a few local changes.  I bundled them in a child PR for your consideration:  https://github.com/algorand/pyteal/pull/308.

Treat as optional.  Feel welcomed to tell me you'd prefer as is.

pyteal/ast/subroutine.py
ahangsu: The purpose of `_validate_parameter_type` is to check the annotation in python function definition.

`pt.abi.StaticArray(pt.abi.ByteTypeSpec(), 10))` is an instance not class.

`isclass` here is to allow for `Expr` or `ScratchVar` annotations.

some other type annotation on ABI side, e.g., `abi.StaticArray[abi.Uint64, Literal[10]]`, are generic alias. This is fed to `SubroutineDefinition.is_abi_annotation`.

tzaffi: Do you have an example of ABI args not treated as classes?

michaeldiamant: Sorry, I confused things with the (incorrect) example.

I began looking here because I deleted ` and not SubroutineDefinition.is_abi_annotation(ptype)` from the highlighted line to see what tests fail.

You _should_ see 3 test failures.  All failures involve a reference to `StaticArray` like the example failure message below.

```
E               pyteal.TealInputError: Function has parameter b of declared type pyteal.StaticArray[pyteal.Uint32, typing.Literal[10]] which is not a class
```

I opened the thread because it's unclear to me why `isclass` evaluation fails here.

ahangsu: let me illustrate more here... `StaticArray[abi.Byte, Literal[10]]` is *not* a class, it is a generic alias: https://docs.python.org/3/library/stdtypes.html#types-genericalias.

If you lift the condition `and not SubroutineDefinition.is_abi_annotation(ptype)`, it will be not class, and directly raise error.

ahangsu: The function name itself might be somewhat misleading: `_validate_parameter_type` should be renamed as `_validate_annotation_type` I guess?

pyteal/ast/subroutine.py
```diff
@@ -116,19 +116,20 @@ def _validate_parameter_type(
...
-                    "Function has parameter {} of disallowed type {}. Only the types {} are allowed".format(
-                        parameter_name, ptype, (Expr, ScratchVar, "ABI")
-                    )
+                    f"Function has parameter {parameter_name} of disallowed type {ptype}. "
+                    f"Only the types {(Expr, ScratchVar, 'ABI')} are allowed"
                 )
 
     @staticmethod
     def _arg_types_and_by_refs(
         sig: Signature,
```
tzaffi, line 125:
```suggestion
        sig: Signature,
```

pyteal/ast/subroutine.py
```diff
@@ -116,19 +116,20 @@ def _validate_parameter_type(
...
-        annotations: Dict[str, type],
-    ) -> Tuple[
-        List[Union[Type[Expr], Type[ScratchVar], abi.TypeSpec]],
-        Set[str],
-        Dict[str, abi.TypeSpec],
+        annotations: dict[str, type],
```
tzaffi, line 126: Consider converting this and others to `@classmethod`. This will allow calling via either `self.method()` in an instance method or `cls.method()` in another class method, instead of referring by `ClassName.method()`

pyteal/ast/subroutine.py
tzaffi: I see what you're saying. The line says `if not isclass(ptype) and not SubroutineDefinition.is_abi_annotation(ptype)` which to me feels incorrect. I would expect all `abi_annotation`'s to satisfy `isclass()` and therefore the second statement should be tautologically true.

pyteal/ast/subroutine.py
```diff
@@ -269,15 +292,39 @@ def has_return(self):
 SubroutineDeclaration.__module__ = "pyteal"
 
 
+@dataclass
+class _OutputKwArgInfo:
```
tzaffi, line 296: I apologize for introducing underscored class names during #249. This was later removed and unless you have a good reason, I recommend removing here as well to adhere with the conventions of this repo.

pyteal/ast/subroutine.py
```diff
@@ -269,15 +292,39 @@ def has_return(self):
...
+    name: str
+    abi_type: abi.TypeSpec
+
+    @staticmethod
+    def from_dict(kwarg_info: dict[str, abi.TypeSpec]) -> Optional["_OutputKwArgInfo"]:
+        if kwarg_info is None or len(kwarg_info) == 0:
```
tzaffi, line 302:
```suggestion
        if not kwarg_info:
```
this is equivalent

pyteal/ast/subroutine.py
```diff
@@ -269,15 +292,39 @@ def has_return(self):
...
+            return None
+        elif len(kwarg_info) == 1:
+            key = list(kwarg_info.keys())[0]
```
tzaffi, line 305:
```suggestion
            (key) = [*kwarg_info]
```
lil destructuring hack... feel free to ignore

pyteal/ast/subroutine.py
```diff
@@ -269,15 +292,39 @@ def has_return(self):
...
+
+
 class SubroutineCall(Expr):
     def __init__(
         self,
         subroutine: SubroutineDefinition,
-        args: List[Union[Expr, ScratchVar, abi.BaseType]],
+        args: list[Expr | ScratchVar | abi.BaseType],
+        *,
+        output_kwarg: Optional[_OutputKwArgInfo] = None,
```
tzaffi, line 322:
```suggestion
        output_kwarg: _OutputKwArgInfo = None,
```
I believe **mypy** is clever enough to understand that your type is `Optional` when defaulting to `None` (but don't trust me on that)

pyteal/ast/subroutine.py
```diff
@@ -374,15 +425,138 @@ def name(self) -> str:
...
 
 
+class ABIReturnSubroutine:
+    """Used to create a PyTeal Subroutine (returning an ABI value) from a python function.
+
+    This class is meant to be used as a function decorator. For example:
+
+        .. code-block:: python
+
+            @ABIReturnSubroutine
```
tzaffi, line 444: "Visual reminder" (as @michaeldiamant would say) to add this to our `user_guide_test()` (once the API is finalized)

pyteal/ast/subroutine.py
```diff
@@ -374,15 +425,138 @@ def name(self) -> str:
...
+    ) -> None:
+        self.output_kwarg_info: Optional[
+            _OutputKwArgInfo
+        ] = self._output_name_type_from_fn(fn_implementation)
+
+        internal_subroutine_ret_type = (
+            TealType.none
+            if self.output_kwarg_info is None
+            else self.output_kwarg_info.abi_type.storage_type()
+        )
```
tzaffi, line 480:
```suggestion
        internal_subroutine_ret_type = None
        if self.output_kwarg_info:
            internal_subroutine_ret_type = self.output_kwarg_info.abi_type.storage_type()
```
Totally optional, but you may find a bit more readable. Given that `_output_name_type_from_fn()` is guaranteed to return `None` or something `truthy` this is equivalent.

tzaffi: (ok, it's ever so slightly slower than the original)

pyteal/ast/subroutine.py
tzaffi: I only saw @ahangsu 's explanation after putting my question forth. I think I understand now.

As far as naming, maybe just `_validate_annotation()` ?

Final, question: Does it make sense for this to be a `@staticmtehod`, or would it simplify things to just make an in instance method and access `user_defined_annotations` from `self` ?

pyteal/ast/subroutine.py
tzaffi: I just remembered, `mypy` may not like that approach and you might need to specify the type. IE:
```python
user_defined_annotations: type = None # or something like that
if ...
```

pyteal/ast/subroutine.py
```diff
@@ -374,15 +425,138 @@ def name(self) -> str:
...
+
+    @staticmethod
+    def _output_name_type_from_fn(
+        fn_implementation: Callable[..., Expr]
+    ) -> Optional[_OutputKwArgInfo]:
+        if not callable(fn_implementation):
+            raise TealInputError("Input to ABIReturnSubroutine is not callable")
+        sig = signature(fn_implementation)
+        fn_annotations = getattr(fn_implementation, "__annotations__", OrderedDict())
```
tzaffi, line 500:
```suggestion
        fn_annotations = get_annotations(fn_implementation)
```
for py 3.10 this is preferred (assuming `get_annotations` has been imported from `inspect`)

pyteal/ast/subroutine.py
```diff
@@ -374,15 +425,138 @@ def name(self) -> str:
...
+
+        potential_abi_arg_names = list(
+            filter(
+                lambda key: sig.parameters[key].kind == Parameter.KEYWORD_ONLY,
+                sig.parameters.keys(),
+            )
+        )
+        if len(potential_abi_arg_names) == 0:
```
tzaffi, line 508:
```suggestion
        if not potential_abi_arg_names:
```

pyteal/ast/subroutine.py
```diff
@@ -446,53 +620,93 @@ def evaluateSubroutine(subroutine: SubroutineDefinition) -> SubroutineDeclaratio
...
-        loaded: Union[ScratchVar, abi.BaseType, Expr]
-        argVar: ScratchVar
+    ) -> tuple[ScratchVar, ScratchVar | abi.BaseType | Expr]:
+        _loaded_var: ScratchVar | abi.BaseType | Expr
+        _argument_var: ScratchVar
 
         if param in subroutine.by_ref_args:
-            argVar = DynamicScratchVar(TealType.anytype)
-            loaded = argVar
+            _argument_var = DynamicScratchVar(TealType.anytype)
```
tzaffi, line 649: nit: these local variables need not be underscored

pyteal/ast/subroutine.py
```diff
@@ -446,53 +620,93 @@ def evaluateSubroutine(subroutine: SubroutineDefinition) -> SubroutineDeclaratio
...
+            _loaded_var = _argument_var
         elif param in subroutine.abi_args:
             internal_abi_var = subroutine.abi_args[param].new_instance()
```
tzaffi, line 652: `internal_abi_var` is unsafe because it only exists for the abi case, but there could be other cases that unwittingly attempt to access it.

pyteal/ast/subroutine.py
```diff
@@ -446,53 +620,93 @@ def evaluateSubroutine(subroutine: SubroutineDefinition) -> SubroutineDeclaratio
...
+
+    argument_vars: list[ScratchVar] = []
+    loaded_args: list[ScratchVar | Expr | abi.BaseType] = []
+    for arg in args:
+        argument_var, loaded_arg = var_n_loaded(arg)
+        argument_vars.append(argument_var)
+        loaded_args.append(loaded_arg)
+
+    abi_output_kwargs: dict[str, abi.BaseType] = {}
+    if len(subroutine.output_kwarg) > 1:
```
tzaffi, line 672: I recommend putting this validation at the top so you can "fail fast"

tzaffi (commented):
Overall, this looks like a solid and clever approach. I made various suggestions for small improvements most of which are optional.

pyteal/ast/subroutine.py
ahangsu: Yes and yes I think, I will rename this to be `_validate_annotation`, and probably I should make this a simple instance method, so we can simplify some function arguments.

ahangsu: I am also reconstruct the function a little bit on my local side, so the readability would be better.

pyteal/ast/subroutine.py
michaeldiamant: @ahangsu Thanks for bringing in #308 - Resolving.

pyteal/ast/subroutine.py
```diff
@@ -374,15 +427,138 @@ def name(self) -> str:
...
 
     def has_return(self):
-        return self.subroutine.getDeclaration().has_return()
+        return self.subroutine.get_declaration().has_return()
 
 
 SubroutineFnWrapper.__module__ = "pyteal"
 
 
+class ABIReturnSubroutine:
```
michaeldiamant, line 439: @ahangsu and reviewers:  Since `ABIReturnSubroutine` is primarily (as I understand it) intended for ARC-4 method definition, I'm wondering how folks feel about matching the name with the use case?

* For example, I can imagine naming it `ABIMethod` in hopes of making the intended usage clearer to PyTeal developers, and providing consistency with https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0004.md#methods.
* On the other hand, it's possible that we _prefer_ a more general name in case use cases change over time.
* My starting 2c:  
  * I'd name it for the use case we have today (something like `ABIMethod`) because I think it's easier to generalize later rather than become increase specificity later. 
  * I'll follow the majority opinion here including keeping as is.  Since name changes are difficult (due to backwards compatibility concerns), I raise the question to sanity check the rubric applied to choose a name.

tzaffi: I agree that it's better to name it based on it's prominent use case. It may, in fact, provide more functionality, but that's not a hindrance in my opinion (e.g. python's `list` has additional methods that let you use it as a **Stack** and this is just fine). An additional benefit is that the more specific the name, the easier it is to grasp.

Upshot: I support renaming to `ABIMethod`

pyteal/ast/subroutine.py
```diff
@@ -230,7 +253,9 @@ def invoke(
                         f"should have ABI typespec {arg_type} but got {arg.type_spec()}"
                     )
 
-        return SubroutineCall(self, args)
+        return SubroutineCall(
+            self, args, output_kwarg=_OutputKwArgInfo.from_dict(self.output_kwarg)
```
michaeldiamant, line 257: @ahangsu I opened an optional PR for your review based on local changes I made while reviewing `output_kwarg`.  I'd appreciate you taking a look as time allows - https://github.com/algorand/pyteal/pull/311.

Treat as optional and feel welcomed to keep as is.

pyteal/ast/subroutine.py
michaeldiamant: @ahangsu and @tzaffi thanks for your efforts to respond to my question. 

* I now understand the situation.  I misunderstood that `StaticArray` takes on `GenericAlias` at runtime.  
* I briefly paired with @tzaffi to deepen my understanding here and ended up with _optional_ changes https://github.com/algorand/pyteal/pull/312.  @ahangsu I'd appreciate a review as time allows.  And feel welcomed to keep as is.

pyteal/ast/subroutine.py
tzaffi: please note that #310 proposes to consolidate the validation logic into `_validate()` and so this function would not exist in that proposal.

pyteal/ast/subroutine.py
michaeldiamant: Resolving - We agreed to close the child PR.

pyteal/ast/subroutine.py
ahangsu: resolved by #308

pyteal/ast/subroutine.py
ahangsu: resolved by #308

pyteal/ast/subroutine.py
ahangsu: resolved by match syntax

pyteal/ast/subroutine.py
```diff
@@ -374,15 +425,138 @@ def name(self) -> str:
...
+            raise TealInputError("Input to ABIReturnSubroutine is not callable")
+        sig = signature(fn_implementation)
+        fn_annotations = getattr(fn_implementation, "__annotations__", OrderedDict())
+
+        potential_abi_arg_names = list(
+            filter(
+                lambda key: sig.parameters[key].kind == Parameter.KEYWORD_ONLY,
+                sig.parameters.keys(),
+            )
+        )
```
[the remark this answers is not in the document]
ahangsu: good point

pyteal/ast/subroutine.py
ahangsu: good point

pyteal/ast/subroutine.py
ahangsu: fixed

pyteal/ast/subroutine.py
ahangsu:
```python
        internal_subroutine_ret_type = TealType.none
        if self.output_kwarg_info:
            internal_subroutine_ret_type = (
                self.output_kwarg_info.abi_type.storage_type()
            )
```
current implementation resolves it

pyteal/ast/subroutine.py
ahangsu: good point

pyteal/ast/subroutine.py
ahangsu: fixed

pyteal/ast/subroutine.py
ahangsu: I am `1 - 9/19` supporting renaming this to `ABIMethod`:
- length of `Subroutine` is 10
- length of `ABIReturnSubroutine` is 19
- length of `ABIMethod` is 9

SWE's finger-joint life matters! If we can reach to a consensus that we want to rename this, I will do it.

pyteal/ast/subroutine.py
ahangsu: I removed the leading underline anyways.

pyteal/ast/subroutine.py
tzaffi: 🍾

pyteal/ast/subroutine_test.py
```diff
@@ -237,13 +331,204 @@ def fnWithMixedAnns4(a: pt.ScratchVar, b, c: pt.abi.Uint16) -> pt.Expr:
...
+            "mixed with ABI annotations 1 with ABI return wrong inputs",
+            fn_mixed_annotation_1_with_ret,
+            [expr_int, abi_static_u32_10],
+            None,
+            pt.TealInputError,
+        ),
+    ]
+
+    for case_name, impl, args, ret_type, err in cases:
+        definition = pt.ABIReturnSubroutine(impl)
```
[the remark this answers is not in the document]
ahangsu: good point

pyteal/ast/subroutine.py
michaeldiamant: Had a verbal with @tzaffi, @ahangsu, @algoidurovic, and @jannotti to discuss naming and landed on keeping as is.  

`ABIMethod` implies the method is auto-registered for routing.  Since auto-registration is _not_ provided, the name doesn't fit well.

Resolving the thread because it feels like we reached agreement.  If folks that weren't in the discussion feel otherwise, feel welcomed to reopen.

michaeldiamant (approved):
@ahangsu Thanks for the effort here - big step forward in functionality!  ☕ 

Leaving approval contingent on these points:
* https://github.com/algorand/pyteal/pull/315 is merged before merging this PR.
* We agree to defer user guide updates beyond this PR until https://github.com/algorand/pyteal/pull/170 looks code complete.

tzaffi (approved):
LGTM

pyteal/ast/subroutine.py
```diff
@@ -389,15 +453,138 @@ def name(self) -> str:
...
+            case []:
+                return None
+            case [name]:
+                annotation = fn_annotations.get(name, None)
+                if annotation is None:
+                    raise TealInputError(
+                        f"ABI subroutine output-kwarg {name} must specify ABI type"
+                    )
+                type_spec = abi.type_spec_from_annotation(annotation)
+                return OutputKwArgInfo(name, type_spec)
```
jasonpaulos, line 545: I originally thought we'd force all functions to use the "output" keyword argument, but I see this implementation allows you to pick any name. Did you have uses cases or benefits in mind when you made this decision? I'm trying to evaluate whether it's better or worse than fixing the argument name.

pyteal/ast/subroutine.py
```diff
@@ -461,53 +648,93 @@ def evaluateSubroutine(subroutine: SubroutineDefinition) -> SubroutineDeclaratio
...
+
+    if len(subroutine.output_kwarg) > 1:
+        raise TealInputError(
+            f"ABI keyword argument num: {len(subroutine.output_kwarg)}. "
+            f"Exceeding abi output keyword argument max number 1."
+        )
 
     args = subroutine.arguments()
-    argumentVars, loadedArgs = zip(*map(var_n_loaded, args)) if args else ([], [])
+    args = [arg for arg in args if arg not in subroutine.output_kwarg]
```
jasonpaulos, line 696: It looks like the `SubroutineDefinition.arguments()` function already filters out the keyword arg, so this is unnecessary?

pyteal/ast/subroutine.py
```diff
@@ -461,53 +648,93 @@ def evaluateSubroutine(subroutine: SubroutineDefinition) -> SubroutineDeclaratio
...
+    if not isinstance(subroutine_body, Expr):
         raise TealInputError(
-            f"Subroutine function does not return a PyTeal expression. Got type {type(subroutineBody)}"
+            f"Subroutine function does not return a PyTeal expression. Got type {type(subroutine_body)}."
+        )
+    # if there is an output keyword argument for ABI, place the storing on the stack
+    if output_carrying_abi:
+        if subroutine_body.has_return():
+            raise TealInputError(
+                "ABI returning subroutine definition should have no return"
```
jasonpaulos, line 724: I _think_ you're choosing to error in this case because it would be impossible to add the output ABI value to the stack if the subroutine has a retsub op?

FYI, `has_return()` is only true if _every_ possible codepath in the expression results in a return. So if only some paths end in a return, it would be false, meaning that case isn't addressed here.

Separately, I think it _should_ be possible for users to use the `Return()` expression in a subroutine, since it's a useful feature. But it looks like supporting this might be more difficult than I realized.

pyteal/ast/subroutine.py
```diff
@@ -325,14 +385,17 @@ def __teal__(self, options: "CompileOptions"):
 
         3. (ABI, or a special case in by-value) In this case, the storage of an ABI value are loaded
             to the stack and will be stored in a local ABI value for subroutine evaluation
+
+        4. (ABI output keyword argument, or by-ref ABI value) In this case of returning ABI values, we do not place
```
jasonpaulos, line 389: nit: ABI output arguments aren't really "by-ref", since they're ouput only

pyteal/ast/subroutine.py
```diff
@@ -461,53 +648,93 @@ def evaluateSubroutine(subroutine: SubroutineDefinition) -> SubroutineDeclaratio
...
+            raise TealInputError(
+                "ABI returning subroutine definition should have no return"
+            )
+        if subroutine_body.type_of() != TealType.none:
+            raise TealInputError(
+                f"ABI returning subroutine definition should evaluate to TealType.none, "
+                f"while evaluate to {subroutine_body.type_of()}."
+            )
+        subroutine_body = Seq(
+            subroutine_body, Return(output_carrying_abi.stored_value.load())
```
jasonpaulos, line 732: Do you think we should worry about if a user never stores anything in the output variable? Or does PyTeal's built-in scratch slot load-before-store logic already prevent this?

In any case, do you think it should or shouldn't be allowed?

jasonpaulos (commented):
It would be nice to see end to end tests as well (meaning a comparison of the generated TEAL, `pyteal/compiler/compiler_test.py` has some good examples for subroutines already).

Some edge cases that I'm interested in seeing are:
- Subroutine body with a conditional return (i.e. `if (x) { return; } else { ... }`)
- Subroutine body that accesses the output variable before storing into it
- Subroutine body that never stores anything in the output variable

pyteal/ast/subroutine.py
```diff
@@ -108,35 +104,58 @@ def _validate(
...
         by_ref_args: set[str] = set()
         abi_args: dict[str, abi.TypeSpec] = {}
+        abi_output_kwarg: dict[str, abi.TypeSpec] = {}
 
-        if input_types is not None and len(input_types) != len(implementation_params):
-            raise TealInputError(
-                f"Provided number of input_types ({len(input_types)}) "
-                f"does not match detected number of parameters ({len(implementation_params)})"
-            )
+        if input_types:
```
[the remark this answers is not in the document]
tzaffi: 👍  (and apologies if I suggested trimming this down)

pyteal/ast/subroutine_test.py
```diff
@@ -243,72 +392,21 @@ def one_dynscratchvar_impl(x: pt.DynamicScratchVar):
...
+            output_kwarg,
+        ) = three_params._validate(input_types=[x, y, z])
         assert len(params) == 3
         assert anns == {}
         assert all(at is pt.Expr for at in arg_types)
         assert byrefs == set()
-        assert abis == {}
-
-    # annotation / abi type handling:
-    abi_annotation_examples = {
```
tzaffi, line 259: These cases are no longer necessary, or they don't make sense anymore? From my experience, erroring on the side of redundancy is preferred over under-coverage.

tzaffi: But if keeping these cases made the merge too difficult, and there is comparable coverage elsewhere, then it makes sense to drop them.

pyteal/ast/subroutine.py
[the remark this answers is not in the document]
tzaffi: I think it's ok to put this issue on the backburner.  `input_types` validation is only used by `Blackbox` and my upcoming PR #322 will need to sort this out.

pyteal/ast/subroutine_test.py
```diff
@@ -475,13 +573,217 @@ def fnWithMixedAnns4(a: pt.ScratchVar, b, c: pt.abi.Uint16) -> pt.Expr:
...
+            (intermediate := pt.abi.Uint32()).set(b[a.load() % pt.Int(10)]),
+            a.store(intermediate.get()),
+            pt.Return(),
+        )
+
+    @pt.ABIReturnSubroutine
+    def fn_mixed_annotation_1_with_ret(
+        a: pt.ScratchVar, b: pt.abi.Uint64, *, c: pt.abi.Bool
+    ) -> pt.Expr:
+        return c.set((a.load() + b.get()) % pt.Int(2))
```
tzaffi, line 633: I recommend adding another erroring test case with 2 keyword args. EG:

```python
    @pt.ABIReturnSubroutine
    def fn_with_too_many_outputs(
        a: pt.ScratchVar, b: pt.abi.Uint64, *, c: pt.abi.Bool, d: pt.abi.Bool
    ) -> pt.Expr:
        return Int(1)
```

pyteal/ast/subroutine.py
ahangsu: yes, I believe this is a segment in an early implementation, going to take that out.

Good catch

pyteal/ast/subroutine.py
ahangsu: yes that is also a segment from the earlier implementation 🤦 should have modify that too

Good catch

pyteal/ast/subroutine_test.py
tzaffi: Great, I missed that example. Thank you for point it out.

Resolving.

pyteal/ast/subroutine.py
```diff
@@ -147,35 +166,44 @@ def _validate(
...
-                if not isinstance(intype, TealType):
+            expected_arg_type = self._validate_annotation(annotations, name)
+
+            if param.kind is Parameter.KEYWORD_ONLY:
+                if not isinstance(expected_arg_type, abi.TypeSpec):
                     raise TealInputError(
-                        f"Function has input type {intype} for parameter {name} which is not a TealType"
+                        f"Function keyword parameter {name} has type {expected_arg_type}"
                     )
+                abi_output_kwarg[name] = expected_arg_type
```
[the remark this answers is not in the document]
ahangsu: this case is eliminated by the condition check at be begin of for-iteration: It checks if any argument other than `POSITION_ONLY/POSITION_OR_KEYWORD`, if name not matching `self.abi_output_arg_name`, then error. 
(There is (at most) one str in `self.abi_output_arg_name: str`.)

The result of such checking allows at most one value in `abi_output_kwarg`.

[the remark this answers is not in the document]
jasonpaulos: Thanks, that makes sense

pyteal/ast/subroutine_test.py
ahangsu: let me look into this... PRs are clobbing each others so I might over-deleted something 🤦

pyteal/ast/subroutine.py
ahangsu: This is a good point of thinking.

I do think it is not allowed to have output kwarg, and not storing anything inside. Essentially this will incur load before store error in ABI implementation.

While people can go down the path, and cause error from ABI (and current subroutine implementation), do you think we want to additionally check this and provide a specific error for defining output kwarg but not using it?

pyteal/ast/subroutine.py
ahangsu: good point, I was assuming that `has_return()` is for *existence* of return expression in expr, now it seems wrong.

I was going down this path, and append `Return(output.get())` to the subroutine body.

Now it seems we should let users to write `Return(output.get())` themselves in the expr, and we still need to check `has_return` to make sure that the subroutine is well-formed, i.e., all code paths returns.

Does that sound reasonable to you?

ahangsu: Think a little bit more of your suggestion, and I realized that one can probably just directly return the underlying representation of ABI value, i.e., `Return(Int(1))`, for an output of `abi.Uint64`.

Feels somewhat hacky, but seems okay to me. Any thoughts on it?

pyteal/ast/subroutine.py
jasonpaulos: Got it, as long as there is some sort of error, I think that's acceptable. I want to avoid reading from an uninitialized scratch slot, since that will produce a value that doesn't make sense, and it sounds like our current scratch slot load-before-store error is good enough.

pyteal/ast/subroutine.py
jasonpaulos: Ok, I see. My current thoughts are that we should be more conservative initially and limit the name to "output" or something similar that makes it clear the argument is meant to capture the return value.

Then going forward we can always drop this restriction without jeopardizing existing code.

ahangsu: ok I see. I added a few lines of restrictions on naming for kwarg in `ABIReturnSubroutine`.

LMK if it looks right

pyteal/ast/subroutine_test.py
tzaffi: Thanks for bringing them back, I hope it wasn't too much pain to do so.  I will probably need to upgrade these in my next PR, but these look good for now. I will approve as all my questions have been addressed.

tzaffi (approved):
LGTM

ahangsu:
good point on Jason's suggestion on compiler test, i.e., comparison on generated TEAL.

I think we probably wait on a better design for conditional-return, assuming it is arriving soon? Then I can proceed to such test without reverting implementation too much.

pyteal/ast/subroutine.py
ahangsu: ok, seems to be resolving in #328

pyteal/ast/subroutine.py
```diff
@@ -154,8 +148,8 @@ def _validate(
                 Parameter.POSITIONAL_OR_KEYWORD,
             ) and not (
                 param.kind is Parameter.KEYWORD_ONLY
-                and self.abi_output_arg_name is not None
-                and name == self.abi_output_arg_name
+                and self.has_abi_output
+                and name == ABIReturnSubroutine.OUTPUT_ARG_NAME
```
tzaffi, line 152:
```suggestion
                and name == self.OUTPUT_ARG_NAME
```
a bit more future-proof / pythonic

tzaffi: Ahhh I get it. Sorry

jasonpaulos (approved):
Looks great! It would be nice if you could resolve the outdated comment I pointed out, otherwise this is good to merge!

pyteal/ast/subroutine.py
```diff
@@ -169,6 +162,9 @@ def _validate(
             expected_arg_type = self._validate_annotation(annotations, name)
 
             if param.kind is Parameter.KEYWORD_ONLY:
+                # this case is only entered when
+                # - `self.abi_output_arg_name is not None`
+                # - `name == self.abi_output_arg_name`
```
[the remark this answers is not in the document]
ahangsu: yes, updating to `self.has_abi_output is True`

pyteal/ast/abi/type.py
```diff
@@ -182,3 +182,40 @@ def use(self, action: Callable[[T], Expr]) -> Expr:
...
+            )
+
+        declaration = self.computation.subroutine.get_declaration()
+
+        if declaration.deferred_expr is None:
+            raise TealInputError(
+                "ABI return subroutine must have deferred_expr to be not-None."
+            )
+        if declaration.deferred_expr.type_of() != output.type_spec().storage_type():
+            raise TealInputError(
```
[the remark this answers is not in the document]
ahangsu: I think the check is needed.

Why we need to trouble `stored_value.slot.store` not `stored_value.store`?
Since `self.computation` is evaluated to `TealType.none`, but latter option does the type check to align the type of `self.computation` to scratchvar's type.

The "runtime behavior of" `self.computation` type is ABI's scratch var's type, which is equal to `deferred_expr`'s type, and this is the missing type check in `stored_value.slot.store`.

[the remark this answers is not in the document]
jasonpaulos: Great point, I misunderstood this. Resolving

PR #259, opened 2022-03-23
Standardize terminology from semantic E2E to blackbox

michaeldiamant:
Standardizes terminology by replacing _semantic E2E_ with _blackbox_.

Rationale:
* What I'm seeing:
  * https://github.com/algorand/py-algorand-sdk/pull/298 introduces the terminology _blackbox_ to mean asserting on a smart contract via dryrun.  
  * #249 uses the terms _semantic_ and _E2E_ in ways that I interpret to mean _blackbox_.  
* Since it seems like multiple terms are applied with the same intent, I felt it'd reduce the chance for confusion to consistently apply terminology.  Since py-algorand-sdk is upstream, I opted for _blackbox_.  
* Consider the PR optional.  
  * If preferred to keep as is, then I'll happily close the PR.
  * I'm also open to feedback on further naming customization.  I feel like even partial movement towards unifying terminology is likely to be worthwhile.


tzaffi (approved):

tzaffi:
thx again!

PR #260, opened 2022-03-23
Run blackbox tests in CI

michaeldiamant:
Defines `Makefile` facilities to run blackbox tests from #249 as part of the build.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -65,7 +65,7 @@ jobs:
         with:
           key: docker-layer-caching-${{ github.workflow }}-${{ hashFiles(env.CONFIG, env.GENESIS) }}-{hash}
           restore-keys: docker-layer-caching-${{ github.workflow }}-${{ hashFiles(env.CONFIG, env.GENESIS) }}-
-#      - name: Install required os level applications
+#      - name: Install required OS level applications
 #        run: |
 #         # https://serverfault.com/a/992421
 #          export DEBIAN_FRONTEND=noninteractive
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -39,7 +39,8 @@ jobs:
       - name: Install required os level applications
         run: |
           apt update -y
-          apt install -y git make
+          apt install -y git make python3-pip
+          pip3 install docker-compose
       - name: Clone sandbox
         uses: actions/checkout@v2
         with:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -41,6 +41,12 @@ jobs:
           sudo apt update -y
           sudo apt install -y curl git nodejs python-is-python3 python3-pip
           pip install docker-compose
+          sudo apt -y install ca-certificates curl gnupg lsb-release
+          sudo curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo gpg --dearmor -o /usr/share/keyrings/docker-archive-keyring.gpg
+          sudo echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/docker-archive-keyring.gpg] https://download.docker.com/linux/ubuntu \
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -28,6 +28,47 @@ jobs:
         run: make pip
       - name: Build and Test
         run: make build-and-test
```
michaeldiamant, line 30: I'm mostly out of ideas here.  

Locally via https://github.com/nektos/act, it works at these commits:
* https://github.com/algorand/pyteal/pull/260/commits/e1ffdadcfaf540686b8d06ea4528ae4cb1461b58
* https://github.com/algorand/pyteal/pull/260/commits/4f269a3a1a63ce6679a0e8187ffb90132b69cb7b
* https://github.com/algorand/pyteal/pull/260/commits/2e8433b0535ae20d2a1004ec564aa244d0e844e9

It's unclear to me how to otherwise debug a failed invocation (e.g. https://github.com/algorand/pyteal/runs/5664812490?check_suite_focus=true).  It's quite possible I'm taking an approach that's a poor fit.

tzaffi: Thanks for suggesting `act`. Really useful!!!

.github/workflows/build.yml
```diff
@@ -28,6 +33,65 @@ jobs:
...
+    # container: ubuntu:22.04
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Install required os level applications
+        run: |
+          sudo apt update -y
+          sudo apt install -y curl git nodejs python-is-python3 python3-pip
```
michaeldiamant, line 47: As part of bringing these changes to a mergeable state, I think we ought to disambiguate steps needed for act.

* In order to get act to pass, I sourced various dependencies that I believe are _not_ needed for github.
* Disambiguation here means creating 1 or more steps dedicated to act dependencies.  Goal is to make it easier to maintain the build.
* I _think_ the following dependencies are act-specific though it requires a round of testing:
  * `curl`, `nodejs`, `python3-pip`
  * Docker install (e.g. `docker-ce`)
  * _Setup docker compose_

tzaffi:
Bringing a comment over from #249 that's better placed here:

> Sync'ed up offline but let's summarize my wish list regarding testing (with the caveat that clearly "testing" is better than "no testing").

> From most desired to least desired (but again, even least desired is really great):

> 1. have a dry run test that doesn't require running a node (a la goal clerk dryrun without the remote option)
> 2. have the dry run test that uses a quickly downloadable pre-built docker image (a la docker pull algorand/stable)
> 3. have a dry run that stands up the minimum necessary sandbox-like environment
> 4. have a dry run that stands up a "heavier" sandbox environment similar to the one that the SDK's do

tzaffi:
> Bringing a comment over from #249 that's better placed here:
> 
> > Sync'ed up offline but let's summarize my wish list regarding testing (with the caveat that clearly "testing" is better than "no testing").
> 
> > From most desired to least desired (but again, even least desired is really great):
> 
> > 1. have a dry run test that doesn't require running a node (a la goal clerk dryrun without the remote option)
> > 2. have the dry run test that uses a quickly downloadable pre-built docker image (a la docker pull algorand/stable)
> > 3. have a dry run that stands up the minimum necessary sandbox-like environment
> > 4. have a dry run that stands up a "heavier" sandbox environment similar to the one that the SDK's do

From my understanding the present solution is an optimized version of **3.**. A simple sandbox is downloaded but it's cached so -in practice- it's almost as good as **2.**

Also, we see from experience that it's taking around 5 minutes to run the tests. It's slower than the 1 minute times we had before the test, but not nearly as bad as the wait times on the SDK's and on `go-algorand`.

michaeldiamant:
Closing - We've evolved these techniques via https://github.com/algorand/graviton/pull/1.  #249 will directly bring in the CI test running.

Issue #242, opened 2022-03-11
ScratchVar:  Disallow @Subroutine recursion


michaeldiamant:
## Problem

Unintentionally, https://github.com/algorand/pyteal/pull/198 opens the door to unintended interactions between `ScratchVar` and recursive `@Subroutine`s.  #230 explores the topic.

We must either implement an improvement or revert #198 to avoid releasing software with known unintended behavior.

## Solution

Fail TEAL source compilation when a recursive `@Subroutine` containing a `ScratchVar` parameter exists.

Rationale:
* Since `DynamicScratchVar` is a new abstraction and we don't know how pervasively it'll be used, the ticket requests constraining usage until we get feedback otherwise.
* We discussed these alternatives and deemed the benefit-vs-cost tradeoff to be weaker than the proposed solution:
  * Support for recursion.
  * Add guard rails around allowed recursion cases.

Additionally:
* Add an inline comment explaining that no technical barrier exists to supporting #232, though we're opting to defer support.  Goal is to help future maintainers understand why the constraint exists.

## Dependencies

N/A

## Urgency

High - `master` cannot be safely tagged.

tzaffi:
#261 resolves this IMO

tzaffi:
this has been released https://github.com/algorand/pyteal/releases/tag/v0.11.0

PR #235, opened 2022-03-05
Closed: ScratchVar: Disallow @Subroutine Recursion

tzaffi:
This as been subsumed by #261

Issue #199, opened 2022-02-17
Quasi-bug: TEAL Output Depends not Only on Teal But on Python's Hash in Unreliable Ways


tzaffi:
# TEAL Output Should be Stable When No Changes are Made to the Actual Compilation

## Why is this useful

This would allow us to have less flaky end-to-end tests and assert that nothing about the PyTEAL compiler has affected between versions.

For example, consider this PyTEAL:
```python
@Subroutine(TealType.none)
def swap(x: ScratchVar, y: ScratchVar):
    z = ScratchVar(TealType.anytype)
    return Seq(
        z.store(x.load()),
        x.store(y.load()),
        y.store(z.load()),
    )

@Subroutine(TealType.none)
def cat(x, y):
    return Pop(Concat(x, y))


def swapper():
    a = ScratchVar(TealType.bytes)
    b = ScratchVar(TealType.bytes)
    return Seq(
        a.store(Bytes("hello")),
        b.store(Bytes("goodbye")),
        cat(a.load(), b.load()),
        swap(a, b),
        Assert(a.load() == Bytes("goodbye")),
        Assert(b.load() == Bytes("hello")),
        Int(1000),
    )
```

I was able to generate the following 2 equivalent TEAL programs by changing some python code that didn't touch the actual compilation code:

### Before Change
```
#pragma version 6
byte "hello"
store 0          
byte "goodbye"
store 1          
load 0
load 1
callsub cat_1
int 0
int 1
callsub swap_0   
load 0           
byte "goodbye"
==
assert           
load 1           
byte "hello"
==
assert           
int 1000         
return

// swap
swap_0:
store 3          
store 2          
load 2           
loads            
store 4          
load 2           
load 3           
loads            
stores           
load 3           
load 4           
stores           
retsub

// cat
cat_1:
store 6
store 5
load 5
load 6
concat
pop
retsub
```

### After Change

```
#pragma version 6
byte "hello"
store 5
byte "goodbye"
store 6
load 5
load 6
callsub cat_1
int 5
int 6
callsub swap_0
load 5
byte "goodbye"
==
assert
load 6
byte "hello"
==
assert
int 1000
return

// swap
swap_0:
store 1
store 0
load 0
loads
store 2
load 0
load 1
loads
stores
load 1
load 2
stores
retsub

// cat
cat_1:
store 4
store 3
load 3
load 4
concat
pop
retsub
```

### Steps to reproduce - this is non-trivial

1.  Compile a big program with several subroutines
[...]

PR #261, opened 2022-03-24
ScratchVar: Disallow @Subroutine Recursion

tzaffi:
## TODO's of issue #242 

- [x] Recursive by-ref treatment: Impose more constraints - Disallow all recursion for any ScratchVar in any `@Subroutine`
- [x] DynamicScratchVar treatment: Leave as is with explanatory inline comment explaining why it's like that
- [x] Break down associated integration tests in `tests/pass_by_ref_test.py` to smaller components
- [x] Add documentation regarding illegality of pass-by-ref recursion
- [x] Add a simple unit test to `pyteal/compiler/subroutines_test.py` that tests the new functionality of `spillLocalSlotsDuringRecursion`

### DIFF
--- a/pyteal/compiler/subroutines.py
+++ b/pyteal/compiler/subroutines.py
@@ -3,6 +3,7 @@ from typing import List, Dict, Set, Optional, TypeVar
 from collections import OrderedDict
 from itertools import chain
 
+from ..errors import TealInputError
 from ..types import TealType
 from ..ast import SubroutineDefinition
 from ..ir import TealComponent, TealOp, Op
@@ -62,6 +63,32 @@ def findRecursionPoints(
     return reentryPoints
 
 
+def find_recursive_path(
+    subroutine_graph: Dict[SubroutineDefinition, Set[SubroutineDefinition]],
+    subroutine: SubroutineDefinition,
+) -> List[SubroutineDefinition]:
+    visited = set()
+    loop = []
+
+    def dfs(x):
+        if x in visited:
+            return False
+
+        visited.add(x)
+        loop.append(x)
+        for y in subroutine_graph[x]:
+            if y == subroutine:
+                loop.append(y)
+                return True
+            if dfs(y):
+                return True
+        loop.pop()
+        return False
+
+    found = dfs(subroutine)
+    return loop if found else []
+
+
 def spillLocalSlotsDuringRecursion(
     version: int,
     subroutineMapping: Dict[Optional[SubroutineDefinition], List[TealComponent]],
@@ -89,6 +116,23 @@ def spillLocalSlotsDuringRecursion(
             the local slots which must be spilled for each subroutine.
     """
     recursivePoints = findRecursionPoints(subroutineGraph)
+
+    recursive_byref = None
+    for k, v in recursivePoints.items():
+        if v and k.by_ref_args:
+            recursive_byref = k
+            break
+
+    if recursive_byref:
+        msg = "ScratchVar arguments not allowed in recursive subroutines, but a recursive call-path was detected: {}()"
+        raise TealInputError(
+            msg.format(
+                "()-->".join(
+                    f.name()
+                    for f in find_recursive_path(subroutineGraph, recursive_byref)
+                )
+            )
+        )
 
     coverAvailable = version >= Op.cover.min_version
 
--- a/pyteal/compiler/subroutines.py
+++ b/pyteal/compiler/subroutines.py
@@ -12,7 +12,7 @@ from ..ir import TealComponent, TealOp, Op
 Node = TypeVar("Node")
 
 
-def depthFirstSearch(graph: Dict[Node, Set[Node]], start: Node, end: Node) -> bool:
+def graph_search(graph: Dict[Node, Set[Node]], start: Node, end: Node) -> bool:
     """Check whether a path between start and end exists in the graph.
 
     This works even if start == end, in which case True is only returned if the
@@ -57,7 +57,7 @@ def findRecursionPoints(
         reentryPoints[subroutine] = set(
             callee
             for callee in subroutineGraph[subroutine]
-            if depthFirstSearch(subroutineGraph, callee, subroutine)
+            if graph_search(subroutineGraph, callee, subroutine)
         )
 
     return reentryPoints
--- a/tests/pass_by_ref_test.py
+++ b/tests/pass_by_ref_test.py
@@ -96,27 +96,6 @@ def sub_mixed():
     return mixed_annotations(x, y, z)
 
 
-@Subroutine(TealType.none)
-def plus_one(n: ScratchVar):
-    tmp = ScratchVar(TealType.uint64)
-    return (
-        If(n.load() == Int(0))
-        .Then(n.store(Int(1)))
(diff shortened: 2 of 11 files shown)


tzaffi:
Continuing the convo from #235 

image

I renamed the function to `graph_search()`

tzaffi:
continuing discussion from #235 

image

Though this is a better test design you are recommending, I'm going to go with Pareto's Principle and claim that the 20% of the benefit gained from this improvement will require 80% of the effort. So I'll just keep the test as is.

pyteal/compiler/subroutines.py
```diff
@@ -11,7 +12,7 @@
 Node = TypeVar("Node")
 
 
-def depthFirstSearch(graph: Dict[Node, Set[Node]], start: Node, end: Node) -> bool:
```
tzaffi, line 14: renaming the function

tests/pass_by_ref_test.py
```diff
@@ -1,145 +1,358 @@
...
+
+import pytest
+
 from pyteal import *
 
 from .compile_asserts import assert_new_v_old, compile_and_save
 
-# Set the following True, if don't want to run pass-by-ref dependent tests
-OLD_CODE_ONLY = False
+# TODO: remove these skips when the following is fixed: https://github.com/algorand/pyteal/issues/199
```
tzaffi, line 9: Changes in the file mostly consist of the following:
* introduce the `def test_pass_by_ref_guardrails()` test to ensure that the compiler errors on recursive functions that contain `ScratchVar` arguments
* move most of the previously defined recursive but pass-by-ref cases into `test_pass_by_ref_guardrails()`
* **SKIP** other tests in this file with the caveat that once #199 is addressed, they should be turned again. (currently they are flaky)

tests/teal/empty_scratches_expected.teal
```diff
@@ -0,0 +1,33 @@
+#pragma version 6
```
tzaffi, line 1: for new test case

tests/teal/lots_o_vars_expected.teal
```diff
@@ -0,0 +1,189 @@
+#pragma version 6
```
tzaffi, line 1: for new test case

pyteal/compiler/subroutines.py
```diff
@@ -90,6 +117,23 @@ def spillLocalSlotsDuringRecursion(
     """
     recursivePoints = findRecursionPoints(subroutineGraph)
```
michaeldiamant, line 119: @tzaffi Unrelated to the line:  Have you reviewed https://pyteal.readthedocs.io/en/stable/control_structures.html#recursion to determine if user guide changes are warranted?

I still don't have a great feel for how extensively we ought to cover topics in the user guide.  If you've reviewed and feel no change is warranted, I can be on-board.  Minimally, I'm gauging that user guide updates are given consideration.

pyteal/compiler/subroutines.py
```diff
@@ -56,12 +57,38 @@ def findRecursionPoints(
...
 
 
+def find_recursive_path(
+    subroutine_graph: Dict[SubroutineDefinition, Set[SubroutineDefinition]],
+    subroutine: SubroutineDefinition,
+) -> List[SubroutineDefinition]:
+    visited = set()
+    loop = []
+
+    def dfs(x):
```
jasonpaulos, line 73: Why create a new method of searching instead of using `graph_search`?

tests/pass_by_ref_test.py
```diff
@@ -1,145 +1,358 @@
...
-            Assert(player_score.index() == Int(129)),
-            Int(100),
-        )
+@pytest.mark.skipif(not STABLE_SLOT_GENERATION, reason="cf. #199")
+@pytest.mark.parametrize("pt", CASES)
+def test_teal_output_is_unchanged(pt):
+    assert_new_v_old(pt, 6)
+
+
+def test_pass_by_ref_guardrails():
```
jasonpaulos, line 226: Can you consider breaking this test up into multiple functions? There appears to be multiple distinct tests in here, but it's currently hard to distinguish between them since they are all interlaced.

jasonpaulos (commented):
Could you also add a simple unit test to `pyteal/compiler/subroutines_test.py` that tests the new functionality of `spillLocalSlotsDuringRecursion`?

pyteal/compiler/subroutines.py
tzaffi: Good call out. I commit to modifying the docs in this PR. Will leave this unresolved as a "visual reminder"

pyteal/compiler/subroutines.py
tzaffi: I needed to get the entire path from beginning of the recursion back to itself. I found I had to modify `graph_search()` too much for my comfort level - i.e. the risk of changing its behavior was too high for the amount gained of code re-usage. Hence, the new function.

tests/pass_by_ref_test.py
tzaffi: 👍 - will add cases/tests to other unit test files.

pyteal/compiler/subroutines.py
tzaffi: `graph_search` doesn't return the entire path, only the parent-child relationships for subroutines that are recursive. I guess an alternative to constructing the entire path from scratch would be to use the output of  `graph_search` and then search again on it to find the path. However, that didn't seem to be any simpler so I didn't pursue it. Do you have an alternate approach in mind?

ahangsu: well, as far as I understand it, Zeph is using `find_recursive_path` function to print a full path of self calling subroutine call-path, which fits into the definition of recursive call.

An alternative solution which fits Jason's preference is to extend `graph_search` to return a full path between start and end. Ignoring DFS/BFS, both functions should be able to find a bath between 2 nodes in finite sized connected graph.

jasonpaulos: My mistake, I didn't see that `dfs` modifies the `loop` variable. I'll resolve

jasonpaulos: Well I guess my suggestion would be that you shouldn't modify a captured variable like this because it's easy to miss; instead, returning a `Union[List, None]` or similar would be easier to understand at a glance. But this is purely a suggestion, and I don't mind if you don't take it

tzaffi: I see. I think I'll pass up this improvement since I'm trying to wrap a few other things up and it's not a blocker. Am resolving.

jasonpaulos:
> Could you also add a simple unit test to `pyteal/compiler/subroutines_test.py` that tests the new functionality of `spillLocalSlotsDuringRecursion`?

I don't see a new unit test here yet

pyteal/compiler/subroutines.py
tzaffi: I believe this is resolvable, assuming we're good with the following docs update:
image

michaeldiamant (approved):
@tzaffi Thanks for the follow up here to enforce constraints.

ahangsu (approved):
LGTM

PR #262, opened 2022-03-28
Upgrade black formatter

ahangsu:
The recent builds (e.g., https://github.com/algorand/pyteal/runs/5728134514?check_suite_focus=true and https://github.com/algorand/pyteal/runs/5727657498?check_suite_focus=true) failed for `black` using a previous click version, which is described https://github.com/psf/black/issues/2969.

Solution is straightforward, upgrading `black` version to latest one.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-black==21.7b0
+black==22.3.0
 mypy==0.931
 pytest
 pytest-timeout


michaeldiamant (approved):
@ahangsu Thanks for the fix ☕ 

Optional - I'm unsure of our stance for locking dependency versions.  I don't feel strongly, but I'd lean towards locking the version or allowing bug fix changes only.

ahangsu:
Yea, Igor and I had a short discussion on that, and we thought it won't hurt to allow higher version of `black` formatter in config. But if we think it is better locking specific version, that's also perfectly fine for me.

PR #264, opened 2022-04-01
Add ABI support

michaeldiamant:
Aggregates child PRs providing initial ABI support (https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0004.md).


tests/integration/ecdsa_test.py
```diff
@@ -164,11 +157,10 @@ def recover():
             )
         )
 
-    approval_app = blackbox_pyteal(recover, Mode.Application)
-    app_teal = compileTeal(approval_app(), Mode.Application, version=5)
     args = []
-    algod = algod_with_assertion()
-    app_result = DryRunExecutor.dryrun_app(algod, app_teal, args)
+    app_result = PyTealDryRunExecutor(recover, Mode.Application).dryrun(
+        args, compiler_version=5
```
tzaffi, line 162: Not sure why `compiler_version=5` here

ahangsu: I read from the current state of code, we are directly taking in `recover` this `Blackbox` decorated subroutine, rather than compiling from Pyteal to Teal. Then we do need `compiler_version` knowledge in dryrun, I think,.

pyteal/ast/abi/bool.py
```diff
@@ -88,11 +87,8 @@ def set(self, value: Union[bool, Expr, "Bool", ComputedValue["Bool"]]) -> Expr:
         if checked:
             return self.stored_value.store(value)
 
-        return Seq(
-            self.stored_value.store(value),
-            # instead of failing if too high of a value is given, it's probably more consistent with the rest of the AVM to convert values >= 2 to 1 (the && and || opcodes do this)
-            Assert(self.stored_value.load() < Int(2)),
-        )
+        # Not(Not(value)) coerces all values greater than 0 to 1
+        return self.stored_value.store(Not(Not(value)))
```
tzaffi, line 91: This is a clear improvement but still a deliberate choice which costs 2 opcodes vs. naively accepting the user's input. I can imagine a future "unsafe optimize" compilation option that stores directly.

jasonpaulos: Yes, there is still room for improvement here. Notably, if an expression is converted to an `abi.Bool` that we can know at compile time to be either 1 or 0, this check is certainly unnecessary. Examples are the output of `==`, `!=`, `!`, etc. I view this as a subclass of the general int size problem, since for all int sizes we perform safety checks of this nature.

But to your point, this improvement is a clear improvement over the old code from a purely performance perspective, since two negations are still cheaper than the old `load X; int 2; <; assert`.

tzaffi: Also, I probly should replace the TEAL in the graviton ABI-roundtrip tests with the mod's here. When are we aiming for release?

tzaffi: @michaeldiamant ^^^

ahangsu: If I remember correctly, are we aiming to release in Monday next week, namely, merge `feature/abi` to master?

ahangsu: I am working on changing the roundtrip test, a PR will be out soon

michaeldiamant: > When are we aiming for release?

@tzaffi Targeting 7/25, next Monday.

tzaffi: @ahangsu @michaeldiamant - I think I'm being silly. This should be happening [here now](https://github.com/algorand/pyteal/runs/7455644470?check_suite_focus=true#step:8:28) and I don't care about stale tests in graviton. Hope you didn't work too much on this.

tzaffi: but if something about the router changed, then I ought to refresh the graviton tests

ahangsu: @tzaffi looks good on my local side, the generated code is not influenced, good news. I am about to remove my branch then.

tzaffi (approved):
LFGTM

ahangsu (approved):
finally...

PR #265, opened 2022-04-03
Ed25519Verify exec mode depending on TEAL version

hernandp:
Ed25519Verify is only available for signature mode, but for TEAL <4.   Any mode otherwise.


jasonpaulos (approved):
Thanks for contributing, this looks good

PR #267, opened 2022-04-05
Allow `ComputedType` to be passed to the `set` method on ABI types

### LINKED ISSUE #266: Allow `ComputedType` to be passed to the `set` method on ABI types like Uint
## Problem

At the moment the only way to set something like the return value of a subroutine or an array element as the value of an ABI type is to use the `store_into` method or the `use` method. Neither of these are obvious to someone without looking up exactly what is available to them.

## Solution

Allow the `set` method to accept a `ComputedType` and call the `store_into` method internally.

This allows for a more natural syntax

```py
#...

(v := abi.Uint64()).set(arr[i.load()])

# Or for a tuple

(acct := abi.String()).set(tpl[0])
(balance := abi.Uint64()).set(tpl[1])
```

ahangsu:
## Description

This PR allows the `set` method to accept a `ComputedType` and call the `store_into` method internally.

This allows for a more natural syntax

```python
#...

(v := abi.Uint64()).set(arr[i.load()])

# Or for a tuple

(acct := abi.String()).set(tpl[0])
(balance := abi.Uint64()).set(tpl[1])
```

closes #266


barnjamin (approved):

pyteal/ast/abi/uint_test.py
```diff
@@ -1,10 +1,10 @@
 from typing import List, Tuple, NamedTuple, Callable, Union, Optional
-import pytest
-
-from pyteal.ast.abi.uint import UintTypeSpec
-
+from .type import ContainerType
+from .uint import UintTypeSpec
```
jasonpaulos, line 3: nit: this should also already be imported due to the next line

jasonpaulos (commented):
Mostly good, I mainly have comments about the type annotations.

Also, do you think `ComputedType` should be renamed to `ComputedValue`? I think that's a little more accurate

pyteal/ast/abi/tuple.py
```diff
@@ -249,20 +258,29 @@ def decode(
...
-        length: Expr = None
+        length: Expr = None,
     ) -> Expr:
         extracted = substringForDecoding(
             encoded, startIndex=startIndex, endIndex=endIndex, length=length
         )
         return self.stored_value.store(extracted)
 
-    def set(self, *values: BaseType) -> Expr:
+    def set(self, *values: Union[BaseType, ComputedType]) -> Expr:
```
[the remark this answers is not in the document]
ahangsu: good point, just changed the declaration to somewhat like `seq.py`.

pyteal/ast/abi/type.py
```diff
@@ -173,3 +182,22 @@ def use(self, action: Callable[[T], Expr]) -> Expr:
 
 
 ComputedType.__module__ = "pyteal"
+
+
+class ContainerType(ComputedType):
```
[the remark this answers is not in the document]
ahangsu: good point. I'm moving them to `type_test.py` and replace the `DummyComputedType`.

ahangsu:
`ComputedValue` sounds good to me, I will change later.

pyteal/ast/abi/uint.py
```diff
@@ -219,8 +222,8 @@ def type_spec(self) -> UintTypeSpec:
     def get(self) -> Expr:
         return self.stored_value.load()
 
-    def set(self, value: Union[int, Expr, "Uint", ComputedType]) -> Expr:
-        if isinstance(value, ComputedType):
+    def set(self, value: Union[int, Expr, "Uint", ComputedValue[T]]) -> Expr:
```
jasonpaulos, line 225: I think you'd want `self: T` as well so you can ensure the `ComputedValue` matches the instance type

jasonpaulos (commented):
Looks good, just one comment

pyteal/ast/abi/uint.py
ahangsu: nice, do you think in `tuple.py` case, your suggestion applies?

jasonpaulos (approved):

PR #213, opened 2022-02-28
Demo flake8 usage

michaeldiamant:
Demonstrate linting with [flake8](https://flake8.pycqa.org/en/latest/) so that we can assess how much further (if at all) we'd like to pursue the topic.

Notes:
* _If_ we wish to go further, I'll open a ticket.  Regardless, I expect to close the PR after a round of discussion.
* I'm loosely aware [pylint](https://pylint.org/) is a more mature linting tool.  I opted for flake8 after a brief docs survey and some googling.  My perception is flake8 might be easier to setup though on some long-term horizon, pylint might be more comprehensive. 
* Without having heard feedback, my initial takeaway is:  It's worthwhile to integrate flake8 without `pep8-naming`.  Here's why:
  * I think it'll address some common PR nitpicks (e.g. unused imports).
  * Moves the ball towards enforcing consistent repo standards so that we don't have to think abuot it.
  * I think it's low cost to integrate with CI tooling.  And/or a script can be added to provide convenience methods for running some vs all tooling (e.g. mypy, pytest, black, etc) locally + CI.
  * Over time, I think we'll want to consider how `pep8-naming` or some other uniform naming convention approach can be implemented.  But, I believe it's too difficult (for a number of reasons) to implement _today_.

Demo results:
* There's 3 summary results to consider for `flake8 pyteal`:
  *  [flake8-01-default.log](https://github.com/algorand/pyteal/files/8155635/flake8-01-default.log) - Out-of-the-box behavior without customization.
  * [flake8-02-excludes-added.log](https://github.com/algorand/pyteal/files/8155636/flake8-02-excludes-added.log) - Applies the PR configuration _without_  `pep8-naming` installed.
  * [flake8-03-enforce-pep8-naming.log](https://github.com/algorand/pyteal/files/8155637/flake8-03-enforce-pep8-naming.log) - Adds `pep8-naming` to the PR's configuration.
* Here's the violation count by scenario:
```
~/dev/pyteal flake8_trial *1 ─────────────────────────────────────────────────────────────────  pyteal 09:31:48 AM
❯ wc -l /tmp/flake8-*-*
    9833 /tmp/flake8-01-default.log
      84 /tmp/flake8-02-excludes-added.log
    1111 /tmp/flake8-03-enforce-pep8-naming.log
   11028 total
```

PR #273, opened 2022-04-12
Flake & Make

tzaffi:
Adding `Makefile` and `flake8`

* Linting related changes:
  *`.flake8`: 
    * Based on my experience in `graviton` and a couple of extra ignores for formatting that was conflicting with `black`
    * replace `from pyteal import *` with `import pyteal as pt` in our tests, and for other files ignore the * import error on a per file basis (_this is the reason line changes number in the thousands which amounts to almost 25% of the entire codebase_)
    * Incorporate #277 
* Additional changes:
  * Removing `requirements.txt` (but keeping `docs/requirements.txt`)
  * `setup.py`: `extras_require={"development" ... ` replaces the former `requirements.txt`
  * `.github/workflows/build.yml` + `Makefile`: unify as much of the logic as possible. Needed regular `python` instead of `python-slim` in order to have `make`. `black` is applied to all python files, as before, and `flake8` is as well. `mypy` is applied to `scripts` in addition to `pyteal` (in upcoming #260 this will be expanded to `tests` as well).
  * `README.md` - applying [mardkownlint](https://github.com/DavidAnson/markdownlint) (without automation)


pyteal/compiler/optimizer.py
```diff
@@ -1,7 +1,7 @@
 from typing import Set
```
tzaffi, line 1: @algoidurovic - pulling up the `optimizer` module. Any objections?

Makefile
```diff
@@ -0,0 +1,48 @@
...
+flake8:
+	flake8 $(ALLPY)
+
+MYPY = pyteal scripts
+mypy:
+	mypy $(MYPY)
+
+lint: black flake8 mypy
+
+test-unit:
```
tzaffi, line 39: I have a PR that will be adding integration tests. So might as well distinguish starting now. But I don't have a strong opinion about this.

.flake8
```diff
@@ -0,0 +1,2 @@
+[flake8]
+ignore = E203, E221, E241, E266, W291, E302, F403, F405, E501, W503, E741
```
michaeldiamant, line 2: @tzaffi Minor:  Unless opposed, can you refactor to one entry per line?

In my experience, having a single entry per line reduces the chance of mistake during PR review as we update entries.

.flake8
```diff
@@ -0,0 +1,2 @@
+[flake8]
```
michaeldiamant, line 1: @tzaffi For https://www.flake8rules.com/rules/F405.html and https://www.flake8rules.com/rules/F403.html, I'm imagining we'd prefer to avoid the violation in source code.  Does it work for you to limit ignoring F403 + F405 to `*_test.py` and `examples/`?

From the previous attempt, there's an example ignoring `*_test.py`:  https://github.com/algorand/pyteal/pull/213/files#diff-6951dbb399883798a226c1fb496fdb4183b1ab48865e75eddecf6ceb6cf46442R11. 

Running locally, I see there's a F403 violation in source code.  I didn't look closely enough to determine how easily it can be avoided vs adding inline ignore.  I can take a deeper look if you'd like - let me know.

```
~/dev/pyteal make-and-flake *1 !1 ?6 ────────────────────────────────────────────────────────────────────────── ✘ 2|1 5s  pyteal 09:53:33 AM
❯ make flake8 | grep -v -E 'test|examples'
pyteal/__init__.py:1:1: F403 'from .ast import *' used; unable to detect undefined names
pyteal/__init__.py:3:1: F403 'from .ir import *' used; unable to detect undefined names
make: *** [flake8] Error 1
```

pyteal/ast/subroutine_test.py
```diff
@@ -23,10 +23,10 @@ def fn2Args(a1, a2):
     def fn10Args(a1, a2, a3, a4, a5, a6, a7, a8, a9, a10):
         return Return()
 
-    lam0Args = lambda: Return()
-    lam1Args = lambda a1: Return()
-    lam2Args = lambda a1, a2: Return()
-    lam10Args = lambda a1, a2, a3, a4, a5, a6, a7, a8, a9, a10: Return()
+    lam0Args = lambda: Return()  # noqa: E731
```
michaeldiamant, line 26: @tzaffi If you prefer, I'm up for ignoring E731 in tests rather than providing inline ignores?

It feels like the kind of check that we'll repeatedly run into during typical development cycles.  I don't feel strongly and am happy to resolve as is.

pyteal/compiler/flatten_test.py
```diff
@@ -466,7 +466,7 @@ def sub3Impl(a1, a2, a3):
         subroutine3: subroutine3Ops,
     }
 
-    expectedL1Label = LabelReference("main_l1")
+    # expectedL1Label = LabelReference("main_l1")
```
michaeldiamant, line 469: @tzaffi Minor:  Friendly pointer to commented out code. I'm wondering if it's a holdover that can either be deleted or uncommented?

I raise the question without having performed an investigation.

Makefile
```diff
@@ -0,0 +1,48 @@
...
+	pip install -r docs/requirements.txt
+	pip install doc2dash
+
+pip-wheel:
+	pip install wheel
+
+bdist-wheel:
+	python setup.py sdist bdist_wheel
+
+docs:
```
michaeldiamant, line 17: @tzaffi Is `make docs` intended to be repeatable?

Here's the output I see upon 2nd run:

```
~/dev/pyteal make-and-flake *1 !1 ?6 ───────────────────────────────────────────────────────────────────────────────────  pyteal 10:27:59 AM
❯ make docs
cd docs && \
	make html && \
Running Sphinx v4.5.0
loading pickled environment... done
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 0 source files that are out of date
updating environment: 0 added, 0 changed, 0 removed
looking for now-outdated files... none found
no targets are out of date.
build succeeded.

The HTML pages are in _build/html.
Destination path "./pyteal.docset" already exists.
make: *** [docs] Error 17
```

For local development, I've typically run `make html` and left the `doc2dash` to the build.  Without having research further, it seems like the command can be made repeatable by deleting `docs/.pyteal.docset`.

Though I don't know enough off-hand to comfortably suggest deleting the dir vs for example, decomposing `docs` into 2 commands.

I can investigate further.  I'm raising the question in case you've got a strong feeling off-hand.  Let me know if you'd like me to look deeper.

.flake8
tzaffi: 👍 

Looking more closely at the [flake8's config definition](https://flake8.pycqa.org/en/latest/user/configuration.html) I believe your suggestion will work and I'll try it

.flake8
tzaffi: I'll research this some more and update here.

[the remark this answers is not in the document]
tzaffi: I'll look more into the effects of un-ignoring E266

pyteal/ast/subroutine_test.py
tzaffi: I believe we should follow python's modern conventions and _not_ create lambda variables where functions work just as well

michaeldiamant: Got it - That also works for me.  I had briefly wondered _why_ the style of conflicting definition is needed.

For the purpose of the PR, do you prefer to keep as is or make changes to remove the need for any inline ignores?

I'll go with either approach.

pyteal/compiler/flatten_test.py
tzaffi: I'm in favor to remove, but let's get @jasonpaulos 's opinion too, as I believe he is the author in #99

Makefile
tzaffi: I don't have a strong opinion. @jasonpaulos - do you have any thoughts about this? Maybe this is an example of where we _shouldn't_ create a `make` command

.flake8
tzaffi: When I first joined I didn't like our profligate use of * imports in pyteal. But the more I use pyteal, the more this convention makes sense to me. We have so many constructs that are typically used in pyteal, that a standard pyteal program would have 10 or more objects being imported. 

An alternate approach in such situations (where a library with many constructs is used) is to abbreviate the imported module and then always refer to it. For example:

```python
import pandas as pd

...
df = pd.DataFrame(...)
```

we could follow a similar convention and start enforcing it here with usages such as below (notice, we are already implicitly recommending this approach with `abi`):

```python
import pyteal as pt

toSum = abi.DynamicArray(abi.Uint64TypeSpec())
toMultiply = abi.DynamicArray(abi.Uint64TypeSpec())
results = abi.Tuple(abi.Uint64TypeSpec(), abi.Uint64TypeSpec())
program = pt.Seq(
    toSum.decode(pt.Txn.application_args[0]),
    toMultiply.decode(pt.Txn.application_args[1]),
    sumAndMultiply(toSum, toMultiply).store_into(results)
)
```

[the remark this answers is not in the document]
tzaffi: Ended up removing E266 (and fixing the flagged instances) and making a single-file exception for 221

michaeldiamant: A few thoughts from my side:
* If folks are on-board, I am on-board to use the `pandas as pd` approach.  I think it'd help disambiguate PyTeal constructs that match Python standard library (e.g. `Array`).
* If folks prefer to keep as is, I'd advocate to add the ignore only for tests and examples.  I'm thinking in source code we can be more perspective.  Though I'll ultimately follow the group's recommendation here.

Makefile
michaeldiamant: @tzaffi If there's no strong opinion, my default stance is to make the command idempotent.  If doing so feels onerous, then I'm open to backing out the `make` command.

tzaffi: Not onerous. I'll just add a `rm -rf`...

tzaffi: Ok, am just adding `rm -rf`...

pyteal/ast/subroutine_test.py
tzaffi: I prefer to explicit ignore. That way I'm not messing with other people's code and also it can be a useful example of how we might want to ignore such situations in the future.

pyteal/compiler/flatten_test.py
tzaffi: Alternately, I could uncomment and put a `# noqa: E841` on it

Makefile
```diff
@@ -0,0 +1,49 @@
+pip:
+	pip install -e .
```
jasonpaulos, line 2: Why would this ever be used over `pip-development`? If you're installing the package locally, you likely want the development dependencies as well. And `pip install -e .[development]` also install non-development dependencies, so there's no advantage to doing this first.

Makefile
jasonpaulos: I think it's misleading to have this be `make docs`, since this does a very specific thing that no one should have to do manually: it builds the docs and packages it in an archival format that's uploaded as a build artifact during our CI runs.

So I'd suggest either renaming it `bundle-docs` or not putting it into the Makefile

Makefile
```diff
@@ -0,0 +1,49 @@
...
+	make html && \
+	doc2dash --name pyteal --index-page index.html --online-redirect-url https://pyteal.readthedocs.io/en/ _build/html && \
+	tar -czvf pyteal.docset.tar.gz pyteal.docset
+
+build:
+	python -m scripts.generate_init --check
+
+ALLPY = docs examples pyteal scripts tests *.py
+black:
+	black --check $(ALLPY)
```
jasonpaulos, line 29: Why `$(ALLPY)` and not `.`?

Makefile
```diff
@@ -0,0 +1,49 @@
...
+
+docs:
+	rm -rf docs/pyteal.docset
+	cd docs && \
+	make html && \
+	doc2dash --name pyteal --index-page index.html --online-redirect-url https://pyteal.readthedocs.io/en/ _build/html && \
+	tar -czvf pyteal.docset.tar.gz pyteal.docset
+
+build:
+	python -m scripts.generate_init --check
```
jasonpaulos, line 25: This isn't really a build, since it doesn't modify anything. Maybe have two steps, `generate` and `check-generated`?

Makefile
```diff
@@ -0,0 +1,49 @@
...
+	black --check $(ALLPY)
+
+flake8:
+	flake8 $(ALLPY)
+
+MYPY = pyteal scripts
+mypy:
+	mypy $(MYPY)
+
+lint: black flake8 mypy
```
jasonpaulos, line 38: I'd prefer if `lint` didn't modify any files and just errored if they had issues. Perhaps have `lint` and `lint-fix` commands?

Makefile
```diff
@@ -0,0 +1,49 @@
+pip:
+	pip install -e .
+
+pip-development: pip
```
jasonpaulos, line 4: nit: if you view `pip` as a verb/idoim for "install pip dependencies for", these commands make sense, but I'm not sure I fully agree with that. Perhaps `setup-development`, `setup-docs`, etc. would be more clear?

pyteal/ast/acct_test.py
```diff
@@ -1,5 +1,3 @@
-import pytest
-
 from .. import *
 
 # this is not necessary but mypy complains if it's not included
```
jasonpaulos, line 3: Since you're already touching a lot of files, would you mind looking into whether mypy still needs these superfluous imports? I suspect it doesn't, since these are relics from an older version of mypy, and that we'd be able to remove them.

pyteal/ast/methodsig.py
```diff
@@ -3,7 +3,6 @@
 
 from pyteal.types import TealType
 
-from ..types import TealType
```
jasonpaulos, line 6: I'm not sure what happened here, there's both `from pyteal.types` and `from ..types`. The former should be removed, not this one

pyteal/ast/multi_test.py
```diff
@@ -1,6 +1,8 @@
 from typing import List
 
-from .. import *
+from pyteal import *
+from pyteal.ast import *
+from pyteal.ir import *
```
jasonpaulos, line 5: Why did this change? The convention in this repo is to use relative imports, and there should be no reason to directly import the `ast` and `ir` modules

pyteal/compiler/optimizer.py
jasonpaulos: I think this should remain a submodule. In the future we will likely have more optimizations and there's no reason to clutter the compiler module with them, since their purpose is more specific

pyteal/compiler/optimizer_test.py
```diff
@@ -1,8 +1,7 @@
 from .optimizer import OptimizeOptions, _apply_slot_to_stack
 
-from ... import *
-
-import pytest
+from pyteal import *
+from pyteal.ast import *
```
jasonpaulos, line 4: nonrelative import, importing ast not necessary

pyteal/compiler/flatten_test.py
jasonpaulos: A bug has been found: `l1Label` should be replaced with `expectedL1Label` on lines 486 and 492. The linter is already useful!

pyteal/compiler/subroutines_test.py
```diff
@@ -2,7 +2,7 @@
 
 import pytest
 
-from .. import *
+from pyteal import *
```
jasonpaulos, line 5: nonrelative import

tests/module_test.py
```diff
@@ -1,6 +1,3 @@
-from pyteal import *
```
jasonpaulos, line 1: This should not be removed. The below test is to make sure this line does not shadow the `int` variable

.flake8
jasonpaulos: There's 3 separate issues:
* Use of star imports in our runtime code: this is highly useful because we have a few modules that import and export _all_ of a submodule. It would be too time consuming and error prone to not use a star import here. My opinion is that adding inline exceptions for these cases is good enough.
* Use of star imports in example: from the beginning we've encouraged this because PyTeal programs can use a lot of different AST elements, and because our syntax is already clunky enough without added prefixed in my opinion. I'm in favor of keeping it as-is since our examples should look natural and not be intimidating.
* Use of star imports in our tests: this matters least to me, so if you'd like to adopt `import pyteal as pt` or similar, I wouldn't mind.

Makefile
tzaffi: 👍

Makefile
tzaffi: 👍

pyteal/ast/acct_test.py
michaeldiamant: @jasonpaulos @tzaffi I attempted to address the feedback via https://github.com/algorand/pyteal/pull/277.  Can you have a look?

Makefile
tzaffi: going with the `bundle-docs` approach

Makefile
tzaffi: 👍

Makefile
tzaffi: It was going into my virutal env directory and failing because of that, so I wanted to explicitly define what is being checked. It's also handy to use this for the other linters. So I would prefer to keep this way. 

Another approach is to specifically exclude certain folders with `--exclude`, but then everyone would have to configure that specifically for their local excludes. 

But admittedly, the downside is that if you add another top-level directory, we wouldn't be formatting it by default. 

I don't feel strongly about this. LMK

Makefile
tzaffi: 👍

.flake8
tzaffi: I'll go with the modifying the tests files only and adding flake8 exceptions for other files.

tests/module_test.py
tzaffi: 👍

pyteal/compiler/subroutines_test.py
tzaffi: 👍

pyteal/compiler/optimizer_test.py
tzaffi: 👍

pyteal/ast/methodsig.py
tzaffi: 👍

pyteal/ast/multi_test.py
tzaffi: 👍

pyteal/compiler/flatten_test.py
tzaffi: Cool. I fixes as suggested.

pyteal/compiler/optimizer.py
tzaffi: 👍

pyteal/ast/acct_test.py
tzaffi: I've merged this in. Thank you!

.flake8
tzaffi: hehe... that resulted in 6000+ changes

michaeldiamant: @tzaffi Thanks for the changes here - I think it's reasonable starting point and we can iterate over time.  Resolving.

michaeldiamant (approved):
@tzaffi Thanks for the effort to make the repo more robust.  I think it's a good starting point and and we can customize config as we get experience.

README.md
```diff
@@ -8,59 +9,66 @@
...
- * `. venv/bin/activate.fish` (if your shell is fish)
 
-Pip install PyTeal in editable state
- * `pip install -e .`
+* `. venv/bin/activate` (if your shell is bash/zsh)
+* `. venv/bin/activate.fish` (if your shell is fish)
+
+Pip install PyTeal in editable state with dependencies:
+
+* `pip install -e.[development]`
```
ahangsu, line 58: nit: should be `pip install -e . [development]`? I am not quite sure what `[development]` refers to.

ahangsu (commented):
generally looking good, just one question

README.md
tzaffi: This PR removes `requirements.txt` which had several dependencies such as `black`, `mypy` and `pytest`. These have been moved into `setup.py` under `extras_require=`. Additionally, in another thread @jasonpaulos pointed out that it's unlikely anyone would want to `pip install` without the development requirements. So I removed that example and replaced with this. I'm definitely not wed to this position and am open to suggested improvements.

ahangsu: I see what's going on my side, I am on `zsh`, and I have to use do `pip install -e .\[development\]`, nothing to worry about!

tzaffi: Oh good catch. I'm using `zsh` too, but I've always just run `make setup-development` which works just fine. But you're right, when I tried the raw `pip` command I got the same error. I'll add a min-section to highlight this problem.

ahangsu: A quick solution: https://kinopyo.com/en/blog/escape-square-bracket-by-default-in-zsh

ahangsu (approved):

pyteal/compiler/optimizer/__init__.py
```diff
@@ -1 +0,0 @@
-from .optimizer import OptimizeOptions, apply_global_optimizations
```
jasonpaulos, line 1: Please restore the optimizer package to how it was before so that our module rules aren't broken. https://github.com/algorand/pyteal/pull/247#discussion_r833510628

pyteal/ast/binaryexpr_test.py
```diff
@@ -1,1499 +1,1560 @@
 import pytest
 
-from .. import *
+import pyteal as pt
```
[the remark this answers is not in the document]
tzaffi: Yes, I haven't been able to figure out how to take the top level relative import `.` into an alias `pt`. The [example I gave in the previous discussion](https://github.com/algorand/pyteal/pull/273#discussion_r849748369) also imports this way (`from pyteal import pt`). I probably should have been more explicit in stating that this was my plan. 

There is one advantage to this approach. We are asserting a non-relative import path in the unit test, and often explicit is preferred to implicit in unit tests. Additionally, this eases the way to moving all the unit tests into `tests/unit` if that is something we choose to do in the future.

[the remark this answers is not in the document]
jasonpaulos: I see the issue, `import .. as pt` does not work since it requires an absolute module path, and I don't see a way around that.

I'll think about the ramifications of this, but it does not block this PR in any way.

jasonpaulos (approved):

PR #274, opened 2022-04-12
Utility for increasing budget during execution of PyTeal applications

algoidurovic:
This PR implements a PyTeal utility for increasing the opcode budget during execution of an app call using inner app call transactions.

Two modes are provided:
1. Explicit: the app to call for increasing budget is provided by the user.
2. OnCall: the app to call is created during each request to increase the budget, then deleted afterwards.

Two budget increase methods are provided:
1. ensure_budget(budget): increase the current opcode budget to at least the budget requested by the user.
2. maximize_budget(fee): increase the current opcode budget as much as possible without spending more than the given fee. This can be used when the contract developer doesn't want to spend more than the fee provided to increase the budget without needing to do in depth opcode cost analysis on their program. If the fee isn't sufficient for executing the program then a failure may be desired. It can also allow the app caller to specify the maximum fee they want to spend for increasing opcode budget. This behavior can fairly easily be extended to support a global fee limit for budget increase requests. (Inspired by the algofi implementation of opup).

Testing
Unit tests for type checking enforcement and sandbox tests for e2e functionality. The PyTeal programs I used for testing in sandbox are located in examples/application/opup.py


pyteal/ast/opup.py
```diff
@@ -0,0 +1,141 @@
...
+                        TxnField.application_id: self.target_app_id,
+                    }
+                ),
+                InnerTxnBuilder.Submit(),
+            )
+        )
+
+    def execute(self, required_budget: Expr) -> Expr:
+        if self.mode == OpUpMode.OnCall:
+            return If(Global.opcode_budget() < required_budget).Then(
```
barnjamin, line 134: Since we do this check I think I'd favor a name like `ensure`

pyteal/ast/opup.py
```diff
@@ -0,0 +1,141 @@
...
+    Explicit = 0
+
+    """The app to call is created/deleted when the main app is created/deleted."""
+    OnCreate = 1
+
+    """The app to call is created then deleted for each request to increase budget."""
+    OnCall = 2
+
+
+NoOpApp = Bytes("byte 0x068101") # teal program "int 1" assembled
```
jasonpaulos, line 33: If this is used as the app approval program for `OnCreate`, then _anyone_ could update or delete this app, since it does not check who the sender is.

Secondary concern, why is `byte` part of the program string?

pyteal/ast/opup.py
```diff
@@ -0,0 +1,141 @@
...
+            )
+        )
+
+    def execute(self, required_budget: Expr) -> Expr:
+        if self.mode == OpUpMode.OnCall:
+            return If(Global.opcode_budget() < required_budget).Then(
+                Seq(
+                    self._create_app(),
+                    self._opup_expr(required_budget),
+                    self._delete_app(),
```
jasonpaulos, line 138: Are you aware that you can create and delete an app in a single transaction? During the app create call, if you set `TxnField.on_completion` to `OnComplete.DeleteApplication` and it will delete the app immediately after it's created.

pyteal/ast/opup.py
jasonpaulos, line 134: Have you considered what happens around boundary conditions? There are two cases I can think of:
* The program already has extremely low budget (e.g. 1 unit left) before entering this function. Likely it will fail because you can't even check if you need to op up successfully with such a small budget, so you should put in the docs what the minimum op budget should be before calling the function.
* The program has only slightly more than the amount it asked for. There _might_ be a situation where this function thinks it has enough budget, but the act of returning from the function causes the budget to drop a few units below the amount the caller asked for (due to branch statements that still need to get executed). I've not verified this is the case, but if you know X additional budget is needed to return from the function, you could check `If(Global.opcode_budget() < required_budget + Int(X))`.

jasonpaulos: Another way to phrase the above questions is: how much budget is used before the first inner txn to op up, and how much budget is used after the last `Global.opcode_budget()` check?

pyteal/ast/opup.py
algoidurovic: This was more of a placeholder than anything so haven't put much thought into it. Point taken though.

pyteal/ast/opup.py
algoidurovic: Totally agree, `ensure` or `ensure_budget` would be more clear.

pyteal/ast/opup.py
algoidurovic: I'm trying to minimize the number of app creations by limiting `execute` to create only one app per invocation, even if reaching the given budget requires multiple inner app calls.

pyteal/ast/opup.py
algoidurovic: Good point. A call to `execute` should guarantee that the opcode budget is at least the amount passed by the user **after** the call is completed.

algoidurovic: Fixed. Resolving.

pyteal/ast/opup.py
algoidurovic: fixed. resolving.

pyteal/ast/opup.py
algoidurovic: is this okay to resolve?

pyteal/ast/opup.py
algoidurovic: fixed. resolving.

tzaffi:
> Two budget increase methods are provided:
> 
> 1. ensure_budget(budget): increase the current opcode budget to at least the budget requested by the user.
> 2. maximize_budget(fee): increase the current opcode budget as much as possible without spending more than the given fee.

I think method 1 is essential, but I'm not as sure about method 2. On the face of it, `maximize_budget()` gives you an unknown amount of extra budget, so you could run out unexpectedly or get much more than you really need. One could argue that method 1 suffers from the opposite problem - what if fees increase and the app can no longer afford to op-up? But that seems a little easier to ameliorate dynamically, if another method is added `fee_for_budget(budget)` which would let the user know how much they'll use up and adapt accordingly.

pyteal/ast/opup.py
```diff
@@ -0,0 +1,163 @@
...
+        """Ensure that the budget will be at least the required_budget."""
+        require_type(required_budget, TealType.uint64)
+
+        buffered_budget = required_budget + self.buffer
+        if self.mode == OpUpMode.OnCall:
+            return If(Global.opcode_budget() < buffered_budget).Then(
+                Seq(
+                    self._create_app(),
+                    self._ensure_budget_expr(buffered_budget),
+                    self._delete_app(self.target_app_id),
```
michaeldiamant, line 127: @algoidurovic Gauging my understanding - Is the following statement accurate?

If `_ensure_budget_expr` fails, then the created app lives indefinitely.

_If_ my understanding is accurate, I'm wondering if it's possible to avoid leaking the app by creating an inner transaction group via `InnerTxnBuilder` containing (1) create app, (2) op up, and (3) delete app?

pyteal/ast/opup.py
```diff
@@ -0,0 +1,163 @@
...
+                )
+            )
+
+        return self._ensure_budget_expr(buffered_budget)
```
michaeldiamant, line 131: @algoidurovic For my understanding:  With `OpUpMode.Explicit`, what operations require a buffered budget?

Here's what I interpreted - let me know if/where I'm missing intuition:
*  `OpUpMode.Explicit` does _not_ require a buffered budget because control returns to the application _immediately_ after increasing opcode budget.
* `OpUpMode.OnCall` _does_ require a buffered budget because a delete app operation follows the opcode budget increase.
  * Aside:  How is `self.buffer = Int(50)` determined?
    * If I'm following, it seems the buffer can be specialized for the delete operation.

pyteal/ast/opup.py
```diff
@@ -0,0 +1,163 @@
...
+
+    def _maximize_budget_expr(self, fee: Expr) -> Expr:
+        i = ScratchVar(TealType.uint64)
+        n = fee / Int(1000)
+        return For(i.store(Int(0)), i.load() < n, i.store(i.load() + Int(1))).Do(
```
michaeldiamant, line 136: @algoidurovic Is it accurate that `maximize_budget` presumes a 1 micro Algo transaction fee?

* I have limited experience here though I know from https://developer.algorand.org/docs/get-details/transactions/#fees that transaction fees are variable.
* I did a brief scan and I don't see a way to gauge the transaction fee without attempting the transaction.
* If I'm following along, then I'm wondering if there's a specific use case that prompted the _maximize_budget_ approach or if it's been added for convenience?
  * I see the intended convenience.  _If_ it's intended as a convenience and _if_ there's there's no reasonable way to account for the dynamic transaction fees, I'm thinking the functionality can be added in a later iteration.

pyteal/ast/opup.py
```diff
@@ -0,0 +1,163 @@
...
+                    {
+                        TxnField.type_enum: TxnType.ApplicationCall,
+                        TxnField.application_id: self.target_app_id,
+                    }
+                ),
+                InnerTxnBuilder.Submit(),
+            )
+        )
+
+    def ensure_budget(self, required_budget: Expr) -> Expr:
```
michaeldiamant, line 117: @algoidurovic and reviewers:  How essential is it to consider apps invoking `ensure_budget` > 1 per invocation?

I raise the use case because I don't know if it's been considered.  And from what I can imagine, the use case _may_ bring substantive PR changes.

* Concerns with multiple invocations:
  * It _may_ be worthwhile to define op up functionality as `@Subroutine`s to minimize program size.  In the PR's current form, the `self` reference prevents `@Subroutine` usage.  Though it's _possible_ to imagine a `@Subroutine` compliant revision. 
  * I can _imagine_ it'd be preferable for `OpUpMode.OnCall` to return the created app ID.  Retaining the app ID reduces opcode cost + transaction fees for subsequent op up calls.  I can imagine a few ways to support the use case though they'll all involve changing the API.
* Provided the use case I'm raising isn't deemed _essential_, I prefer to resolve without changes and retain a simpler approach for now.

pyteal/ast/opup.py
```diff
@@ -0,0 +1,163 @@
...
+
+        # A budget buffer is necessary to deal with an edge case of ensure_budget():
+        #   if the current budget is equal to or only slightly higher than the
+        #   required budget then it's possible for ensure_budget() to return with a
+        #   current budget less than the required budget. The buffer prevents this
+        #   from being the case.
+        self.buffer = Int(50)
+
+    def _create_app(self) -> Expr:
+        self.target_app_id = self.target_app_id_slot.load()
```
michaeldiamant, line 76: @algoidurovic I'm not sure I follow how the statement is having the intended effect.  Can you help me understand what's happening here?

I _think_ the intent is to updated `target_app_id` for later usage in delete.  However, at the highlighted line, I don't see how the created app ID can be known.  It seems app creation happens _next_.

I'd need to inspect TEAL to be more confident, but I also _believe_ the `load` invocation needs to be part of the returned AST.

pyteal/ast/opup.py
michaeldiamant: Ah, I realize I missed `self.target_app_id_slot.store(Gitxn[0].created_application_id()),`.  

It seems an attempt is made to cache the created app ID.  However, unless I missed it, I don't see an attempt to read the cached ID.

I'm open to a caching approach.  Though I feel the mode name (or at least docs) should reflect the caching behavior because the semantics are non-obvious and may not match the caller's intent.

algoidurovic:
> > Two budget increase methods are provided:
> > 
> > 1. ensure_budget(budget): increase the current opcode budget to at least the budget requested by the user.
> > 2. maximize_budget(fee): increase the current opcode budget as much as possible without spending more than the given fee.
> 
> I think method 1 is essential, but I'm not as sure about method 2. On the face of it, `maximize_budget()` gives you an unknown amount of extra budget, so you could run out unexpectedly or get much more than you really need. One could argue that method 1 suffers from the opposite problem - what if fees increase and the app can no longer afford to op-up? But that seems a little easier to ameliorate dynamically, if another method is added `fee_for_budget(budget)` which would let the user know how much they'll use up and adapt accordingly.

I updated the PR description to further elaborate on the use cases for maximize_budget(). I do think it's useful to have a fee_for_budget() method available as well.

pyteal/ast/opup.py
michaeldiamant: Had a verbal with @algoidurovic - I misunderstood inner transaction semantics.  Any failing inner transaction (regardless of grouping) fails the app invocation and all associated transactions. 

So - The PR does _not_ dangle app IDs.  Sorry for the confusion.  Resolving.

pyteal/ast/opup.py
michaeldiamant: Summarizing a verbal with @algoidurovic:
* I missed the following in my original note:  Since the budget is checked in a while loop, there's _always_ a check after op-ing up to see if the loop completed.  Consequently, there's _always_ a budget needed. 
* Igor to investigate:  Discussion highlighted that it'd probably make sense to invert the comparison ordering.  Concretely, switch from `Global.opcode_budget() < buffered_budget` to `buffered_budget >= Global.opcode_budget()`.  The change ensures `opcode_budget()` is interpreted as late as possible and provides the most up-to-date value.
* Igor to consider:  We discussed that the 50 opcode buffer is intended to be a rough, best fit for many use cases.  Since it's likely too large in some use cases, we discussed optionally offering a configurable override (e.g. factory method) for the app developer to supply a specific value.

pyteal/ast/opup.py
michaeldiamant: Summarizing a verbal with @algoidurovic:
* There's 2 levels of configurable fees:
  * Congestion fees
  * Min transaction fees configured via consensus parameter
* Igor to investigate:
  * Replace `Int(1000)` with a min transaction fee field to catch consensus parameter changes.
  * Identify if congestion fees apply to inner transactions.  If yes, we ought to highlight that the method _can_ fail at evaluation time by either (or both) changing method name and clarifying docs.

pyteal/ast/opup.py
michaeldiamant: Summarizing a verbal with @algoidurovic:

* I misinterpreted the AST evaluation.  I mistakenly treated `self.target_app_id` as a local variable.
* We discussed that it's worthwhile to consider different representations that might be less error prone to follow.  Leaving the thread open while the topic is considered though we're good to keep as is.

pyteal/ast/opup.py
michaeldiamant: Summarizing a verbal with @algoidurovic:
* Igor to consider:  Reworking to support defining as `@Subroutine`.  If that doesn't seem like a good fit, Igor suggested we can point out to the app developer how to wrap invocations in docs.
* We agreed returning the created app ID is out of scope.  It'd be less error prone for the app developer to create the app ID out-of-band and use `Explicit` mode.

pyteal/ast/opup.py
michaeldiamant: Thanks to @algoidurovic for pointing out that inner transactions are _not_ subject to congestion fees because they're _not_ transmitted over the network.

https://github.com/algorand/pyteal/pull/274/commits/7152af649c6ab0d5a1a6690fcaee95268240b3ea brings in the feedback to factor out the min txn fee constant.    

Resolving.

pyteal/ast/opup.py
```diff
@@ -0,0 +1,164 @@
...
+                    TxnField.type_enum: TxnType.ApplicationCall,
+                    TxnField.on_completion: OnComplete.DeleteApplication,
+                    TxnField.application_id: app_id,
+                }
+            ),
+            InnerTxnBuilder.Submit(),
+        )
+
+    def _ensure_budget_expr(self, required_budget: Expr) -> Expr:
+        return While(required_budget < Global.opcode_budget()).Do(
```
michaeldiamant, line 105: @algoidurovic Minor:  Is the intend inequality `<` or `<=`?

I raise the question because I thought it'd be `<=`.  When `=`, I thought there's no need to acquire more opcode budget because the compared budget reflects the budget.

If there's a change to be made, note that there's 2 places to make the change.

pyteal/ast/opup.py
michaeldiamant: @algoidurovic Thanks for bringing in some feedback via https://github.com/algorand/pyteal/pull/274/commits/7152af649c6ab0d5a1a6690fcaee95268240b3ea.  

The remaining open question I see is:

> Igor to consider: We discussed that the 50 opcode buffer is intended to be a rough, best fit for many use cases. Since it's likely too large in some use cases, we discussed optionally offering a configurable override (e.g. factory method) for the app developer to supply a specific value.

pyteal/ast/opup.py
algoidurovic: > When =, I thought there's no need to acquire more opcode budget because the compared budget reflects the budget.

If I'm understanding you correctly, the existing comparison does enforce this. `<=` would continue the loop when the values being compared are equal.

michaeldiamant: @algoidurovic Duh - Sorry for the noise here.  Disregard - resolving.

pyteal/ast/opup.py
```diff
@@ -0,0 +1,164 @@
...
+                    {
+                        TxnField.type_enum: TxnType.ApplicationCall,
+                        TxnField.application_id: self.target_app_id,
+                    }
+                ),
+                InnerTxnBuilder.Submit(),
+            )
+        )
+
+    def ensure_budget(self, required_budget: Expr) -> Expr:
```
jasonpaulos, line 118: I still don't see any docs here saying what the minimum budget should be before calling this function.

pyteal/ast/opup.py
algoidurovic: I think we should have a team discussion on coding practices in PyTeal relating to this topic. There's definitely an argument for representing PyTeal "subprograms" as functions rather than variables but this may complicate the code in certain places so we likely shouldn't apply it universally even if we do adopt this practice.

for example:
```
expr = Int(1) + Int(2), program = Seq(..., expr, ...)
->
def expr():
  return Int(1) + Int(2)
program = Seq(..., expr(), ...)
```

The benefit to the second approach is that it's more clear to the developer that `expr` stores an unevaluated expression. With that said, my opinion is that this is somewhat outside the scope of this PR and shouldn't hold up approval. Let me know what you think.

pyteal/ast/opup.py
algoidurovic: To your first point, my position is that wrapping is a better option because the overhead is low (in terms of both implementation time/complexity and efficiency) but doesn't prevent the user from using the standard approach, which has less slot usage and fewer arguments. I'll add to the docs as soon as the code is no longer in flux. Open to feedback here.

pyteal/ast/opup.py
algoidurovic: I added the option for the user to pass a custom buffer value to the OpUp constructor and tightened the default buffer value after some experimentation. Let me know if this addresses your concerns.

pyteal/ast/opup.py
algoidurovic: Should be taken care of now.

pyteal/ast/opup.py
```diff
@@ -0,0 +1,164 @@
...
+        """Ensure that the budget will be at least the required_budget."""
+        require_type(required_budget, TealType.uint64)
+
+        buffered_budget = required_budget + self.buffer
+        if self.mode == OpUpMode.OnCall:
+            return If(buffered_budget < Global.opcode_budget()).Then(
+                Seq(
+                    self._create_app(),
+                    self._ensure_budget_expr(buffered_budget),
+                    self._delete_app(self.target_app_id),
```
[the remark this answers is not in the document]
algoidurovic: The only reason for my original approach was to go easy on indexer but with that concern addressed I agree this approach is better for multiple reasons, one being that it allows for much cleaner code. Made the changes and tested in sandbox.

pyteal/ast/opup.py
```diff
@@ -0,0 +1,162 @@
...
+    """
+
+    """The app to call must be provided by the user."""
+    Explicit = 0
+
+    """The app to call is created then deleted for each request to increase budget."""
+    OnCall = 1
+
+
+ON_CALL_APP = Bytes("base16", "06810143")  # v6 pyteal program "Int(1)"
```
jasonpaulos, line 32:
```suggestion
ON_CALL_APP = Bytes("base16", "068101")  # v6 teal program "int 1"
```

Pyteal automatically adds a `return` op at the end of the program, but you can exclude it to save space and 1 extra compute unit (though it's kinda strange that the `return` op consumes budget in the first place 🤔 )

jasonpaulos (commented):
Except for my one minor comment, this looks good to me, thank you for incorporating my feedback.

The one comment is about shrinking the on call program code. I edited the TEAL by hand, and even though it's a simple change I'd still recommend you briefly test it in sandbox again just to make sure there are no unintended consequences.

pyteal/ast/opup.py
michaeldiamant: @algoidurovic A few thoughts here below.  I'll resolve because I think we're good though feel welcomed to re-open if you'd like to continue discussion.

A few thoughts:
* Thanks for giving the topic consideration.
* At least for the convention in-question, I don't feel strongly that a single convention _must_ always be followed.  I can imagine different styles making sense in different places.  Which is also to say - I think it's reasonable to try something in 1 place without applying it to _all_ places.
* I agree we need not hold up the PR.  As of https://github.com/algorand/pyteal/pull/274/commits/399a1b915f77dae95eaee41f5837e31c6d5d9ee3, the PR is in a different form and I think the concern is addressed.
* I'm up for a group discussion.  Can make for a post-stand up discussion for those interested in the topic.

pyteal/ast/opup.py
michaeldiamant: @algoidurovic Got it - I haven't considered the topic further.  Let's see how the docs look.  Leaving the thread open as a visual reminder.

pyteal/ast/opup.py
michaeldiamant: @algoidurovic Thanks for bringing in the feedback.  

Before resolving, I have an _optional_ thought.  If you prefer as is, feel welcome to tell me and resolve the thread.

* Idea:  Since `budget` is _only_ referenced by `ensure_budget`, I'd optionally consider moving `buffer` from constructor parameter to method parameter.
* Rationale:  
  * Explicitly demarcates where `buffer` applies.  When provided to the constructor, the user must be aware of the implementation to see usage (e.g. Does `maximize_budget` use `buffer`?). 
  * Additionally, a user may wish to qualify `buffer` on a per invocation basis.  With a constructor parameter, the user must construct separate objects to express the use case.

docs/opup.rst
```diff
@@ -0,0 +1,85 @@
...
+================= ================================================================================
+:code:`Explicit`  Calls a user provided external app when more budget is requested.
+:code:`OnCall`    Creates the app to be called when more budget is requested.
+================= ================================================================================
+
+
+:any:`Explicit` has the benefit of constructing more lightweight inner transactions but requires the
+target app ID to be provided in the foreign apps array field of the transaction and the :any:`pyteal.OpUp`
+constructor in order for it to be accessible. :any:`OnCall` is easier to use but has slightly more overhead
+because the target app must be created and deleted during the evaluation of an app call.
```
michaeldiamant, line 31: @algoidurovic Minor:  _Suggests_ adding commas for readability.

```suggestion
:any:`Explicit` has the benefit of constructing more lightweight inner transactions, but requires the
target app ID to be provided in the foreign apps array field of the transaction and the :any:`pyteal.OpUp`
constructor in order for it to be accessible. :any:`OnCall` is easier to use, but has slightly more overhead
because the target app must be created and deleted during the evaluation of an app call.
```

docs/opup.rst
```diff
@@ -0,0 +1,85 @@
...
+
+:any:`pyteal.OpUp.ensure_budget` attempts to ensure that the available budget is at least the budget requested by
+the caller. If there aren't enough funds for issuing the inner transactions or the inner transaction limit
+is exceeded, the evaluation will fail. The typical usage pattern is to insert the :any:`pyteal.OpUp.ensure_budget`
+call just before a particularly heavyweight subroutine or expression. Keep in mind that the required budget
+expression will be evaluated before the inner transactions are issued so it may be prudent to avoid expensive
+expressions, which may exhaust the budget before it can be increased.
+
+In the example below, the Ed25519Verify expression is used, which costs 1900.
```
michaeldiamant, line 40: @algoidurovic Minor thoughts:
* It's possible to link the `Ed25519Verify` reference.   https://stackoverflow.com/questions/22700606/how-would-i-cross-reference-a-function-generated-by-autodoc-in-sphinx helped me understand how to do it.
* Optionally, I prefer to add commas for readability on numbers with > 3 digits.


```suggestion
In the example below, the :py:meth:`pyteal.Ed25519Verify` expression is used, which costs 1,900.
```

docs/opup.rst
```diff
@@ -0,0 +1,85 @@
...
+    )
+
+:any:`pyteal.OpUp.maximize_budget` attempts to issue as many inner transactions as possible with the given fee.
+This essentially maximizes the available budget while putting a ceiling on the amount of fee spent. Just
+as with :any:`pyteal.OpUp.ensure_budget`, the evaluation will fail if there aren't enough funds for issuing the
+inner transactions or the inner transaction limit is exceeded. This method may be preferred to
+:any:`pyteal.OpUp.ensure_budget` when the fee spent on increasing budget needs to be capped or if the developer
+would rather just maximize the available budget instead of doing in depth cost analysis on the program.
+
+In the example below, the fee is capped at 3000 for increasing the budget. This works out to 3 inner
```
michaeldiamant, line 65: @algoidurovic Optional _suggestions_:
* Add commas for numbers with > 3 digits.
* Specify units to avoid ambiguity - microAlgos in this case.

```suggestion
In the example below, the fee is capped at 3,000 microAlgos for increasing the budget. This works out to 3 inner
```

pyteal/ast/opup.py
michaeldiamant: Resolving - As of https://github.com/algorand/pyteal/pull/274/commits/07bd6d25338c3f6da89b38dd9bfcebaceb777e34, `buffer` is no longer needed as a user-configurable parameter.

pyteal/ast/opup.py
michaeldiamant: As of https://github.com/algorand/pyteal/pull/274/commits/07bd6d25338c3f6da89b38dd9bfcebaceb777e34, the docs briefly mention wrapping the logic to define as `@Subroutine`. 

I'm _mostly_ comfortable as is.  If program size becomes an issue due to repeated opup calls per app invocation, then we can revisit.  Resolving.

docs/opup.rst
```diff
@@ -0,0 +1,85 @@
+.. _opup:
+
+Budget Increase Utility
```
michaeldiamant, line 3: @algoidurovic Optionally, I propose mentioning _OpUp_ in the page title.  Since it seems we'll use the language _OpUp_, I think it'd be helpful for readers to see the term in the table of contents + at top of page.

```suggestion
OpUp:  Budget Increase Utility
```

examples/application/program.teal
```diff
@@ -0,0 +1,39 @@
+#pragma version 6
```
michaeldiamant, line 1: @algoidurovic Minor:  I'm unfamiliar with our `examples/application` conventions.  I don't see other `.teal` sources versioned.  

I'm wondering if 1 of the following ought to happen?
* Declare we do _not_ version `teal` sources in `examples/application` by adding a `.gitignore` entry.
* Declare we _want_ to version `teal` sources and either rename to a more specific name (e.g. `opup.teal`) or reorganize applications (e.g. dir per app).
* Delete `program.teal` and deal with the concern another time.

If a policy doesn't exist, I think it's useful to eventually declare a policy.  Though I don't want to hold the PR.

I can be convinced to leave the file as is rather than deleting or renaming it.  Though leaving as is feels crufty to me.

michaeldiamant (approved):
@algoidurovic Thanks for your effort here - I'm excited to see the feature used by the community!  ☕ 

I left an optional remark in https://github.com/algorand/pyteal/pull/274#discussion_r856485694.  I'd appreciate considering it before merging though it need _not_ hold us up.

docs/opup.rst
```diff
@@ -0,0 +1,85 @@
...
+is exceeded, the evaluation will fail. The typical usage pattern is to insert the :any:`pyteal.OpUp.ensure_budget`
+call just before a particularly heavyweight subroutine or expression. Keep in mind that the required budget
+expression will be evaluated before the inner transactions are issued so it may be prudent to avoid expensive
+expressions, which may exhaust the budget before it can be increased.
+
+In the example below, the :py:meth:`pyteal.Ed25519Verify` expression is used, which costs 1,900.
+
+.. code-block:: python
+
+    target_app_id = Int(1) # the application id to be called when more budget is requested
```
jasonpaulos, line 44: nit: something more explicit saying this should be replaced with an app ID the user provides would help

docs/opup.rst
```diff
@@ -0,0 +1,85 @@
...
+
+    # OnCall mode works the exact same way, just omit the target_app_id
+    opup = OpUp(OpUpMode.Explicit, target_app_id)
+    program = Seq(
+        If(Txn.application_id() != Int(0)).Then(
+            Seq(
+                opup.ensure_budget(Int(2000)),
+                Pop(Ed25519Verify(args[0], args[1], args[2])),
```
jasonpaulos, line 52: nit: I think `Assert` would be a better example than `Pop`

docs/opup.rst
```diff
@@ -0,0 +1,85 @@
...
+            )
+        ),
+        Approve(),
+    )
+
+:any:`pyteal.OpUp.maximize_budget` attempts to issue as many inner transactions as possible with the given fee.
```
jasonpaulos, line 58: nit: would be nice if this section for `maximize_budget` and the previous section for `ensure_budget` were under different subheadings to break up the explanations & examples

docs/opup.rst
```diff
@@ -0,0 +1,85 @@
...
+This essentially maximizes the available budget while putting a ceiling on the amount of fee spent. Just
+as with :any:`pyteal.OpUp.ensure_budget`, the evaluation will fail if there aren't enough funds for issuing the
+inner transactions or the inner transaction limit is exceeded. This method may be preferred to
+:any:`pyteal.OpUp.ensure_budget` when the fee spent on increasing budget needs to be capped or if the developer
+would rather just maximize the available budget instead of doing in depth cost analysis on the program.
+
+In the example below, the fee is capped at 3,000 microAlgos for increasing the budget. This works out to 3 inner
+transactions being issued, each increasing the available budget by 700.
```
jasonpaulos, line 66: I'd say ~700, since there is some overhead

docs/opup.rst
```diff
@@ -0,0 +1,85 @@
...
+========================
+
+Some opcode budget is consumed during execution of every Algorand Smart Contract because every teal
+instruction has a corresponding cost. In order for the evaluation to succeed, the budget consumed must not
+exceed the budget provided. This constraint may introduce a problem for expensive contracts that quickly
+consume the initial budget of 700. The OpUp budget increase utility provides a workaround using inner
+transactions. Note that there is a context specific limit to the number of inner transactions issued in a
+transaction group so budget cannot be increased arbitrarily. The available budget when using the OpUp
+utility will need to be high enough to execute the teal code that issues the inner transactions. A budget
+of ~20 is enough for most use cases.
```
jasonpaulos, line 13: nit: it would be nice if the overview explicitly stated that how this works (by executing noop inner txns that increase the group's pooled compute budget) and where the funding comes from (the app's account, not the caller or creator's). The first is a nice-to-have, but the second seems more important for practical purposes

jasonpaulos (commented):
The docs look great, I just have some suggestions, only one of which is important (the 700 vs ~700 comment).

examples/application/program.teal
algoidurovic: Sorry, this file should've been deleted. Taken care of now.

examples/application/opup.py
```diff
@@ -0,0 +1,97 @@
+# This example is provided for informational purposes only and has not been audited for security.
+
+from pyteal import *
+
+from pyteal.ast.opup import OpUp, OpUpMode
```
jasonpaulos, line 5: This line shouldn't be necessary

jasonpaulos (commented):
Looks great, I just have one comment that I missed before

pyteal/ast/opup.py
```diff
@@ -0,0 +1,157 @@
...
+
+    # The app to call must be provided by the user.
+    Explicit = 0
+
+    # The app to call is created then deleted for each request to increase budget.
+    OnCall = 1
+
+
+ON_CALL_APP = Bytes("base16", "068101")  # v6 teal program "int 1"
+MIN_TXN_FEE = Int(1000)
```
ahangsu, line 33: minor proposal, do you think `MIN_TXN_FEE` should be available in a larger scope, say it lives in `pyteal/config.py`? I guess some folks would want this available not just in `opup`.

pyteal/ast/opup.py
```diff
@@ -0,0 +1,157 @@
...
+        consider moving the call to ensure_budget() earlier in the pyteal program."""
+        require_type(required_budget, TealType.uint64)
+
+        # A budget buffer is necessary to deal with an edge case of ensure_budget():
+        #   if the current budget is equal to or only slightly higher than the
+        #   required budget then it's possible for ensure_budget() to return with a
+        #   current budget less than the required budget. The buffer prevents this
+        #   from being the case.
+        buffer = Int(10)
+        buffered_budget = ScratchVar(TealType.uint64)
```
ahangsu, line 132: just one question: is these lines of code ensuring that in the case of: *current budget slightly higher than required budget*, the `_construct_itxn` will be executed at least once?

Since I am not sure if I understand
> it's possible for ensure_budget() to return with a current budget less than the required budget.

ahangsu (commented):
just one question about `ensure_budget`

pyteal/ast/opup.py
jasonpaulos: `Global.min_txn_fee()` is available to PyTeal, perhaps this code should use that instead?

ahangsu: ^ sounds pretty reasonable to me

examples/application/program.teal
michaeldiamant: @algoidurovic No worries - Thanks for the change in https://github.com/algorand/pyteal/pull/274/commits/a9d22ea6a16fcf4a0acbab00760395668d94b325.  Resolving.

pyteal/ast/opup.py
algoidurovic: Say the user calls the method with `required_budget=1000` then, after issuing an inner app call, the current budget is 1000. When the comparison `current_budget < required_budget` is executed it will return false, but the comparison itself consumes budget so the resulting budget may be something like 999 instead of 1000.

PR #275, opened 2022-04-12
Test py-algorand-sdk Python version bump

michaeldiamant:
Provides a proof point that https://github.com/algorand/py-algorand-sdk/pull/313 works for downstream consumers.

I create the PR only to support running the build.  I intend to close the PR after successful build.


michaeldiamant:
Closing - The PR provided the needed proof point.

PR #269, opened 2022-04-11
Upgrade minimum Python version to 3.10

michaeldiamant:
https://github.com/algorand/pyteal/issues/244 - Upgrades minimum Python version 3.10.

Details:
* Reuses concurrent Python version tooling suggestion from https://github.com/algorand/py-algorand-sdk/pull/313/.

PR #276, opened 2022-04-13
Add readthedocs config file

jasonpaulos:
Since merging #269, our readthedocs build has failed (as an example, see https://readthedocs.org/projects/pyteal/builds/16635690/), and I believe adding this configuration will properly tell it to use Python 3.10 to build our docs.

If this works, a similar thing will likely need to be added to the Python SDK following https://github.com/algorand/py-algorand-sdk/pull/313 (cc @algochoi)


michaeldiamant (approved):
@jasonpaulos Thanks for catching!

I'll follow up out-of-band to understand how to better monitor the build.

A few thoughts:
* Reading the docs more closely, I see it specifies `.yaml` extension.  We previously defined `.readthedocs.yml`.  Presumably the file extension explains the build failure.
* I don't feel strongly about making the change in this PR, but it feels like we ought to delete `.readthedocs.yml`.
 * Additionally, `.readthedocs.yml` explicitly specifies `python.version: 3.10`.  From my read of the docs, `build.os.tools` _seems equivalent.  Though it feels prudent to set both once the build works (or to make it work).

.readthedocs.yaml
```diff
@@ -0,0 +1,29 @@
...
+   configuration: docs/conf.py
+
+# If using Sphinx, optionally build your docs in additional formats such as PDF
+# formats:
+#    - pdf
+
+# Optionally declare the Python requirements required to build your docs
+python:
+   install:
+   - requirements: docs/requirements.txt
```
tzaffi, line 29:
```suggestion
   - requirements: docs/requirements.txt

```

tzaffi (approved):
LGTM - (even without the suggested extra line in `.readthedocs.yaml`)

algochoi:
Can this yaml file be copied over to the `py-algorand-sdk` directly? It looks like that repo doesn't define one yet.

jasonpaulos:
> Can this yaml file be copied over to the `py-algorand-sdk` directly? It looks like that repo doesn't define one yet.

Likely yes

PR #278, opened 2022-04-14
ABI Strings

### LINKED ISSUE #280: ABI data type support:  String and Address
## Problem

https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0004.md supports `String` and `Address` data types.  To help support ARC-4 application App definition and calls, 1st class support for these data types is needed.

## Dependencies

N/A

## Urgency

TBD

barnjamin:
Adding Address and String that subclass other Array types with hardcoded type specs

Closes #280 

- [x] Address type def
- [x] Address Testing
- [x] String type def
- [x] String testing

Follow up PR should implement setting directly from a byte string and mayybee a slice overload for __getitem__ on string?

### DIFF
--- a/pyteal/ast/abi/__init__.py
+++ b/pyteal/ast/abi/__init__.py
@@ -1,5 +1,5 @@
 from .string import String, StringTypeSpec
-from .address import AddressTypeSpec, Address
+from .address import AddressTypeSpec, Address, ADDRESS_LENGTH
 from .type import TypeSpec, BaseType, ComputedValue
 from .bool import BoolTypeSpec, Bool
 from .uint import (
@@ -38,6 +38,7 @@ __all__ = [
     "StringTypeSpec",
     "Address",
     "AddressTypeSpec",
+    "ADDRESS_LENGTH",
     "TypeSpec",
     "BaseType",
     "ComputedValue",
--- a/pyteal/ast/abi/__init__.py
+++ b/pyteal/ast/abi/__init__.py
@@ -1,3 +1,4 @@
+from .string import String, Address, StringTypeSpec, AddressTypeSpec
 from .type import TypeSpec, BaseType, ComputedValue
 from .bool import BoolTypeSpec, Bool
 from .uint import (
@@ -32,6 +33,10 @@ from .util import type_spec_from_annotation
 from .method_return import MethodReturn
 
 __all__ = [
+    "String",
+    "StringTypeSpec",
+    "Address",
+    "AddressTypeSpec",
     "TypeSpec",
     "BaseType",
     "ComputedValue",
--- a/pyteal/ast/abi/__init__.py
+++ b/pyteal/ast/abi/__init__.py
@@ -1,4 +1,5 @@
-from .string import String, Address, StringTypeSpec, AddressTypeSpec
+from .string import String, StringTypeSpec
+from .address import AddressTypeSpec, Address
 from .type import TypeSpec, BaseType, ComputedValue
 from .bool import BoolTypeSpec, Bool
 from .uint import (
--- a/pyteal/ast/abi/address.py
+++ b/pyteal/ast/abi/address.py
@@ -2,12 +2,12 @@ from .array_static import StaticArray, StaticArrayTypeSpec
 from .uint import ByteTypeSpec
 from ..expr import Expr
 
-address_length = 32
+ADDRESS_LENGTH = 32
 
 
 class AddressTypeSpec(StaticArrayTypeSpec):
     def __init__(self) -> None:
-        super().__init__(ByteTypeSpec(), address_length)
+        super().__init__(ByteTypeSpec(), ADDRESS_LENGTH)
 
     def new_instance(self) -> "Address":
         return Address()
@@ -21,7 +21,7 @@ AddressTypeSpec.__module__ = "pyteal"
 
 class Address(StaticArray):
     def __init__(self) -> None:
-        super().__init__(AddressTypeSpec(), address_length)
+        super().__init__(AddressTypeSpec(), ADDRESS_LENGTH)
 
     def type_spec(self) -> AddressTypeSpec:
         return AddressTypeSpec()
--- a/pyteal/ast/abi/address.py
+++ b/pyteal/ast/abi/address.py
@@ -1,0 +1,33 @@
+from .array_static import StaticArray, StaticArrayTypeSpec
+from .uint import ByteTypeSpec
+from ..expr import Expr
+
+address_length = 32
+
+
+class AddressTypeSpec(StaticArrayTypeSpec):
+    def __init__(self) -> None:
+        super().__init__(ByteTypeSpec(), address_length)
+
+    def new_instance(self) -> "Address":
+        return Address()
+
+    def __str__(self) -> str:
+        return "address"
+
+
+AddressTypeSpec.__module__ = "pyteal"
+
+
+class Address(StaticArray):
+    def __init__(self) -> None:
+        super().__init__(AddressTypeSpec(), address_length)
+
+    def type_spec(self) -> AddressTypeSpec:
+        return AddressTypeSpec()
+
+    def get(self) -> Expr:
+        return self.stored_value.load()
+
+
+Address.__module__ = "pyteal"
--- a/pyteal/ast/abi/address_test.py
+++ b/pyteal/ast/abi/address_test.py
@@ -1,21 +1,25 @@
 from .type_test import ContainerType
-from os import urandom
 
 from ... import *
 
 options = CompileOptions(version=5)
 
+
 def test_AddressTypeSpec_str():
     assert str(abi.AddressTypeSpec()) == "address"
-        
+
+
 def test_AddressTypeSpec_is_dynamic():
     assert not (abi.AddressTypeSpec()).is_dynamic()
 
-def test_AddressTypeSpec_is_dynamic():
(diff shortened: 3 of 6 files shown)


jasonpaulos (commented):
Approach looks correct. The only thing this is "missing" (which can be added in a later PR, I don't think we're about to decide how to do this right now) is a way to convert between equivalent types, like `address` -> `[32]byte`

pyteal/ast/abi/address_test.py
```diff
@@ -0,0 +1,78 @@
...
+    expected = TealSimpleBlock([TealOp(expr, Op.load, value.stored_value.slot)])
+    actual, _ = expr.__teal__(options)
+    assert actual == expected
+
+
+def test_Address_decode():
+    from os import urandom
+
+    value = abi.Address()
+    for value_to_set in [urandom(32) for x in range(10)]:
```
ahangsu, line 50: nit: reuse the `address_length`?

pyteal/ast/abi/address.py
```diff
@@ -0,0 +1,33 @@
+from .array_static import StaticArray, StaticArrayTypeSpec
+from .uint import ByteTypeSpec
+from ..expr import Expr
+
+address_length = 32
```
ahangsu, line 5: nit: probably uppercase this for it's being a constant, and probably export this?

ahangsu (commented):
seems correct to me, just some minor thoughts

ahangsu (approved):
lgtm, waiting on JasonP's comments

pyteal/ast/abi/__init__.py
```diff
@@ -1,3 +1,5 @@
+from .string import String, StringTypeSpec
```
tzaffi, line 1: FYI: as of #283 - we're going to be using absolute imports. So this PR may need to follow those conventions, depending on which PR gets merged in first.

tzaffi: Additionally, can you please merge in the latest algorand origin master.

tzaffi: Please ignore the above suggestions. I see now that this is a _relative_ PR against the ABI branch, and therefore it's up to `feature/abi` to resolve this issue

pyteal/ast/abi/address_test.py
```diff
@@ -0,0 +1,76 @@
...
+        abi.DynamicArrayTypeSpec(abi.ByteTypeSpec()),
+    ):
+        assert abi.AddressTypeSpec() != otherType
+
+
+def test_Address_encode():
+    value = abi.Address()
+    expr = value.encode()
+    assert expr.type_of() == TealType.bytes
+    assert not expr.has_return()
```
tzaffi, line 37: similar suggestion to above: prefer `is`

pyteal/ast/abi/address_test.py
```diff
@@ -0,0 +1,76 @@
...
+
+        with TealComponent.Context.ignoreExprEquality():
+            assert actual == expected
+
+
+def test_Address_get():
+    value = abi.Address()
+    expr = value.get()
+    assert expr.type_of() == TealType.bytes
+    assert not expr.has_return()
```
tzaffi, line 72: prefer `is not` ...

pyteal/ast/abi/string_test.py
```diff
@@ -0,0 +1,79 @@
...
+
+options = CompileOptions(version=5)
+
+
+def test_StringTypeSpec_str():
+    assert str(abi.StringTypeSpec()) == "string"
+
+
+def test_StringTypeSpec_is_dynamic():
+    assert (abi.StringTypeSpec()).is_dynamic()
```
tzaffi, line 11: prefer `is`

pyteal/ast/abi/string_test.py
```diff
@@ -0,0 +1,79 @@
...
+        abi.DynamicArrayTypeSpec(abi.Uint8TypeSpec()),
+    ):
+        assert abi.StringTypeSpec() != otherType
+
+
+def test_String_encode():
+    value = abi.String()
+    expr = value.encode()
+    assert expr.type_of() == TealType.bytes
+    assert not expr.has_return()
```
tzaffi, line 33: prefer `is`

pyteal/ast/abi/string_test.py
```diff
@@ -0,0 +1,79 @@
...
+def test_String_decode():
+    import random
+    from os import urandom
+
+    value = abi.String()
+    for value_to_set in [urandom(random.randint(0, 50)) for x in range(10)]:
+        expr = value.decode(Bytes(value_to_set))
+
+        assert expr.type_of() == TealType.none
+        assert not expr.has_return()
```
tzaffi, line 49: prefer `is`

pyteal/ast/abi/string_test.py
```diff
@@ -0,0 +1,79 @@
...
+
+        with TealComponent.Context.ignoreExprEquality():
+            assert actual == expected
+
+
+def test_String_get():
+    value = abi.String()
+    expr = value.get()
+    assert expr.type_of() == TealType.bytes
+    assert not expr.has_return()
```
tzaffi, line 69: ... `is`

pyteal/ast/abi/string_test.py
```diff
@@ -0,0 +1,79 @@
...
+
+    expected = TealSimpleBlock(
+        [TealOp(expr, Op.load, value.stored_value.slot), TealOp(None, Op.extract, 2, 0)]
+    )
+    actual, _ = expr.__teal__(options)
```
tzaffi, line 74: optional: just as you had random binary strings used for `Address` some random strings of random lengths could also be useful here for extra confidence.

tzaffi (commented):
Very useful indeed! I left some optional suggestions.

pyteal/ast/abi/__init__.py
tzaffi: Update: In fact, #284 and #285 do address these issues.

michaeldiamant:
Merging - I'll open a subsequent PR to reflect https://github.com/algorand/pyteal/pull/286 in this PR.

Since this PR is opened on a fork, it'll be more work to update the existing PR than to merge into a feature branch and provide a new PR.  _If_ the merge is into `master`, I'd find a way to update this PR first.  It is _rare_ that I condone breaking a build on merge.

jasonpaulos (commented):
My review is late, but I brought up some points that should be addressed before this can be released. @barnjamin if you agree with the comments, could you open a follow up PR? No rush, I just don't want us to forget about it

edit: or maybe incorporate the changes into #289?

pyteal/ast/abi/address_test.py
```diff
@@ -0,0 +1,76 @@
...
+        abi.DynamicArrayTypeSpec(abi.ByteTypeSpec()),
+    ):
+        assert abi.AddressTypeSpec() != otherType
+
+
+def test_Address_encode():
+    value = abi.Address()
+    expr = value.encode()
+    assert expr.type_of() == TealType.bytes
+    assert expr.has_return() is False
```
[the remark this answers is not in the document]
tzaffi: That's how it was before... I asked for this version because `has_return()` is boolean so you can pin the value precisely without worrying about truthiness.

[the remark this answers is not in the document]
tzaffi: (it's not really a big deal. In theory, we've tested that `has_return()` is boolean elsewhere and works as expected so that using the more concise _truthy_ version is just as good)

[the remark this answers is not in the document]
jasonpaulos: I see, you're concerned that `not expr.has_return()` would also appear true if `expr.has_return()` returned some other false-y value, like 0? That's something I've not considered before...

I suppose you'd also be in favor of `assert X is True` over `assert X` if we wanted to test that a value is _really_ true, not merely truth-y?

I don't love the syntax but the argument makes a lot of sense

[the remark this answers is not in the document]
barnjamin: so going with is False?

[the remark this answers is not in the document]
tzaffi: Just to answer @jasonpaulos - yeah. I would also use `is True` or `X is None` instead of `not X` ... etc ...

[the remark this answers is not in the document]
jasonpaulos: Thanks for the discussion, let's keep it as `is False` then.

pyteal/ast/abi/address_test.py
```diff
@@ -0,0 +1,76 @@
...
+def test_AddressTypeSpec_new_instance():
+    assert isinstance(abi.AddressTypeSpec().new_instance(), abi.Address)
+
+
+def test_AddressTypeSpec_eq():
+    assert abi.AddressTypeSpec() == abi.AddressTypeSpec()
+
+    for otherType in (
+        abi.ByteTypeSpec,
+        abi.StaticArrayTypeSpec(abi.ByteTypeSpec(), 31),
```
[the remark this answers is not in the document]
barnjamin: Hm, with 32 it fails the assert since they're seen as equal, I guess this means I should implement an __eq__ method on Address and raise an error if the other is not strictly an Address type?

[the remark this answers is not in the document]
jasonpaulos: Yep exactly

PR #279, opened 2022-04-14
Pin versions of docs dependencies

jasonpaulos:
Based on the failed build https://readthedocs.org/projects/pyteal/builds/16655648/, readthedocs is by default installing older versions of the following relevant libraries:
* `sphinx` <2
* `sphinx-rtd-theme` <0.5

Pinning them to a specific Python 3.10 supported version might fix the build.

Follow up to #276.


michaeldiamant (approved):
@jasonpaulos Seems reasonable.  Let's try it.

Aside:  https://github.com/sphinx-doc/sphinx/issues/9562 suggests 4.2.0 is the _minimum_ version supporting Python 3.10.  I see latest release is 4.5.0.  We can revisit updating further later on.  It feels secondary at best.

PR #283, opened 2022-04-19
Use absolute imports

jasonpaulos:
Lately our relative imports have been a source of confusion/problems, and as of #273, absolute imports are now used by test code, so we might as well make the switch to absolute imports everywhere for consistency.

Though if there are objections, feel free to voice them.

Of particular note is the `CONTRIBUTING.md` file, in which I documented for the first time in this repo how our module and import conventions should work.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,33 +1,20 @@
 # Contribution Guide
 
-If you are interested in contributing to the project, we welcome and thank you. We want to make the
-best decentralized and effective blockchain platform available and we appreciate your willingness to
-help us.
+If you are interested in contributing to the project, we welcome and thank you. We want to make the best decentralized and effective blockchain platform available and we appreciate your willingness to help us.
 
 ## Filing Issues
 
-Did you discover a bug? Do you have a feature request? Filing issues is an easy way anyone can
-contribute and helps us improve PyTeal. We use GitHub Issues to track all known bugs and feature
-requests.
+Did you discover a bug? Do you have a feature request? Filing issues is an easy way anyone can contribute and helps us improve PyTeal. We use GitHub Issues to track all known bugs and feature requests.
 
-Before logging an issue be sure to check current issues, check the [open GitHub issues][issues_url]
-to see if your issue is described there.
+Before logging an issue be sure to check current issues, check the [open GitHub issues][issues_url] to see if your issue is described there.
 
-If you’d like to contribute to any of the repositories, please file a [GitHub issue][issues_url]
-using the issues menu item. Make sure to specify whether you are describing a bug or a new
-enhancement using the **Bug report** or **Feature request** button.
+If you’d like to contribute to any of the repositories, please file a [GitHub issue][issues_url] using the issues menu item. Make sure to specify whether you are describing a bug or a new enhancement using the **Bug report** or **Feature request** button.
 
 See the GitHub help guide for more information on [filing an issue](https://help.github.com/en/articles/creating-an-issue).
 
 ## Contribution Model
 
-For each of our repositories we use the same model for contributing code. Developers wanting to 
-contribute must create pull requests. This process is described in the GitHub [Creating a pull request from a fork](https://help.github.com/en/articles/creating-a-pull-request-from-a-fork)
-documentation. Each pull request should be initiated against the master branch in the Algorand
-repository.  After a pull request is submitted the core development team will review the submission
-and communicate with the developer using the comments sections of the PR. After the submission is
-reviewed and approved, it will be merged into the master branch of the source. These changes will be
-merged to our release branch on the next viable release date.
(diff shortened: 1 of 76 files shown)


CONTRIBUTING.md
```diff
@@ -2,33 +2,79 @@
...
 
 In order to retain a consistent style, we prefer to continue deviating from PEP 8 naming conventions in the following cases.  We try to balance minimizing exceptions against providing a consistent style for existing software.
 * Enums - Define with lowercase camelcase.  Example:  https://github.com/algorand/pyteal/blob/7c953f600113abcb9a31df68165b61a2c897f591/pyteal/ast/txn.py#L37
 * Factory methods - Define following [class name](https://peps.python.org/pep-0008/#class-names) conventions.  Example:  https://github.com/algorand/pyteal/blob/7c953f600113abcb9a31df68165b61a2c897f591/pyteal/ast/ternaryexpr.py#L63
 
 Since it's challenging to enforce these exceptions with a linter, we rely on PR creators and reviewers to make a best-effort attempt to enforce agreed upon naming conventions.
+
+### Module Guidelines
+
+Every directory containing source code should be a Python module, meaning it should have an `__init__.py` file. This `__init__.py` file is responsible for exporting all public objects (i.e.  classes, functions, and constants) for use outside of that module.
```
tzaffi, line 36:
```suggestion
Every directory containing source code should be a Python module, meaning it should have an `__init__.py` file. This `__init__.py` file is responsible for exporting all public objects (i.e. classes, functions, and constants) for use outside of that module.
```

tzaffi: (removing a space before " classes")

CONTRIBUTING.md
```diff
@@ -2,33 +2,79 @@
...
+
+Modules may be created inside of other modules, in which case the deeper module is called a submodule or child module, and the module that contains it is called the parent module. For example, `pyteal` is the parent module to `pyteal.ast`.
+
+A sibling module is defined as a different child module of the parent module. For example, `pyteal.ast` and `pyteal.ir` as sibling modules.
```
tzaffi, line 40:
```suggestion
A sibling module is defined as a different child module of the parent module. For example, `pyteal.ast` and `pyteal.ir` are sibling modules.
```

CONTRIBUTING.md
```diff
@@ -2,33 +2,79 @@
...
+
+### Module Guidelines
+
+Every directory containing source code should be a Python module, meaning it should have an `__init__.py` file. This `__init__.py` file is responsible for exporting all public objects (i.e. classes, functions, and constants) for use outside of that module.
+
+Modules may be created inside of other modules, in which case the deeper module is called a submodule or child module, and the module that contains it is called the parent module. For example, `pyteal` is the parent module to `pyteal.ast`.
+
+A sibling module is defined as a different child module of the parent module. For example, `pyteal.ast` and `pyteal.ir` are sibling modules.
+
+### Import Guidelines
```
michaeldiamant, line 42: Leaving a note to document a verbal with @jasonpaulos:

* There _might_ be a flake8 plugin we can use to mechanically check the import guidelines:  https://pypi.org/project/flake8-absolute-import/. 
* I've identified an option exists, but didn't evaluate it.  @jasonpaulos will take a quick look.  If it works out-of-the-box and looks straight-forward, then we can include it here.  Otherwise, we don't want to hold the PR up.

michaeldiamant: Sorry - I confused myself.  There's 2 options that might work for us.  Same disclaimer as above:
* https://pypi.org/project/flake8-tidy-imports/
* https://pypi.org/project/flake8-absolute-import/

Both projects appear to be small scale though recently updated.  A quick scan suggests _flake8-tidy-imports_ is a bit more flexible.  Unclear if that's desirable.

CONTRIBUTING.md
```diff
@@ -2,33 +2,79 @@
...
+
+#### In Runtime Code
+
+When a runtime file in this codebase needs to import another module/file in this codebase, you should import the absolute path of the module/file, not the relative one, using the `from X import Y` method.
```
tzaffi, line 46: Do you want to also allow for the possibility of importing the module without `from`, e.g. `mypkg.sibling`  and `mypkg.sibling as sister`? (I got the first example from [pep-8](https://peps.python.org/pep-0008/#imports))

CONTRIBUTING.md
jasonpaulos: Added `flake8-tidy-imports`, thanks for the suggestion

CONTRIBUTING.md
jasonpaulos: For simplicity, I'd say _no_, let's not allow that, but I'm open to revisiting if there's a motivating use case.

michaeldiamant (approved):
@jasonpaulos I'm on-board to try these options.  Thanks for evaluating _flake8-tidy-imports_.

CONTRIBUTING.md
tzaffi: One motivation is for the tests use case where we use `import pyteal as pt` or in general, or in general, importing many objects like `from X import Y1, Y2, Y3, .... , Y30` and it gets pretty gnarly. 
There is one obligatory use case: avoiding namespace collisions as was mentioned in the PEP itself.

tzaffi (approved):
Great improvement to code quality and ease of development!

PR #289, opened 2022-04-21
Add `set` on address and string, change array init to accept typespec

barnjamin:
Initial attempt at set for both Address and String ABI types. 

One complication of the String type is that it may be prefixed with length and we want to be careful not to prepend length if it is already present.

> for set the user should pass in the actual value they want to set (i.e. no prefix with length for strings/arrays). If they have a prefixed version already, they should call decode.


pyteal/ast/__init__.py
```diff
@@ -140,6 +140,8 @@
 # abi
 from pyteal.ast import abi
 
+# from pyteal.ast.abi import *
```
barnjamin, line 143: I can remove this but also we should figure out how to expose the stuff in ABI to editors

pyteal/ast/abi/address.py
```diff
@@ -15,19 +24,56 @@ def new_instance(self) -> "Address":
...
+        return isinstance(other, AddressTypeSpec)
+
 
 AddressTypeSpec.__module__ = "pyteal"
 
 
 class Address(StaticArray):
     def __init__(self) -> None:
-        super().__init__(AddressTypeSpec(), ADDRESS_LENGTH)
+        super().__init__(AddressTypeSpec())
```
ahangsu, line 36: If I am understanding correctly, you are trying to init `Address` object on a chain of inheritance:
```
BaseType -> StaticArray -> Address
```
A direct `super` method only allow you to inherit from `StaticArray`, which requires you to provide:
- argument TypeSpec
- array length

One way to get around is: init from grandparent object, e.g., from this link: https://stackoverflow.com/questions/62744176/how-to-overwrite-parent-class-init-method-and-use-super-to-call-grandparent-ini

Such technique would allow you to cause smaller change to interfaces for classes including:
- DynamicArray
- StaticArray
- Tuple

pyteal/ast/abi/string_test.py
```diff
@@ -20,9 +25,10 @@ def test_StringTypeSpec_eq():
     assert abi.StringTypeSpec() == abi.StringTypeSpec()
 
     for otherType in (
-        abi.ByteTypeSpec,
+        abi.ByteTypeSpec(),
```
ahangsu, line 28: good catch

pyteal/ast/abi/string.py
```diff
@@ -28,10 +46,45 @@ def type_spec(self) -> StringTypeSpec:
...
 
+    def set(
+        self,
+        value: Union[
+            Sequence[T],
+            DynamicArray[T],
+            ComputedValue[DynamicArray[T]],
+            "String",
+            str,
+            Expr,
```
ahangsu, line 61: same here: do we have the implementation supporting all these types? For `Sequence[T]` and `DynamicArray[T]`, support seems lacking.

ahangsu (commented):
initial pass on the implementation, the main concern is on the interface change for object initialization for `Array` and `Tuple`.

barnjamin:
> the main concern is on the interface change for object initialization for `Array` and `Tuple`.

This was done based on suggestion here:

https://github.com/algorand/pyteal/pull/278#discussion_r855381049

pyteal/ast/abi/string.py
```diff
@@ -28,9 +46,56 @@ def type_spec(self) -> StringTypeSpec:
...
+            )
+
+        # Assume not length prefixed
+        if isinstance(value, str) or isinstance(value, bytes):
+            return self.stored_value.store(encoded_string(Bytes(value)))
+
+        if isinstance(value, Expr):
+            return self.stored_value.store(encoded_string(value))
+
+        if isinstance(value, Sequence):
```
ahangsu, line 94: oh I forgot, `Sequence` comments also apply here...

pyteal/ast/abi/address.py
```diff
@@ -1,33 +1,99 @@
...
+        elif isinstance(value, str) and len(value) == ADDRESS_LENGTH_STR:
+            return self.stored_value.store(Addr(value))
+
+        elif isinstance(value, bytes) and len(value) == ADDRESS_LENGTH_BYTES:
+            return self.stored_value.store(Bytes(value))
+
+        elif isinstance(value, Expr):
+            return self.stored_value.store(value)
+
+        elif isinstance(value, Sequence):
```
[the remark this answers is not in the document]
barnjamin: I'm not sure what to do on this one, @jasonpaulos or @tzaffi ?

I could do a different check like to see if `.__iter__` is an attr or just try/catch against super().set and raise TealInputError?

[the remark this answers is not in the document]
tzaffi: `isinstance(value, Iterable)` [will probably do the trick](https://stackoverflow.com/questions/1952464/in-python-how-do-i-determine-if-an-object-is-iterable#answer-61139278) but it depends on if it has an `.__iter__` and not just a `__get_item__`. 

On a different matter, I think this is a great candidate for exercising your[ py 3.10 pattern matching muscles](https://towardsdatascience.com/python-structural-pattern-matching-top-3-use-cases-to-get-you-started-262160007fa0). (see 3rd use case in that Medium article)

pyteal/ast/abi/address.py
```diff
@@ -1,33 +1,101 @@
...
         return "address"
 
+    def __eq__(self, other: object) -> bool:
+        return isinstance(other, AddressTypeSpec)
+
 
 AddressTypeSpec.__module__ = "pyteal"
 
 
 class Address(StaticArray):
```
jasonpaulos, line 37: No need for `TypeVar`s `T` and `N` in this file. Instead, you should make `Address` inherit from `StaticArray[Byte, Literal[32]]` (or replace 32 with `ADDRESS_LENGTH_BYTES`).

Then all instances of `T` and `N` in the rest of the file can be replaced with `Byte` and `Literal[32]`, respectively.

Same thing applies to `string.py`

pyteal/ast/abi/address.py
```diff
@@ -1,33 +1,101 @@
...
+            "Address",
+            str,
+            bytes,
+            Expr,
+        ],
+    ):
+
+        match value:
+            case ComputedValue():
+                if value.produced_type_spec() == AddressTypeSpec():
```
jasonpaulos, line 62: Should this also check for equality to `StaticArrayTypeSpec(ByteTypeSpec(), ADDRESS_LENGTH_BYTES)`?

pyteal/ast/abi/address.py
```diff
@@ -1,33 +1,101 @@
...
+                raise TealInputError(
+                    f"Got bytes with length {len(value)}, expected {ADDRESS_LENGTH_BYTES}"
+                )
+            case Expr():
+                return self.stored_value.store(value)
+            case CollectionSequence():
+                return super().set(value)
+
+        raise TealInputError(
+            f"Got {type(value)}, expected StaticArray, ComputedValue, String, str, bytes, Expr"
```
jasonpaulos, line 97: nit: needs updating, `String` not relevant and missing `Sequence` and `Address`

pyteal/ast/abi/address_test.py
```diff
@@ -1,5 +1,12 @@
+import pytest
 import pyteal as pt
 from pyteal import abi
+from pyteal.ast.abi.address import ADDRESS_LENGTH_BYTES
+
+from pyteal.ast.global_ import Global, GlobalField
```
jasonpaulos, line 6: `Global` and `GlobalField` should be imported in the `from pyteal` line, since they are accessible from the top-level package

`abi.ADDRESS_LENGTH_BYTES` should be used, not `from pyteal.ast.abi.address import ADDRESS_LENGTH_BYTES`, so that we're testing what's exported

Recently added this justification to the repo here: https://github.com/algorand/pyteal/blob/master/CONTRIBUTING.md#in-test-code

pyteal/ast/abi/address_test.py
```diff
@@ -45,37 +52,210 @@ def test_Address_encode():
...
+    )
+    actual, _ = expr.__teal__(options)
+    assert actual == expected
+
 
+def test_Address_set_StaticArray():
+    sa = abi.StaticArray(
+        abi.StaticArrayTypeSpec(abi.ByteTypeSpec(), ADDRESS_LENGTH_BYTES)
+    )
+    for value_to_set in (sa,):
```
jasonpaulos, line 113: nit: iterating over one value?

pyteal/ast/abi/tuple.py
```diff
@@ -218,7 +218,7 @@ def length_static(self) -> int:
         return len(self.value_specs)
 
     def new_instance(self) -> "Tuple":
-        return Tuple(*self.value_specs)
+        return Tuple(TupleTypeSpec(*self.value_specs))
```
jasonpaulos, line 221: nit: `Tuple(self)` is simpler

pyteal/ast/abi/tuple.py
```diff
@@ -381,7 +381,9 @@ def __init__(
         value2_type_spec: TypeSpec,
         value3_type_spec: TypeSpec,
     ) -> None:
-        super().__init__(value1_type_spec, value2_type_spec, value3_type_spec)
+        super().__init__(
+            TupleTypeSpec(value1_type_spec, value2_type_spec, value3_type_spec)
```
jasonpaulos, line 385: I'm afraid for consistency, these would have to accept a single `TupleTypeSpec` as an arg and error if the number of values in that type spec is different than what this class should expect.

pyteal/ast/abi/array_dynamic.py
```diff
@@ -20,7 +20,7 @@
 
 class DynamicArrayTypeSpec(ArrayTypeSpec[T]):
     def new_instance(self) -> "DynamicArray[T]":
-        return DynamicArray(self.value_type_spec())
+        return DynamicArray(DynamicArrayTypeSpec(self.value_type_spec()))
```
jasonpaulos, line 23: same thing about using `DynamicArray(self)` instead

pyteal/ast/abi/array_dynamic.py
```diff
@@ -47,8 +47,8 @@ def __str__(self) -> str:
 class DynamicArray(Array[T]):
     """The class that represents ABI dynamic array type."""
 
-    def __init__(self, value_type_spec: TypeSpec) -> None:
-        super().__init__(DynamicArrayTypeSpec(value_type_spec))
+    def __init__(self, array_type_spec: DynamicArrayTypeSpec) -> None:
```
jasonpaulos, line 50: For more descriptive types, should be `DynamicArrayTypeSpec[T]`

pyteal/ast/abi/array_static.py
```diff
@@ -28,7 +28,9 @@ def __init__(self, value_type_spec: TypeSpec, array_length: int) -> None:
         self.array_length: Final = array_length
 
     def new_instance(self) -> "StaticArray[T, N]":
-        return StaticArray(self.value_type_spec(), self.length_static())
+        return StaticArray(
+            StaticArrayTypeSpec(self.value_type_spec(), self.length_static())
```
jasonpaulos, line 32: same comment here about `self`

pyteal/ast/abi/array_static.py
```diff
@@ -72,8 +74,8 @@ def __str__(self) -> str:
 class StaticArray(Array[T], Generic[T, N]):
     """The class that represents ABI static array type."""
 
-    def __init__(self, value_type_spec: TypeSpec, array_length: int) -> None:
-        super().__init__(StaticArrayTypeSpec(value_type_spec, array_length))
+    def __init__(self, array_type_spec: StaticArrayTypeSpec) -> None:
```
jasonpaulos, line 77: same thing here about `[T]`

pyteal/ast/abi/address.py
jasonpaulos: Good point, that would work, but it's a little hacky.

I think this change is acceptable since we'll likely want to support more subtypes in the future, so might as well make it easier now.

The only downside is that it makes it more cumbersome for users to create objects. But in my opinion it's already far from ideal, so a little worse doesn't hurt that much.

To really streamline things, I think we could provide a Go-inspired `make` function like this:

```python
from typing import TypeVar
from pyteal import *

T = TypeVar("T", bound=abi.BaseType)

def make(t: type[T]) -> T:
    return abi.type_spec_from_annotation(t).new_instance()

myTuple = make(abi.Tuple2[abi.Uint32, abi.Uint64])
```

barnjamin: Is this something I should implement in this PR?

pyteal/ast/abi/address.py
```diff
@@ -1,33 +1,120 @@
...
+                if len(value) == AddressLength.Bytes.value:
+                    return self.stored_value.store(Bytes(value))
+                raise TealInputError(
+                    f"Got bytes with length {len(value)}, expected {AddressLength.Bytes.value}"
+                )
+            case Expr():
+                return self.stored_value.store(value)
+            case CollectionSequence():
+                # TODO: mypy thinks its possible for the type of `value` here to be
+                # Union[Sequence[Byte], str, bytes] even though we check above?
```
barnjamin, line 101: This confuses me, is it an issue with mypy or am I just confused about how this match case should work?

michaeldiamant: @barnjamin Take my thoughts with a grain of salt - I think my understanding of the Python type system is still limited.

My guess:   I think it's a language limitation rather than a user error.

Possible workaround: 
* Capture the matched pattern as shown in https://peps.python.org/pep-0636/#capturing-matched-sub-patterns.
* I think the proposal is succinct and _seems_ to avoid introducing additional classes of error.  
* Though I'm hesitant to use the approach more broadly because the IDE (at least Pycharm) resolves `cs` as type `Any`.  Presumably the lack of type inference will be improved over time.  For the mentioned issue, I'd be comfortable applying it if you feel it's an improvement.

```
            case CollectionSequence() as cs:
                return super().set(cs)
```

pyteal/ast/abi/address.py
```diff
@@ -1,33 +1,120 @@
...
+
+        # TODO: wdyt? something like this maybe should be in utils? or just manually update the args each time?
+        # from inspect import signature
+        # from typing import get_args
+        # sig = signature(self.set)
+        # expected = ", ".join([repr(a) for a in get_args(sig.parameters['value'].annotation)])
+        # raise TealInputError(
+        #     f"Got {type(value)}, expected {expected}"
+        # )
+
```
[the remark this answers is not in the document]
michaeldiamant: How strongly do we feel about showing the expected types in the error message?

Ultimately, I'll follow the group's opinion.  Trying to gauge thinking.

* I'm wondering if it's sufficient to specify that the type is unsupported, and have the programmer review the method signature.
* I think the commented out introspection is neat.  I'd favor using it avoid a stale error message provided we think the formatting is clear for the programmer (example below).
  * My take is:  The message probably needs a bit of formatting to read as "expected one of these types A, B, or C".  Possibly splitting types on newlines because each type is complex.
    *  Given the work involved to format, I circle back to wondering if a vaguer message is sufficient.   

```
E       pyteal.TealInputError: Got <class 'int'>, expected typing.Sequence[pyteal.Byte], pyteal.StaticArray[pyteal.Byte, typing.Literal[32]], pyteal.ComputedValue[pyteal.StaticArray[pyteal.Byte, typing.Literal[32]]], ForwardRef('Address'), <class 'str'>, <class 'bytes'>, <class 'pyteal.Expr'>
```

pyteal/ast/abi/address.py
jasonpaulos: Mypy is wrong here. Eventually they should get this right, but in the meantime you should explicitly cast the variable to the right type (i.e. `cast(Sequence[Byte], value)`).

FWIW VS Code (PyLance) correctly types `value` as `Sequence[Byte]` for me.

pyteal/ast/abi/address.py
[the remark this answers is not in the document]
jasonpaulos: My current thoughts are that introspection is overkill for this purpose. Giving a generic error message with the type of the passed in value will be enough information for the user to debug what went wrong, since they can also very easily see the type annotations of the function args.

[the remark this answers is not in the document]
ahangsu: ^ inspect and signature on `self.get` seems too strong for this one.

We typically throw `TealInputError` with message about allowed input type, and that is assumed to be sufficient.

pyteal/ast/abi/address.py
```diff
@@ -1,33 +1,120 @@
...
+                    return self.stored_value.store(value.stored_value.load())
+
+                raise TealInputError(
+                    f"Got {value} with type spec {value.type_spec()}, expected AddressTypeSpec"
+                )
+            case str():
+                if len(value) == AddressLength.String.value:
+                    return self.stored_value.store(Addr(value))
+                raise TealInputError(
+                    f"Got string with length {len(value)}, expected {AddressLength.String.value}"
```
ahangsu, line 89: actually, if you noticed https://github.com/algorand/pyteal/blob/master/pyteal/ast/addr.py#L21:
the implementation https://github.com/algorand/pyteal/blob/master/pyteal/types.py#L51 checks whether the string length is *exactly* 58, wonder if this would help make implementation lighter.

ahangsu (commented):
sorry for the late review, was caught up in abi-subroutine part.

Overall looking good, just some nits

pyteal/ast/abi/address.py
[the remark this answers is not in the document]
barnjamin: 👍

ahangsu (approved):

pyteal/ast/abi/address.py
```diff
@@ -1,33 +1,104 @@
...
 from pyteal.ast.abi.array_static import StaticArray, StaticArrayTypeSpec
-from pyteal.ast.abi.uint import ByteTypeSpec
+from pyteal.ast.abi.uint import ByteTypeSpec, Byte
 from pyteal.ast.expr import Expr
 
-ADDRESS_LENGTH = 32
+
+class AddressLength(IntEnum):
+    String = 58
+    Bytes = 32
```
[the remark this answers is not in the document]
ahangsu: ugh nvm im being dumb

PR #295, opened 2022-04-22
Reword account and asset Transaction field references

michaeldiamant:
Rewords account and asset Transaction field references to _hopefully_ make the intent clearer based on feedback from @algoanne.


algoanne (approved):

algochoi (commented):
Looks good, although I am biased toward specifying that these are `foreign assets` and `foreign applications`, as per [go-algorand docs](https://developer.algorand.org/docs/get-details/transactions/transactions/?from_query=transaction#common-fields-header-and-type) in addition to these changes.

algoanne:
> 

there's definitely some inconsistency across the documentations when referring to these and it is confusing, but would probably require more concerted effort to reach consistency in all the places. This smaller change at least fixes an instance that I found particularly misleading since "application accounts" can have a distinct meaning.

PR #296, opened 2022-04-25
WIP: Sorting Imports

tzaffi:
TODO: https://pypi.org/project/flake8-import-order/

### DIFF
--- a/pyteal/ast/gitxn.py
+++ b/pyteal/ast/gitxn.py
@@ -1,11 +1,10 @@
 from typing import TYPE_CHECKING, cast, Union
 
-from pyteal.config import MAX_GROUP_SIZE
-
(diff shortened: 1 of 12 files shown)


tzaffi:
Dropping this for now after offline conversation with @michaeldiamant. Here's the summary of our discussion:

___________
There's nothing essential about sorting imports. [PEP8](https://peps.python.org/pep-0008/#imports) does mention:

> Imports should be grouped in the following order:
> * Standard library imports.
> * Related third party imports.
> * Local application/library specific imports.
> You should put a blank line between each group of imports.

So the grouping of imports is referred to (but not ordering within grouping). It makes it a little easier to know where to put imports and where to find them. I initially thought that perhaps this is a good way to catch dupes, but it turns out that flake8's F811 already does that for us.

PR #297, opened 2022-04-26
Blackbox Decorator

tzaffi:
This child PR to #249 implements @michaeldiamant 's [suggestion](https://github.com/algorand/pyteal/pull/249#discussion_r858198163) for a new `@Blackbox` decorator.


jasonpaulos (approved):
I have minor comments but I'll save those for the parent PR. On the whole this looks good

michaeldiamant (approved):
@tzaffi Thanks for trying the approach - it looks like what I had in mind.  Provided folks are on-board, then let's roll with it.

PR #301, opened 2022-04-27
Add creator to AssetParam documentation

Curtis-D:


ahangsu (commented):
seems correct to me, by https://github.com/algorand/go-algorand/tree/master/data/transactions/logic#loading-values, I am not sure if we want to specify that it is available after v5.

jasonpaulos (approved):
This looks good, thanks!

I don't think we need to specify the minimum TEAL version on this page, but we probably should specify it in the docstring for `AssetParam.creator()` (we currently don't).

PR #299, opened 2022-04-26
Partition act-specific steps for run-integration-tests

michaeldiamant:
Partitions act-specific steps for `run-integration-tests` for several reasons:
* Reduces surface area to consider when making build changes.
* Hopefully, minimizes Workflow build durations.

PR #302, opened 2022-04-27
Make build-docset act-compatible

michaeldiamant:
Makes `build-docset` job act-compatible to address https://github.com/algorand/pyteal/pull/299#issuecomment-1110296347.

Provides compatibility by marking the _upload-artifact_ step as Github-only.  Act does not have access to Github's storage system.

Additionally, fixes the following warning observed during `build-docset`:
```
| /Users/michael/dev/pyteal/docs/opup.rst:4: WARNING: Title underline too short.
|
| OpUp:  Budget Increase Utility
| ========================
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -85,12 +85,8 @@ jobs:
         run: make setup-docs
       - name: Make docs
         run: make bundle-docs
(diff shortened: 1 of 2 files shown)


tzaffi (approved):
Thanks for these great fixes!

PR #307, opened 2022-04-28
ECDSA verify, decompress, and recover support

algoidurovic:
Adding support to PyTeal for ECDSA opcodes: ecdsa_verify, ecdsa_pk_decompress, and ecdsa_pk_recover. The MultiValue expression was used to handle the public key returned by ecdsa_pk_decompress and ecdsa_pk_recover, which is represented by two bytes values.

Tested: Unit tests and basic manual test in sandbox. I'll attempt to use the blackbox tested before merging this PR.


pyteal/ast/ecdsa_test.py
```diff
@@ -0,0 +1,197 @@
...
+
+    actual, _ = expr.__teal__(teal5Options)
+    actual.addIncoming()
+    actual = pt.TealBlock.NormalizeBlocks(actual)
+
+    with pt.TealComponent.Context.ignoreExprEquality():
+        assert actual == expected
+
+
+def test_ecdsa_invalid():
```
algochoi, line 170: I'd also try compiling with TEAL 4 to check that it raises a `TealInputError` as a sanity check.

I'm also personally curious what would happen if the first argument was something other than the `EcdsaCurve` enum, e.g. what happens if we pass in `2`?

pyteal/ast/ecdsa.py
```diff
@@ -0,0 +1,153 @@
...
+
+if TYPE_CHECKING:
+    from pyteal.compiler import CompileOptions
+
+
+EcdsaPubkey = [TealType.bytes, TealType.bytes]
+
+
+class EcdsaCurve(Enum):
+    Secp256k1 = 0
```
algochoi, line 17: personal nit: maybe add a comment here (might sound redundant but could be something like: Enum representing the ECDSA curve used for the public key)

algochoi (commented):
Implementation looks good and clean to me 👍  

Unfortunately, I'm not the most knowledgeable in this space so another review would probably be safe.

pyteal/ast/ecdsa.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def __teal__(self, options: "CompileOptions"):
+        verifyTealVersion(
+            self.op.min_version,
+            options.version,
+            "TEAL version too low to use op {}".format(self.op),
+        )
+
+        return TealBlock.FromOp(
+            options, TealOp(self, self.op, cast(int, self.curve)), *self.args
```
jasonpaulos, line 56: I have doubts about this curve conversion, since Python enums behave a bit strangely. I'd recommend making the enum inherit from `IntEnum`, then you can do `int(self.curve)` here

jasonpaulos (commented):
I have concerns about the enum, but otherwise this looks good.

Do you plan to add docs in this PR? They would fit in really well in this page: https://pyteal.readthedocs.io/en/stable/crypto.html

pyteal/ast/ecdsa.py
algoidurovic: You're right, my original enum broke during assemble. This wasn't caught by the tests because I was comparing the compilation result prior to assembly. I'll expand the tests as well as correct this problem.

pyteal/ast/ecdsa.py
```diff
@@ -0,0 +1,169 @@
...
+    two Byte expressions, ex. the value returned by EcdsaDecompress or EcdsaRecover. All byte
+    arguments must be big endian encoded.
+
+    Args:
+        curve: Enum representing the ECDSA curve used for the signature and public key
+        data: Hash value of the signed data. Must be 32 bytes long.
+        sigA: First component of the signature. Must evaluate to bytes.
+        sigB: Second component of the signature. Must evaluate to bytes.
+        pubkey: Public key used to verify signature. Represented as either a tuple of expressions
+            that must evaluate to bytes of as a MultiValue expression the returns two byte values.
```
michaeldiamant, line 94: @algoidurovic Minor:  I _think_ there's a typo here. I left a _suggestion_ attempting to infer the intended phrasing.

```suggestion
            that must evaluate to bytes or as a MultiValue expression that returns two byte values.
```

algoidurovic: good catch.

docs/crypto.rst
```diff
@@ -11,18 +11,22 @@ Simple TEAL opcodes have cost `1`, and more advanced cryptographic operations ha
 cost. Below is how you express cryptographic primitives in PyTeal:
```
michaeldiamant, line 11: @algoidurovic Minor:  The paragraph that's out of view from the PR presents stale info because it references how many elements appear in the table.

Open to customization - I suggest simplifying to something like:

> Algorand Smart Contracts support these cryptographic primitives.  The table below discusses usage and presents opcode cost.

michaeldiamant: @algoidurovic Thanks for the change in https://github.com/algorand/pyteal/pull/307/commits/59d811948250975c7e26f6d077900de2c500900e.  

I think in a future iteration we can make a broader pass through `crypto.rst`.  For the PR, it looks reasonable to me - resolving.

michaeldiamant (approved):

pyteal/ast/ecdsa_test.py
```diff
@@ -1,3 +1,4 @@
+from pyteal.ir.ops import Mode
```
jasonpaulos, line 1: Should be accessible as `pt.Mode` already

jasonpaulos (commented):
This looks good, I just have some minor issues with imports.

It's nice to see the integration tests as well!

docs/crypto.rst
```diff
@@ -3,26 +3,29 @@
...
-:code:`Sha512_256(e)`           `45`      `SHA-512/256` hash function, produces 32 bytes
-:code:`Ed25519Verify(d, s, p)`  `1900`\*  `1` if :code:`s` is the signature of :code:`d` signed by private key :code:`p`, else `0`
-=============================== ========= ========================================================================================
+==================================== ========= ==================================================================================================================
+Operator                             Cost      Description
+==================================== ========= ==================================================================================================================
+:code:`Sha256(e)`                    `35`      `SHA-256` hash function, produces 32 bytes
+:code:`Keccak256(e)`                 `130`     `Keccak-256` hash funciton, produces 32 bytes
+:code:`Sha512_256(e)`                `45`      `SHA-512/256` hash function, produces 32 bytes
+:code:`Ed25519Verify(d, s, p)`       `1900`\*  `1` if :code:`s` is the signature of :code:`d` signed by private key :code:`p`, else `0`
```
jasonpaulos, line 19: Take note of the changes from #324 when merging master into here

pyteal/ast/ecdsa.py
```diff
@@ -0,0 +1,169 @@
...
+
+class EcdsaCurve(Enum):
+    """Enum representing an elliptic curve specification used in ECDSA."""
+
+    Secp256k1 = (0, "Secp256k1", 5)
+
+    def __init__(self, id: int, name: str, min_version: int) -> None:
+        self.id = id
+        self.arg_name = name
+        self.min_version = min_version
```
ahangsu, line 24: nit: is there anywhere in the code checking the curve's min-version, or it is equal to `Ecdsa*` bundle of 3 opcodes' minimum version?

ahangsu (commented):
just one question

pyteal/ast/ecdsa.py
algoidurovic: Ah good catch.  At the moment, the latter is true but it may not be the case in the future if Secp256r1 is supported.

pyteal/ast/ecdsa_test.py
algoidurovic: good point, I really should stop relying on the ide to handle imports for me.

tests/integration/ecdsa_test.py
```diff
@@ -0,0 +1,169 @@
+from graviton.blackbox import DryRunExecutor
+from pyteal.ast.bytes import Bytes
+from pyteal.ast.ecdsa import EcdsaCurve, EcdsaDecompress, EcdsaRecover, EcdsaVerify
+from pyteal.ast.int import Int
+from pyteal.ast.naryexpr import And
+from pyteal.ast.subroutine import Subroutine
+from pyteal.ast.unaryexpr import Sha512_256
+from pyteal.compiler.compiler import compileTeal
+from pyteal.ir.ops import Mode
+from pyteal.types import TealType
```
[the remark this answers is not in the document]
algoidurovic: should be okay now.

pyteal/ast/ecdsa.py
ahangsu: (I think it is also the case when you use decompress and recover)

jasonpaulos (approved):

ahangsu (approved):

PR #308, opened 2022-04-28
Bundle optional refactorings to subroutine.py

michaeldiamant:
Optionally bundles the following refactorings to #256:
* Rename `is_abi_annotation` to `_is_abi_annotation` because its intent appears to be private.
* Refactor `from_dict` to pattern match statements.  The change is minor though subjectively, I feel it's easier to interpret.
  * I confirmed  `kwarg_info` isn't supplied as `None`.  I presume it's _safe_ to remove the null check because we don't elsewhere rigorously check for nulls when parameter isn't optional.
  * Refactor `_arg_types_and_by_refs` to remove `continue` statement usage.  Subjectively, I find it difficult to follow jumps with `continue` statements.

Feel welcomed to tell me that you'd prefer to keep as is and I'll happily close the PR.  If some of the changes are desirable, feel welcomed to ask me to itemize discrete changesets.

### DIFF
--- a/pyteal/ast/subroutine.py
+++ b/pyteal/ast/subroutine.py
@@ -193,13 +193,14 @@ class SubroutineDefinition:
                         f"Function keyword parameter {name} has type {expected_arg_type}"
                     )
                 abi_output_kwarg[name] = expected_arg_type
-            else:
-                expected_arg_types.append(expected_arg_type)
+                continue
 
-                if expected_arg_type is ScratchVar:
-                    by_ref_args.add(name)
-                if isinstance(expected_arg_type, abi.TypeSpec):
-                    abi_args[name] = expected_arg_type
+            expected_arg_types.append(expected_arg_type)
+
+            if expected_arg_type is ScratchVar:
+                by_ref_args.add(name)
+            if isinstance(expected_arg_type, abi.TypeSpec):
+                abi_args[name] = expected_arg_type
 
         return expected_arg_types, by_ref_args, abi_args, abi_output_kwarg
 
--- a/pyteal/ast/subroutine.py
+++ b/pyteal/ast/subroutine.py
@@ -79,7 +79,7 @@ class SubroutineDefinition:
         self.__name = self.implementation.__name__ if name_str is None else name_str
 
     @staticmethod
-    def is_abi_annotation(obj: Any) -> bool:
+    def _is_abi_annotation(obj: Any) -> bool:
         try:
             abi.type_spec_from_annotation(obj)
             return True
@@ -105,14 +105,16 @@ class SubroutineDefinition:
             # * `invoke` type checks provided arguments against parameter types to catch mismatches.
             return Expr
         else:
-            if not isclass(ptype) and not SubroutineDefinition.is_abi_annotation(ptype):
+            if not isclass(ptype) and not SubroutineDefinition._is_abi_annotation(
+                ptype
+            ):
                 raise TealInputError(
                     f"Function has parameter {parameter_name} of declared type {ptype} which is not a class"
                 )
 
             if ptype in (Expr, ScratchVar):
                 return ptype
-            elif SubroutineDefinition.is_abi_annotation(ptype):
+            elif SubroutineDefinition._is_abi_annotation(ptype):
                 return abi.type_spec_from_annotation(ptype)
             else:
                 raise TealInputError(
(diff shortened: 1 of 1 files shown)


pyteal/ast/subroutine.py
```diff
@@ -191,14 +193,13 @@ def _arg_types_and_by_refs(
...
                 expected_arg_types.append(expected_arg_type)
 
-            if expected_arg_type is ScratchVar:
-                by_ref_args.add(name)
-            if isinstance(expected_arg_type, abi.TypeSpec):
-                abi_args[name] = expected_arg_type
+                if expected_arg_type is ScratchVar:
+                    by_ref_args.add(name)
+                if isinstance(expected_arg_type, abi.TypeSpec):
+                    abi_args[name] = expected_arg_type
```
ahangsu, line 202: we can still keep the continue, and remove the `else` to lessen the indentation, like:
```python
# some code before ...
                abi_output_kwarg[name] = expected_arg_type
                continue

            expected_arg_types.append(expected_arg_type)

            if expected_arg_type is ScratchVar:
                by_ref_args.add(name)
            if isinstance(expected_arg_type, abi.TypeSpec):
                abi_args[name] = expected_arg_type
# some code after ...
```

pyteal/ast/subroutine.py
```diff
@@ -299,15 +300,15 @@ class _OutputKwArgInfo:
...
-        if kwarg_info is None or len(kwarg_info) == 0:
-            return None
-        elif len(kwarg_info) == 1:
-            key = list(kwarg_info.keys())[0]
-            return _OutputKwArgInfo(key, kwarg_info[key])
-        else:
-            raise TealInputError(
-                f"illegal conversion kwarg_info length {len(kwarg_info)}."
-            )
+        match list(kwarg_info.keys()):
```
ahangsu, line 303: (probably) need to additionally handle `None` case, I need to remember if I really want to support that case...

tzaffi, line 303: wow! match in action. so exciting!!!

tzaffi: but isn't this going to blow up in the case that `kwarg_info is None` ?

pyteal/ast/subroutine.py
michaeldiamant: @ahangsu Gauging your preference:  Are you saying you _prefer_ the `continue` statement?

* If yes, then I'll change to your preference because I'm presenting the PR optionally.
* Context from my side: 
  * I specifically removed it because I think it complicates comprehension (as noted in PR description).  
  * The indentation did _not_ strike me as a concern.  I can imagine a broader change that removes indentation, but I'm trying to present minimal changes.  Particularly for a refactoring.

pyteal/ast/subroutine.py
michaeldiamant: @ahangsu Let me know if you feel the `None` case needs to be handled.  As mentioned in PR description, I traced code paths and convinced myself the value won't be `None`.

However, I consider my Python interpretation abilities to be mediocre at best.  So, if you feel the `None` check is needed, I'll add it.

michaeldiamant: Hopefully the 1st `match` of many. 😅 

> isn't this going to blow up in the case that kwarg_info is None ?

I responded in https://github.com/algorand/pyteal/pull/308#discussion_r861360073 because the question arose in multiple places.

ahangsu: okay, I took another pass, and all the use-cases for `from_dict` is on `Subroutine`'s output kwarg. It is never None, so I am happy to merge as for this one.

pyteal/ast/subroutine.py
ahangsu: I am slightly preferring `continue` statement, but they should be having the same, so I have no too much comments on that, and feel free to ignore my previous thoughts.

ahangsu (approved):

pyteal/ast/subroutine.py
michaeldiamant: @ahangsu Sounds like a plan.  Thanks for confirming - Resolving.

pyteal/ast/subroutine.py
michaeldiamant: @ahangsu I appreciate the flexibility though I'm only looking to merge if you prefer the changes.  Via https://github.com/algorand/pyteal/pull/308/commits/134e10631eb00d4f40b3b6ab780b1adc356b41cc, I switched to the suggestion you raised earlier.

If it looks good to you, let's resolve the thread.

ahangsu: lgtm, im resolving!

ahangsu (approved):

PR #311, opened 2022-04-29
Remove potentially extraneous SubroutineCall parameter

michaeldiamant:
Removes a potentially extraneous SubroutineCall parameter (`output_kwarg`) from #256.

Treat the PR optionally.  I explored the change while trying to understand `output_kwarg` flow. 

Since the value is _only_ referenced in `__str__`, it's unclear to me if either:
* (a) The value _was_ used elsewhere in an earlier iteration and is now inadvertently left behind.
* (b) The value is intentionally _only_ used in `__str__` as a way to provide more debugging context or another reason I missed.
* (c) The value _should_ be used elsewhere.  And implies a test is missing because the build passes.

I convinced myself it's _not_ (c).  If it's (b), then I'll happily close the PR.

### DIFF
--- a/pyteal/ast/subroutine.py
+++ b/pyteal/ast/subroutine.py
@@ -253,9 +253,7 @@ class SubroutineDefinition:
                         f"should have ABI typespec {arg_type} but got {arg.type_spec()}"
                     )
 
-        return SubroutineCall(
-            self, args, output_kwarg=_OutputKwArgInfo.from_dict(self.output_kwarg)
-        )
+        return SubroutineCall(self, args)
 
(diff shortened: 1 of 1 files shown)


ahangsu (commented):
Such design is for the sake of passing option (b) in, which may help in telling the difference between:
- normal subroutine returning `Expr`
- subroutine returning `ABI` values

I agree that the design is not ideal, but we have to pass the info about return in `SubroutineCall` somewhere.

michaeldiamant:
@ahangsu Thanks for the context - Closing the PR.

PR #310, opened 2022-04-29
Feature/abi merge master again

tzaffi:
This merges master back into `feature/abi`. The only major conflicts were in the in `subroutine.py::SubroutineDefinition`. Validation was conducted in two different ways:
* `feature/abi` - used `_arg_types_and_by_refs`. This has now been removed in favor of:
* `master` - `_validate()`

Admittedly, this is not a fun PR to review in great detail, but something like this was inevitable given all the changes happening on the main branch.

PR #312, opened 2022-04-29
Reorder _validate_parameter_type to avoid isclass check

michaeldiamant:
Reorders checks in `Reorder _validate_parameter_type` to avoid `isclass` check.

Rationale:  Since there was some discussion on the topic in https://github.com/algorand/pyteal/pull/256#discussion_r861330174, I paired with @tzaffi to better understand the runtime type system checks.  While pairing, we observed that it's possible to remove a check and retain equivalent functionality _provided_ we're _OK_ with a different granularity error message.

Treat as optional.  I'll happily close if you'd prefer to keep as is.

### DIFF
--- a/pyteal/ast/subroutine.py
+++ b/pyteal/ast/subroutine.py
@@ -1,6 +1,6 @@
 from collections import OrderedDict
 from dataclasses import dataclass
-from inspect import isclass, Parameter, signature, Signature
+from inspect import Parameter, signature, Signature
 from typing import (
     Callable,
     Optional,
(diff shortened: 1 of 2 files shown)


pyteal/ast/subroutine.py
```diff
@@ -105,13 +105,6 @@ def _validate_parameter_type(
...
             return Expr
         else:
-            if not isclass(ptype) and not SubroutineDefinition._is_abi_annotation(
-                ptype
-            ):
-                raise TealInputError(
-                    f"Function has parameter {parameter_name} of declared type {ptype} which is not a class"
-                )
-
             if ptype in (Expr, ScratchVar):
```
tzaffi, line 108: FYI - [my merges ](https://github.com/algorand/pyteal/pull/310/files) are about to repeat this work.

And optionally: Since you are returning out of the if's you can remove the else's and write the last 2 conditions as:

```python
            if SubroutineDefinition._is_abi_annotation(ptype):
                return abi.type_spec_from_annotation(ptype)
            raise TealInputError(
                f"Function has parameter {parameter_name} of disallowed type {ptype}. "
                f"Only the types {(Expr, ScratchVar, 'ABI')} are allowed"
            )
```

michaeldiamant:
Closing - https://github.com/algorand/pyteal/pull/310 brings the same change in.

PR #314, opened 2022-05-02
no utils pls

tzaffi:
Removing a file that shouldn't be in `master` but which got merged in by mistake in #249


ahangsu (commented):
need to update build script

tzaffi:
> need to update build script

👍

PR #317, opened 2022-05-02
v0.12.1b2: back to requirements.txt

tzaffi:
# Fix the problems that prevented `v0.12.0` from getting to `PyPi` 

* Adding `requirements.txt` for development builds (because of the graviton [dependency as a non-PyPi resource](https://stackoverflow.com/questions/54887301/how-can-i-use-git-repos-as-dependencies-for-my-pypi-package/54894359#54894359))

## Also making publication depend on all tests passing:

image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -95,7 +95,7 @@ jobs:
   upload-to-pypi:
     runs-on: ubuntu-20.04
     container: python:3.10
-    needs: ['build-test']
+    needs: ['build-test', 'run-integration-tests', 'build-docset']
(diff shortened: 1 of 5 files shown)


jasonpaulos (approved):
As a nit, the README should also be updated since it references `pip install -e .[development]`, but that can always happen after we've confirmed this works

michaeldiamant:
@tzaffi and @jasonpaulos Thanks for the effort here!  The PR fixed the PyPi upload issue:  https://github.com/algorand/pyteal/actions/runs/2264000664.

PR #326, opened 2022-05-04
Add convenience `make` method for ABI types

jasonpaulos:
As described in https://github.com/algorand/pyteal/pull/289#discussion_r862178354, add a convenience method to instantiate ABI types.

Example usage:

```python
myTuple = abi.make(abi.Tuple2[abi.Uint64, abi.DynamicArray[abi.Bool]])
```

The above produces the same as:
```python
myTuple = abi.Tuple(abi.TupleTypeSpec(abi.Uint64TypeSpec(), abi.DynamicArrayTypeSpec(abi.BoolTypeSpec())))
```

### DIFF
--- a/pyteal/ast/abi/__init__.py
+++ b/pyteal/ast/abi/__init__.py
@@ -35,8 +35,8 @@ from pyteal.ast.abi.array_base import ArrayTypeSpec, Array, ArrayElement
 from pyteal.ast.abi.array_static import StaticArrayTypeSpec, StaticArray
 from pyteal.ast.abi.array_dynamic import DynamicArrayTypeSpec, DynamicArray
 
-from pyteal.ast.abi.util import type_spec_from_annotation, make
 from pyteal.ast.abi.method_return import MethodReturn
+from pyteal.ast.abi.util import type_spec_from_annotation, make
 
 __all__ = [
     "String",
@@ -77,7 +77,7 @@ __all__ = [
     "StaticArray",
     "DynamicArrayTypeSpec",
     "DynamicArray",
+    "MethodReturn",
     "type_spec_from_annotation",
     "make",
-    "MethodReturn",
 ]
--- a/pyteal/ast/abi/__init__.py
+++ b/pyteal/ast/abi/__init__.py
@@ -34,7 +34,8 @@ from pyteal.ast.abi.tuple import (
 from pyteal.ast.abi.array_base import ArrayTypeSpec, Array, ArrayElement
 from pyteal.ast.abi.array_static import StaticArrayTypeSpec, StaticArray
 from pyteal.ast.abi.array_dynamic import DynamicArrayTypeSpec, DynamicArray
-from pyteal.ast.abi.util import type_spec_from_annotation
+
+from pyteal.ast.abi.util import type_spec_from_annotation, make
 from pyteal.ast.abi.method_return import MethodReturn
 
 __all__ = [
@@ -77,5 +78,6 @@ __all__ = [
     "DynamicArrayTypeSpec",
     "DynamicArray",
     "type_spec_from_annotation",
+    "make",
     "MethodReturn",
 ]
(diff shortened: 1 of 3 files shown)


ahangsu (approved):
THANK YOU!!

pyteal/ast/abi/__init__.py
```diff
@@ -34,7 +34,8 @@
 from pyteal.ast.abi.array_base import ArrayTypeSpec, Array, ArrayElement
 from pyteal.ast.abi.array_static import StaticArrayTypeSpec, StaticArray
 from pyteal.ast.abi.array_dynamic import DynamicArrayTypeSpec, DynamicArray
-from pyteal.ast.abi.util import type_spec_from_annotation
+
+from pyteal.ast.abi.util import type_spec_from_annotation, make
 from pyteal.ast.abi.method_return import MethodReturn
```
tzaffi, line 39:
```suggestion
from pyteal.ast.abi.method_return import MethodReturn
from pyteal.ast.abi.util import type_spec_from_annotation, make
```
nit: suggest alphabetize

pyteal/ast/abi/util.py
```diff
@@ -205,3 +205,30 @@ def type_spec_from_annotation(annotation: Any) -> TypeSpec:
...
+
+    This is purely a convenience method over instantiating the type directly, which can be cumbersome
+    due to the lengthy TypeSpec class names.
+
+    Args:
+        t: A fully-specified annotation of a subclass of abi.BaseType, meaning all generic parameters
+            must be given values.
+
+    Returns:
+        A new instance of the given type class.
```
tzaffi, line 232: While this is technically correct, the word `given` to me sounds like the input (and we're given an annotation not a type class). So maybe slightly clearer would be:

```
    Returns:
        A new instance of the corresponding type class.
```

pyteal/ast/abi/util.py
```diff
@@ -205,3 +205,30 @@ def type_spec_from_annotation(annotation: Any) -> TypeSpec:
...
+def make(t: type[T]) -> T:
+    """Create a new instance of an ABI type. The type to create is determined by the input argument,
+    which must be a fully-specified annotation of type's class.
+
+    For example:
+        .. code-block:: python
+
+                # both of these are equivalent
+                myTuple = abi.make(abi.Tuple2[abi.Uint64, abi.DynamicArray[abi.Bool]])
+                myTuple = abi.Tuple(abi.TupleTypeSpec(abi.Uint64TypeSpec(), abi.DynamicArrayTypeSpec(abi.BoolTypeSpec())))
```
tzaffi, line 222: totally optional...

Since this is expected to end up in the docs, I recommend adding this as a test case to `user_guide_test.py`. 

EG you could add something like "abi_make_snippet" to our [COPACETIC_SNIPPETS](https://github.com/algorand/pyteal/blob/master/tests/unit/user_guide_test.py#L63) with an assertions such as `assert myTuple_make == myTuple_explicit`

tzaffi (approved):
LGTM

I have some totally optional suggestions. As these shouldn't hold up the merge I am approving right away.

pyteal/ast/abi/util.py
jasonpaulos: I opted to drop the use of annotation elsewhere, since an annotation is a class and both terms added confusion.

pyteal/ast/abi/util.py
jasonpaulos: I've opted not to do this for two reasons:
1. Existing user guide tests seem to be primarily concerned with comparing compiled TEAL code, and `make` does not affect this directly
2. `test_make` in `pyteal/ast/abi/util_test.py` already tests something very similar to the example. To fully close the loop, I updated the example to use the same expression as the test.

tzaffi (approved):
LGTM again

PR #328, opened 2022-05-06
Use deferred subroutine expression for ABI returns

jasonpaulos:
Inspired by Go's `defer` feature, this PR adds the ability for subroutines to have a deferred expression that will be inserted by the compiler before every instance of `retsub` in that subroutine.

Note that it's only added before `retsub` ops, not program ending ops like `return` or `err`. Why not? Because we cannot insert code before _every_ program ending operation, for two reasons:
* Some ops will fail the program if they are given invalid inputs, e.g. adding two ints that overflow. We can't know ahead of time what will fail and insert code before them.
* The subroutine can call another subroutine which (either conditionally or always) invokes a `return` op. Our subroutine has no visibility into this, so we can't insert code before it.

This feature is only used to implement ABI return value subroutines and is not directly exposed to users.

### DIFF
--- a/pyteal/ast/subroutine.py
+++ b/pyteal/ast/subroutine.py
@@ -11,7 +11,6 @@ from typing import (
 
 from pyteal.ast import abi
 from pyteal.ast.expr import Expr
-from pyteal.ast.return_ import Return
 from pyteal.ast.seq import Seq
 from pyteal.ast.scratchvar import DynamicScratchVar, ScratchVar
 from pyteal.errors import TealInputError, verifyTealVersion
--- a/pyteal/ast/subroutine.py
+++ b/pyteal/ast/subroutine.py
@@ -731,10 +731,6 @@ def evaluate_subroutine(subroutine: SubroutineDefinition) -> SubroutineDeclarati
 
     # if there is an output keyword argument for ABI, place the storing on the stack
     if output_carrying_abi:
-        if subroutine_body.has_return():
-            raise TealInputError(
-                "ABI returning subroutine definition should have no return"
-            )
         if subroutine_body.type_of() != TealType.none:
             raise TealInputError(
                 f"ABI returning subroutine definition should evaluate to TealType.none, "
--- a/pyteal/ast/subroutine.py
+++ b/pyteal/ast/subroutine.py
@@ -302,10 +302,16 @@ SubroutineDefinition.__module__ = "pyteal"
 
 
 class SubroutineDeclaration(Expr):
-    def __init__(self, subroutine: SubroutineDefinition, body: Expr) -> None:
+    def __init__(
+        self,
+        subroutine: SubroutineDefinition,
+        body: Expr,
+        deferred_expr: Optional[Expr] = None,
+    ) -> None:
         super().__init__()
         self.subroutine = subroutine
         self.body = body
+        self.deferred_expr = deferred_expr
 
     def __teal__(self, options: "CompileOptions"):
         return self.body.__teal__(options)
@@ -720,6 +726,9 @@ def evaluate_subroutine(subroutine: SubroutineDefinition) -> SubroutineDeclarati
         raise TealInputError(
             f"Subroutine function does not return a PyTeal expression. Got type {type(subroutine_body)}."
         )
+
+    deferred_expr: Optional[Expr] = None
+
     # if there is an output keyword argument for ABI, place the storing on the stack
     if output_carrying_abi:
         if subroutine_body.has_return():
@@ -731,13 +740,11 @@ def evaluate_subroutine(subroutine: SubroutineDefinition) -> SubroutineDeclarati
                 f"ABI returning subroutine definition should evaluate to TealType.none, "
                 f"while evaluate to {subroutine_body.type_of()}."
             )
-        subroutine_body = Seq(
-            subroutine_body, Return(output_carrying_abi.stored_value.load())
-        )
+        deferred_expr = output_carrying_abi.stored_value.load()
 
     # Arg usage "A" to be pick up and store in scratch parameters that have been placed on the stack
     # need to reverse order of argumentVars because the last argument will be on top of the stack
     body_ops = [var.slot.store() for var in arg_vars[::-1]]
     body_ops.append(subroutine_body)
 
-    return SubroutineDeclaration(subroutine, Seq(body_ops))
+    return SubroutineDeclaration(subroutine, Seq(body_ops), deferred_expr)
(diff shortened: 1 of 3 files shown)


pyteal/compiler/compiler.py
```diff
@@ -133,26 +133,62 @@ def compileSubroutine(
...
+            deferred_start.incoming = block.incoming
+            block.incoming = [deferred_end]
+            deferred_end.nextBlock = block
+
+            for prev in deferred_start.incoming:
+                prev.replaceOutgoing(block, deferred_start)
+
+            if block is start:
+                # this is the start block, replace start
+                start = deferred_start
```
jasonpaulos, line 177: For reviewers studying this code, it's very similar to the implementation of `TealBlock.NormalizeBlocks`: https://github.com/algorand/pyteal/blob/5596a71c5de238a33167cc5cb940f7397225d0d9/pyteal/ir/tealblock.py#L187

pyteal/compiler/compiler.py
```diff
@@ -133,26 +133,62 @@ def compileSubroutine(
...
```
ahangsu, line 177: as far as I understood, this part looks like you are trying to insert the `deferred` expression before a single block of `retsub`, right?

pyteal/compiler/compiler_test.py
```diff
@@ -1627,6 +1627,109 @@ def storeValue(key: pt.Expr, t1: pt.Expr, t2: pt.Expr, t3: pt.Expr) -> pt.Expr:
...
+return
+deferredExample_0_l6:
+int 1
+return
+deferredExample_0_l7:
+byte "deferred"
+retsub
+deferredExample_0_l8:
+byte "deferred"
+retsub
```
ahangsu, line 1725: as far as I understand, this looks like:
- `== 0` case with deferred expression
- `!= 0, 1, 2, 3` case, and normally return subroutine with deferred expression, and return.

ahangsu (commented):
looking good to me, it seems for abi-returning-subroutine, we can just call `Return()` anywhere to return from subroutine, with no argument of `output.get()` or expressions alike.

Statically, this would align with previous design of subroutine evaluate to `TealType.none`, while at compile time, compiler would insert the returning ABI scratch-var expression into AST.

Is there anything on your mind for more complete implementation?

pyteal/compiler/compiler_test.py
jasonpaulos: It's making sure that the deferred expression is only added before `retsub`, not `return` or any other opcode.

ahangsu: right, it only applies to `retsub` not `return`. looks correct. resolving.

pyteal/compiler/compiler.py
jasonpaulos: Yes, if the first block contains `retsub`, we need to update the `start` variable as well, otherwise it will skip the new block we inserted before it

ahangsu: looking right, resolving.

pyteal/compiler/compiler.py
```diff
@@ -133,26 +133,64 @@ def compileSubroutine(
...
+            if len(block.ops) != 1:
+                # we expect all retsub ops to be in their own block at this point since
+                # TealBlock.NormalizeBlocks has not yet been used
+                raise TealInternalError(
+                    f"Expected retsub to be the only op in the block, but there are {len(block.ops)} ops"
+                )
 
-    order = sortBlocks(start, end)
-    teal = flattenBlocks(order)
+            deferred_start, deferred_end = get_deferred_blocks()
```
ahangsu, line 165: nit: is it possible to move line 152 `__teal__` to this line?

ahangsu (approved):
lgtm, just one nit

jasonpaulos:
> looking good to me, it seems for abi-returning-subroutine, we can just call `Return()` anywhere to return from subroutine, with no argument of `output.get()` or expressions alike.

Correct, I think this is more aligned with the subroutine return model we developed.

> Statically, this would align with previous design of subroutine evaluate to `TealType.none`, while at compile time, compiler would insert the returning ABI scratch-var expression into AST.

👍 

> Is there anything on your mind for more complete implementation?

No, I think this is good enough as it is

pyteal/compiler/compiler.py
jasonpaulos: Good idea, the function approach made sense initially but I think we can get rid of it now

pyteal/ast/subroutine.py
```diff
@@ -720,24 +725,21 @@ def var_n_loaded(
         raise TealInputError(
             f"Subroutine function does not return a PyTeal expression. Got type {type(subroutine_body)}."
         )
+
+    deferred_expr: Optional[Expr] = None
+
     # if there is an output keyword argument for ABI, place the storing on the stack
     if output_carrying_abi:
-        if subroutine_body.has_return():
-            raise TealInputError(
```
tzaffi, line 726: Are there any downstream implications for removing this check? I removed the check in my own branch and saw that no unit tests failed. This either indicates that this was a dead code, or that we didn't have coverage.

pyteal/compiler/compiler.py
```diff
@@ -133,26 +133,60 @@ def compileSubroutine(
...
+
+        for block in TealBlock.Iterate(start):
+            if not any(op.getOp() == Op.retsub for op in block.ops):
+                continue
+
+            if len(block.ops) != 1:
+                # we expect all retsub ops to be in their own block at this point since
+                # TealBlock.NormalizeBlocks has not yet been used
+                raise TealInternalError(
+                    f"Expected retsub to be the only op in the block, but there are {len(block.ops)} ops"
```
tzaffi, line 156: I recommend adding coverage for this error in a unit test.

tzaffi (commented):
Recommend adding a unit test, and question about a removed check.

pyteal/ast/subroutine.py
tonyfloatersu: Here's the story, referring back to this discussion: https://github.com/algorand/pyteal/pull/256/files#r865461530, where Jason pointed out that `has_return` is true *only* when every possible codepath in expr results in a return.

The code above thus does not quite work, with an example:
supposing the codeblock has an `If` expr with `then` branch has return, and `else` branch no return, then `has_return` is not returning true.

- So is this a deadcode, I don't think so.
- Is it wrong, yea it is a misuse of `has_return`.
- Is there coverage not enough, I think that's also right, but the restriction of no `return` expr in subroutine codeblock is lifted, so the coverage does not matter here.

What we need to do in abi-subroutine PR is to introduce subroutine with conditional return, as Jason suggested.

jasonpaulos: Resolving because this should be addressed in #256

tzaffi (approved):
LGTM

PR #329, opened 2022-05-06
Fix `If` type bug

jasonpaulos:
PyTeal `If` statements have the following rules:
1. If a then branch (what executes when the condition is true) and an else branch (what executes when the condition is false) exist, then both must return the same `TealType`.
2. If a then branch exists and an else branch does not, then the then branch must return `TealType.none`, since in this case the else branch is essentially `Seq()` (a noop expression that returns nothing).

These rules are properly enforced for the original `If` syntax (`If(<condition>, <then>, <else>)`), but the new syntax (`If(<condition>).Then(<then>).Else(<else>)`) unfortunately does not enforce these rules. This PR fixes that.

### DIFF
--- a/pyteal/ast/if_.py
+++ b/pyteal/ast/if_.py
@@ -1,6 +1,9 @@
 from typing import TYPE_CHECKING
 
-from pyteal.errors import TealCompileError, TealInputError
+from pyteal.errors import (
+    TealCompileError,
+    TealInputError,
+)
 from pyteal.types import TealType, require_type
 from pyteal.ir import TealSimpleBlock, TealConditionalBlock
 from pyteal.ast.expr import Expr
@@ -78,6 +81,11 @@ class If(Expr):
     def type_of(self):
         if self.thenBranch is None:
             raise TealCompileError("If expression must have a thenBranch", self)
+
+        if self.elseBranch is None:
+            # if there is only a thenBranch, it must evaluate to TealType.none
+            require_type(self.thenBranch, TealType.none)
+
         return self.thenBranch.type_of()
 
     def has_return(self):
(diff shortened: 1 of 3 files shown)


pyteal/ast/multi_test.py
```diff
@@ -37,7 +37,7 @@ def __test_single_conditional(
     ifExpr = (
         pt.If(expr.output_slots[1].load())
         .Then(expr.output_slots[0].load())
-        .Else(pt.Bytes("None"))
+        .Else(pt.App.globalGet(pt.Bytes("None")))
```
jasonpaulos, line 40: This unit test was previously passing despite its then and else branches returning different types for some iterations.

To fix it, I've made the else branch return `App.globalGet`, which is a `TealType.any`, so it will be correct for all then branches unless the then branch is `TealType.none`.

pyteal/ast/if_.py
```diff
@@ -78,6 +81,11 @@ def __str__(self):
     def type_of(self):
         if self.thenBranch is None:
             raise TealCompileError("If expression must have a thenBranch", self)
+
+        if self.elseBranch is None:
+            # if there is only a thenBranch, it must evaluate to TealType.none
+            require_type(self.thenBranch, TealType.none)
+
         return self.thenBranch.type_of()
```
ahangsu, line 89: ~~nit: do we need to worry about `then` and `else` evaluate to `none` and `uint64`?~~

For anyone who has the same question, look at `else` branch: type is enforced to be equal to then branch.

ahangsu: I also believe for line 48/49, we should probably call `then` and `else` to enforce the type check, or developers can write in initialization to bypass such test.

ahangsu (commented):
just one comment

pyteal/ast/if_.py
jasonpaulos: Lines 38-45 perform the type check for values given to the constructor. Did you see that, or do you have a different concern?

ahangsu: ok great! I overlooked the checking, resolving.

ahangsu (approved):

pyteal/ast/multi_test.py
tzaffi: I see. I believe @algoidurovic wrote this test originally, so I'm curious if he has any thoughts about changing this test.

PR #332, opened 2022-05-10
Bug fix for #331 `continue` skips `while` condition check

### LINKED ISSUE #331: Continue skips while loop condition check
### Subject of the issue

`Continue()` statement skips the check inside the while loop.

### Your environment

`pyteal 0.12.1`

### Steps to reproduce

Running the script below:
```python
from pyteal import *

i = ScratchVar(slotId=10)
ast = Seq(
    i.store(Int(0)),
    While(i.load() < Int(30)).Do(
        Seq(
            i.store(i.load() + Int(1)),
            Continue(),
        )
    ),
    Return(Int(1))
)

comp = compileTeal(ast, Mode.Application, version=6)

print(comp)
```

...prints the following TEAL:
```
#pragma version 6
int 0
store 10
load 10
int 30
<
bz main_l2
main_l1:
load 10
int 1
+
store 10
b main_l1
main_l2:
int 1
return
```

Note, that `main_l1:` is just after the condition check, therefore the loop never terminates (it technically does after 18446744073709551615 iterations, but that's an edge case I'm not concerned about).

### Expected behaviour

I expect the PyTEAL compiled code to actually terminate, to better portray what I mean, I have created the following script in Python:

```python
i = 0
while i < 30:
    i += 1
    continue

# control reaches here, therefore the program halts 🥳 
```

### Actual behaviour

∞ lööp

### Extra info
`Continue()` works as expected inside the `For` loop

ahangsu:
## Description

The error is caused by https://github.com/algorand/pyteal/blob/master/pyteal/ast/while_.py#L55, where `continue` block's next block is wired to the start of `do` block. Expected behavior is wire `continue`'s next to `cond`'s start.

The code incurring bug is
```python
from pyteal import *

i = ScratchVar(slotId=10)
ast = Seq(
    i.store(Int(0)),
    While(i.load() < Int(30)).Do(
        Seq(
            i.store(i.load() + Int(1)),
            Continue(),
        )
    ),
    Return(Int(1))
)

comp = compileTeal(ast, Mode.Application, version=6)

print(comp)
```

After the bug fix, the (plausibly correct to my eyes) produced code is
```
#pragma version 6
int 0
store 10
main_l1:
load 10
int 30
<
bz main_l3
load 10
int 1
+
store 10
b main_l1
main_l3:
int 1
return
```

This PR also contains some testcase change in compiler test, relating with `continue` in `while` behavior.

Closes #331

### DIFF
--- a/tests/integration/graviton_test.py
+++ b/tests/integration/graviton_test.py
@@ -13,6 +13,9 @@ from pyteal import (
     Mode,
     ScratchVar,
     Seq,
+    While,
+    Continue,
+    Return,
     Subroutine,
     SubroutineFnWrapper,
     TealType,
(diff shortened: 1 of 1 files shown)


tzaffi (commented):
Nice catch.

I think this is a great candidate for a graviton test and I'll be very happy to assist in this regard.,

jasonpaulos (approved):
This looks correct, thank you for the fix and updated tests

tzaffi (approved):
LGTM

Glad graviton could give greater confidence

PR #322, opened 2022-05-03
WIP: Blackbox abi

tzaffi:

PR #335, opened 2022-05-11
Expand BlackboxPyTealer in an attempt to simplify dryrun request generation

michaeldiamant:
Optionally proposes a change to #322 that _may_ simplify blackbox test setup.  

While reviewing #322, it seemed to me that it's common to heavily reference `BlackboxPyTealer` when constructing a Graviton dryrun request.  To simplify construction, the PR proposes moving construction into `BlackboxPyTealer`.

If the idea is worth keeping, then other `DryRunInspector` method ought to be provided.

Aside:  I originally attempted to implement the PR by exposing a partially applied function in order to minimize parameter duplication.
* However, due to my lack of Python understanding and/or available tooling,  adding a class method as shown seems to be path of least resistance.
* A disadvantage of the PR's approach is that it hides `DryRunInspector`.  I can see downsides to adding another layer, which is why I present the PR optionally.


tzaffi:
Have you looked at [functools.partial ](https://docs.python.org/3/library/functools.html#functools.partial)?

tests/blackbox.py
```diff
@@ -1,15 +1,18 @@
-from typing import Callable
+from typing import Callable, List, Sequence, Union
```
tzaffi, line 1:
```suggestion
from typing import Callable, Sequence
```

tests/blackbox.py
```diff
@@ -373,3 +376,35 @@ def approval():
             return Seq(*(preps + results))
 
         return approval
+
+    def dryrun_on_sequence(
+        self,
+        inputs: List[Sequence[Union[str, int]]],
```
tzaffi, line 382:
```suggestion
        inputs: list[Sequence[str | int]],
```

tzaffi (commented):
This is clearly an improvement, thank you! Since you're touching this code I also recommend you consider the following:
1. the minor python 3.10 typing conventions I suggest
2. I would be supportive of a clever use of `functools.partials`. I'm curious how this would look like.
3. Now that `BlackboxPyTealer` can also run dry-runs, I think the suffix `PyTealer` is somewhat misleading. Consider a better? name such as `SubroutineRunner` or `BlacboxTester` or you can probably come up with something better. (make sure to announce it on slack's #christening channel)

tests/blackbox.py
michaeldiamant: @tzaffi Addressed in https://github.com/algorand/pyteal/pull/335/commits/d799e3752e2a92f52102a06ae85f4f499419ae69.  I didn't directly take in the _suggestion_ because I had to make several changes in 1 commit.

If it looks good, can you resolve?

tests/blackbox.py
michaeldiamant: @tzaffi Addressed in https://github.com/algorand/pyteal/pull/335/commits/d799e3752e2a92f52102a06ae85f4f499419ae69.  I didn't directly take in the _suggestion_ because I had to make several changes in 1 commit.

If it looks good, can you resolve?

michaeldiamant:
As of https://github.com/algorand/pyteal/pull/335/commits/7f5f5262aa4ea3fafc769634a68836e177638dbb, I've made these changes per PR feedback:
* Rename from `BlackboxPyTealer` to `PyTealDryRunExecutor`.  I'm not strongly tied to the proposed name.  I am also OK to keep the existing name.
* Replace existing `DryRunExecutor.dryrun_on_*_sequence` invocations with `PyTealDryRunExecutor.dryrun_on_sequence`.  In a child PR, I'll provide updates for other usages in case we want to limit changes.

Note - The build issue stems from the parent PR.  I'll update the PR once the parent branch issue is resolved.

tests/blackbox.py
tzaffi: 👍

tests/blackbox.py
tzaffi: 👍

PR #338, opened 2022-05-12
Replace blackbox_pyteal with PyTealDryRunExecutor

michaeldiamant:
Replaces `blackbox_pyteal` with `PyTealDryRunExecutor` to provide more consistent usage patterns throughout pyteal. 

The PR is a follow up to #335 and is a child PR to #322.


tests/blackbox.py
```diff
@@ -192,7 +196,7 @@ def handle_arg(arg):
 
     def abi_return_type(self) -> None | algosdk.abi.ABIType:
         if not self.is_abi():
-            raise NotImplementedError("this {type(self)} is not A.B.I. compatible")
```
michaeldiamant, line 195: @tzaffi Highlighting a change I'm unsure about.

For ease of use, the PR relies on `abi_return_type` (and `abi_argument_types`) to return `None` when the caller is _not_ requesting an ABI dryrun.

In my inspection, I saw _no_ reason the exception need be thrown here.  So, I changed the behavior.  Though I'm less familiar with the logic than you, and I request you double check the change.

tzaffi: Though the previous is admittedly clunky, I don't support this change. The reason is that the `abi_*` methods only make sense on `PyTealDryRunExecutor`'s which have been instantiated on `@ABIReturnSubroutine` decorated methods (as opposed to regular `@Subroutine`'s). Perhaps this calls for a class-hierarchy whereby the `@ABIReturnSubroutine`-compliant objects subclass the more generic ones, but it's probably overkill in this testing context. But I'm open to further discussion.

tests/blackbox.py
```diff
@@ -131,21 +131,25 @@ def decorator_blackbox(func: SubroutineFnWrapper | ABIReturnSubroutine):
 
 # ---- API ---- #
 
+O = TypeVar("O")
```
tzaffi, line 134: If we follow this approach, I recommend changing the type variable to another name such as `Y` our `Output` because the `O` looks to much like a zero to me, and this is confusing

tests/blackbox.py
```diff
@@ -131,21 +131,25 @@ def decorator_blackbox(func: SubroutineFnWrapper | ABIReturnSubroutine):
...
 
-def mode_to_execution_mode(mode: Mode) -> blackbox.ExecutionMode:
-    if mode == Mode.Application:
-        return blackbox.ExecutionMode.Application
-    if mode == Mode.Signature:
-        return blackbox.ExecutionMode.Signature
 
-    raise Exception(f"Unknown mode {mode} of type {type(mode)}")
+def _match_mode(mode: Mode, app: Callable[[], O], signature: Callable[[], O]) -> O:
```
tzaffi, line 137: This is clever, but perhaps too clever? It took me quite a while to grok what was happening and I'm not sure such simple logic really requires all this fancy footwork. I think a little violation of D.R.Y. is ok in this case.

If you feel strongly about it then consider renaming `signature` and `app` resp. to `app_case_lambda` and `signature_case_lambda`. 

I think I see now why you were thinking about partial. I believe you can achieve similar functionality with a little class such as the following (but haven't tested it and not convinced it has any advantages):

```python
@dataclass
class _mode_matcher:
    app_case_lambda: Callable[[], Output]
    signature_case_lambda: Callable[[], Output]

   def __call__(self, mode: Mode) -> Output:
       ... match similar to the as above ...
```
Usage:
```python
>>> mode_to_execution_mode = _match_mode(lambda: blackbox.ExecutionMode.Application, lambda: blackbox.ExecutionMode.Signature)
...
>>> graviton_mode = mode_to_execution_mode(mode)
# inside `PyTealDryRunExecutor`:
    dryrun_sequence_runner = _match_mode(lambda: compileTeal...)
# etc...
```

tests/integration/ecdsa_test.py
```diff
@@ -1,5 +1,3 @@
-from graviton.blackbox import DryRunExecutor
```
tzaffi, line 1: Thanks for refactoring!

tests/integration/graviton_test.py
```diff
@@ -11,8 +11,8 @@
     Blackbox,
```
tzaffi, line 11: Much better API!

tzaffi (changes requested):
I'm asking for the assertions regarding abi-type related methods to be put back as unfortunately these only make sense for `@ABIReturnSubroutine` and we lack a class-hierarchy that can handle it.

Regarding the more economical `mode`-switching capability. I'm somewhat conflicted but don't feel strongly enough about it to block it. Maybe  **Callable** `@dataclass` approach that I suggest is a middle ground?

tzaffi (commented):
Upon further thought, I see no reason to object to removing the `NotImplementedError` from the `abi_*` methods. You were saying that the exceptions are never thrown, and in fact, it appears that in the case of non-abi, we actually get sensible results (e.g. `[None] * num_args` in the case of `abi_argument_types()`) which could actually be useful.

I would just ask that a unit test be added (or cases be added to existing tests) which assert this behavior. That way we know it's intentional. 

For example I could see a test added to `tests/unit/blackbox_test.py` which looks something like:

```python
def test_PyTealDryRunExecutor_methods():
    ptdr_vanilla = PyTealDryRunExecutor(... some dummy @Subroutine ...)

    # basic assertions on ptdr_vanilla abi methods

    ptdr_abi = PyTealDryRunExecutor(... some dummy @ABIReturnSubroutine ...)
    # basic assertions on ptdr_abi abi methods

```

tests/blackbox.py
michaeldiamant: @tzaffi I've attempted to add requested unit tests in https://github.com/algorand/pyteal/pull/338/commits/cf938ae962e4499c294f0b39d1e5efcd655b6fd3.  If it looks good to you, can you resolve the thread?

tests/blackbox.py
michaeldiamant: @tzaffi A few thoughts here:

> This is clever, but perhaps too clever?

I went down the shown path because of the need to repeat the pattern match N-times over.  Due to lack of exhaustiveness checking, my primary concern is ensuring the unsupported case exists.  The structure adds a little overhead in exchange for guaranteeing correct usage.

I can see the approach is probably less Pythonic and more Scala-like.  It's not my intent to make comprehension opaque.

>  achieve similar functionality with a little class 

https://github.com/algorand/pyteal/pull/338/commits/685ae794536cb6d83ccb8fab1ff3d21b69bbdb89 implements the suggestion.

My 2c (biased by Scala background) is that the class approach adds more indirection than the function.  That said, my primary desire to is stay within Pythonic norms:
* _If_ you find the class approach easier to reason about and valuable enough to retain, I'm happy to keep as is.
* _If_ you're still on the edge, feel welcomed to tell me.  I'd rather remove the questionable functionality.

tests/blackbox.py
michaeldiamant: @tzaffi No problem - I made a change in https://github.com/algorand/pyteal/commit/685ae794536cb6d83ccb8fab1ff3d21b69bbdb89 and we may move away from the approach wholesale.  Resolving though feel welcomed to re-open if there's additional concerns.

tests/unit/blackbox_test.py
```diff
@@ -200,15 +200,71 @@ def test_abi_blackbox_pyteal(subr_abi, mode):
...
     with open(save_to, "w") as f:
         f.write(compiled)
 
     assert_teal_as_expected(save_to, FIXTURES / "abi" / (name + ".teal"))
+
+
+@pytest.mark.parametrize("mode", (pt.Mode.Application, pt.Mode.Signature))
+@pytest.mark.parametrize(
+    "fn, expected_is_abi", ((utest_noop, False), (fn_0arg_uint64_ret, True))
+)
```
tzaffi, line 220: Nice. I didn't realize you can stack the parametrizations this way!

tzaffi (approved):
Great improvement!!!

PR #339, opened 2022-05-13
adding byte type hint to DyanamicArray parent class

barnjamin:
I noticed I was getting `Any` as the type hint on the argument to the lambda function in something like this:

```py
a[idx.load()].use(lambda v: buff.store(Concat(buff.load(), v.encode())))
```
Where `a` is a String.

This seems to give the correct type hint

### DIFF
--- a/pyteal/ast/abi/string.py
+++ b/pyteal/ast/abi/string.py
@@ -1,6 +1,7 @@
 from typing import Union, TypeVar, Sequence
 from collections.abc import Sequence as CollectionSequence
 
+from pyteal.ast.abi.uint import Byte
 from pyteal.ast.abi.type import ComputedValue, BaseType
 from pyteal.ast.abi.array_dynamic import DynamicArray, DynamicArrayTypeSpec
 from pyteal.ast.abi.uint import ByteTypeSpec, Uint16TypeSpec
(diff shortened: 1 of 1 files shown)


pyteal/ast/abi/string.py
```diff
@@ -67,7 +68,7 @@ def set(
         match value:
             case ComputedValue():
                 if value.produced_type_spec() == StringTypeSpec():
-                    return value.store_into(self)
+                    return value.store_into(cast(DynamicArray[T], self))
```
ahangsu, line 71: nit: I think you can circumvent both cast cases by removing `T` type var in `set` signature, by using `Byte` directly?

ahangsu (commented):
seems correct and build passes, just one thing

pyteal/ast/abi/string.py
barnjamin: This worked for removing the cast on computed but not on the sequence because it thinks its possible for `str` and `byte` types to make it there:

```
pyteal/ast/abi/string.py:90: error: Argument 1 to "set" of "DynamicArray" has incompatible type "Union[Sequence[Byte], str, bytes]"; expected "Union[Sequence[Byte], DynamicArray[Byte], ComputedValue[DynamicArray[Byte]]]"
```

ahangsu (approved):
seems correct

pyteal/ast/abi/string.py
ahangsu: ok I see, then that makes sense. I guess you just need to remove the `T` type var.

tzaffi (approved):
LGTM

PR #340, opened 2022-05-13
Method sig for router

barnjamin:
Adding method to get the method sig from the ABIReturnSubroutine

### DIFF
--- a/pyteal/ast/router.py
+++ b/pyteal/ast/router.py
@@ -80,6 +80,11 @@ class Router:
         )
         clear_state_conds: list[Expr] = []
 
+        if method_signature == "" and method_to_register is not None:
+            raise TealInputError(
+                "A method_signature must only be provided if method_to_register is not None"
+            )
+
         # Check:
         # - if current condition is for *ABI METHOD*
         #   (method selector && numAppArg == 1 + min(METHOD_APP_ARG_NUM_LIMIT, subroutineSyntaxArgNum))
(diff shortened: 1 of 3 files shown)


pyteal/ast/subroutine_test.py
```diff
@@ -170,6 +187,7 @@ def fn_2arg_1ret_with_expr(
             ],
             "fn_2arg_1ret_with_expr",
             pt.abi.ByteTypeSpec(),
+            "fn_2arg_1ret_with_expr(byte[10])byte",  # TODO: this seems wrong, no?
```
barnjamin, line 190: No way this is right, right? 

The first arg is an Int, not an abi type

ahangsu: this leads back to the comment I had on `signature` method, which only applies to registrable method.

pyteal/ast/router.py
```diff
@@ -68,6 +68,7 @@ def __init__(self, name: Optional[str] = None) -> None:
 
     @staticmethod
     def __parse_conditions(
+        method_signature: str,
```
ahangsu, line 71: nit: would be good to add a check for `method_signature == None` if `method_to_register == None`

ahangsu (commented):
looks largely good to me, a few things to notice

ahangsu:
you probably need to satisfy the build checkers first

PR #343, opened 2022-05-16
ABI Reference Types

barnjamin:
Adding reference types to ABI module so we can easily use them in Router.

Targeting abi-router since I'm using it for testing so branched off abi-router 

Needs tests but want to make sure this is a reasonable way to attempt this


ahangsu:
I guess this should comes to `feature/abi` first, not `abi-router`, since router work is kind of parallel to reference type?

pyteal/ast/abi/reference_type.py
```diff
@@ -0,0 +1,119 @@
...
+
+from pyteal.ast.abi.uint import Uint, UintTypeSpec
+
+
+class ReferenceTypeSpec(UintTypeSpec):
+    def __init__(self) -> None:
+        return super().__init__(8)
+
+
+ReferenceTypeSpec.__module__ = "pyteal"
```
ahangsu, line 17: nit: some ignorable comments
- can be inherited from `uint8`
- `__str__` and `new_instance` not yet defined, probably defined as abstract method?

ahangsu (commented):
took a quick pass and have some negligible comments, and the only unsettled question is `set`, I think.

pyteal/ast/abi/reference_type.py
barnjamin: can it be? I just tried and maybe confused myself but if I use uint8 I have no opportunity to set the typespec. I _think_ I need to follow the model of abi.Byte

pyteal/ast/abi/reference_type_test.py
```diff
@@ -0,0 +1,262 @@
+import pyteal as pt
```
algoidurovic, line 1: Might be a good idea to have a test that calls `compileTeal()`, to enforce that the other steps of the compilation process (like assembly) work as intended.

pyteal/ast/abi/reference_type.py
```diff
@@ -0,0 +1,103 @@
+from abc import abstractmethod
+from typing import cast
+from pyteal.ast.abi.uint import Uint, UintTypeSpec
+
+
+class ReferenceTypeSpec(UintTypeSpec):
```
algoidurovic, line 6: I'm not really sure what purpose this class and `ReferenceType` serve. They complicate the inheritance hierarchy so I think there should be a good justification for their presence. Do you mind elaborating on why they're necessary?

algoidurovic (commented):
Looks good overall, just a few points I'd like you to address.

pyteal/ast/abi/reference_type.py
barnjamin: I had originally done it to keep the bit size setting to 8 in a single spot but since both you and @ahangsu both mentioned it, I've taken it out.

ahangsu:
one thing to notice, Zeph merged in #322 to feature/abi today, you might want to update too...

tzaffi:
> one thing to notice, Zeph merged in #322 to feature/abi today, you might want to update too...

RIght. In particular, it introduced a [new member annotation_type](https://github.com/algorand/pyteal/blob/dd415b9ff0c83a23270e2f0595979b04a68899db/pyteal/ast/abi/type.py#L23) that should also be implemented in this PR.

tzaffi:
Also, not to hold up the PR, but if simple test cases could be added [here](https://github.com/algorand/pyteal/blob/dd415b9ff0c83a23270e2f0595979b04a68899db/tests/integration/abi_roundtrip_test.py#L24), that would be great. I believe that the companion `abi_roundtrip.py` class should be able to handle these new types (because their `TypeSpec`s subclass the `UintTypeSpec`) but I'm not 100% sure.

pyteal/ast/abi/reference_type.py
algoidurovic: Sounds good. Resolving.

barnjamin:
I think to actually fix tests we need to have a change to the py sdk to support reference types https://github.com/algorand/py-algorand-sdk/blob/master/algosdk/abi/base_type.py#L59

tzaffi:
> I think to actually fix tests we need to have a change to the py sdk to support reference types https://github.com/algorand/py-algorand-sdk/blob/master/algosdk/abi/base_type.py#L59

That's a good finding. Interestingly, we do have a [related test case](https://github.com/algorand/algorand-sdk-testing/blob/master/features/integration/abi.feature#L69) for handling reference types in method signatures, so I am a bit puzzled by how we pass that test without explicitly handling reference types in `base_type.py#L59`.

@algochoi - it's been a while since you worked on this but maybe you have an insight?

algochoi:
> > I think to actually fix tests we need to have a change to the py sdk to support reference types https://github.com/algorand/py-algorand-sdk/blob/master/algosdk/abi/base_type.py#L59
> 
> That's a good finding. Interestingly, we do have a [related test case](https://github.com/algorand/algorand-sdk-testing/blob/master/features/integration/abi.feature#L69) for handling reference types in method signatures, so I am a bit puzzled by how we pass that test without explicitly handling reference types in `base_type.py#L59`.
> 
> @algochoi - it's been a while since you worked on this but maybe you have an insight?

The transaction references are handled in the [`atomic_transaction_composer.py`](https://github.com/algorand/py-algorand-sdk/blob/de80435d5cc712956ff3a4dc4f96b0c06713d691/algosdk/atomic_transaction_composer.py#L274) and is only used when the method signature is parsed. +1 to this was a relatively recent thing we added.

tzaffi:
Is there a reason that this is masing of `abi-router` instead of `feature/abi` ?

tests/integration/abi_roundtrip_test.py
```diff
@@ -103,11 +109,44 @@ def roundtrip_setup(abi_type):
...
+    missing_cases = all_abi_names - top_level_names
+    assert not missing_cases, f"missing round trip tests for {missing_cases}"
+
+
 @pytest.mark.parametrize("abi_type", ABI_TYPES)
 def test_pure_compilation(abi_type):
     print(f"Pure Compilation Test for {abi_type=}")
     abi_type, dynamic_length, roundtripper = roundtrip_setup(abi_type)
 
+    # TODO: Remove once we have necessary bits in SDK for reference types
```
tzaffi, line 146: I recommend a slightly more communicative approach. Please see the `BAD_TEALS` dict below. I recommend adding a new value variable
```python
GAI_ISSUE_2068 = "https://github.com/algorand/go-algorand-internal/issues/2068"
```
as well as the following keys to `BAD_TEALS`:
```
"account": GAI_ISSUE_2068,
"asset": GAI_ISSUE_2068,
"application": GAI_ISSUE_2068,
```
and this will skip the graviton round-trip tests. It won't skip the pure-TEAL generation though. If that's still broken, we can either see how easy it would be to fix the TEAL generation, or just skip those generations in the same way.

ahangsu:
> Is there a reason that this is basing off of `abi-router` instead of `feature/abi` ?

I think Ben is testing his code against my router implementation, this PR going into `feature/abi` makes more sense to me.

barnjamin:
superseded by #361

pyteal/ast/abi/reference_type_test.py
algoidurovic: On second thought, I don't think this is needed. I was worried that you weren't testing the full compilation - sometimes bugs crop up towards the end of the process. The integration tests cover this though so feel free to disregard.

PR #345, opened 2022-05-17
Static decode

barnjamin:
When decoding a static array we should know ahead of time what the length is.

Here we override the decode method from Array to set the length manually to the length computed from `length()` method on StaticArray type.

This _does_ make the output teal longer during decode since it applies an extra Extract operation

### DIFF
--- a/pyteal/ast/abi/address_test.py
+++ b/pyteal/ast/abi/address_test.py
@@ -76,7 +76,7 @@ def test_Address_decode():
                     substringForDecoding(
                         encoded,
                         startIndex=startIndex,
-                        length=value.length(),
+                        length=pt.Int(value.type_spec().byte_length_static()),
                     )
                 )
                 expected, _ = expectedExpr.__teal__(options)
--- a/pyteal/ast/abi/address_test.py
+++ b/pyteal/ast/abi/address_test.py
@@ -1,6 +1,7 @@
 import pytest
 import pyteal as pt
 from pyteal import abi
+from pyteal.ast.abi.address import AddressLength
 
 from pyteal.ast.abi.type_test import ContainerType
 from pyteal.ast.abi.util import substringForDecoding
@@ -68,15 +69,15 @@ def test_Address_decode():
                         )
                     continue
 
-                expr = value.decode(
-                    encoded, startIndex=startIndex, endIndex=endIndex, length=length
-                )
+                expr = value.decode(encoded, startIndex=startIndex)
                 assert expr.type_of() == pt.TealType.none
                 assert expr.has_return() is False
 
                 expectedExpr = value.stored_value.store(
                     substringForDecoding(
-                        encoded, startIndex=startIndex, endIndex=endIndex, length=length
+                        encoded,
+                        startIndex=startIndex,
+                        length=pt.Int(int(AddressLength.Bytes)),
                     )
                 )
                 expected, _ = expectedExpr.__teal__(options)
--- a/pyteal/ast/abi/address_test.py
+++ b/pyteal/ast/abi/address_test.py
@@ -1,7 +1,6 @@
 import pytest
 import pyteal as pt
 from pyteal import abi
-from pyteal.ast.abi.address import AddressLength
 
 from pyteal.ast.abi.type_test import ContainerType
 from pyteal.ast.abi.util import substringForDecoding
@@ -77,7 +76,7 @@ def test_Address_decode():
                     substringForDecoding(
                         encoded,
                         startIndex=startIndex,
-                        length=pt.Int(int(AddressLength.Bytes)),
+                        length=value.length(),
                     )
                 )
                 expected, _ = expectedExpr.__teal__(options)
--- a/pyteal/ast/abi/array_static.py
+++ b/pyteal/ast/abi/array_static.py
@@ -92,11 +92,6 @@ class StaticArray(Array[T], Generic[T, N]):
                 encoded, startIndex=startIndex, endIndex=endIndex, length=length
             )
 
-        if length is not None:
-            raise TealInputError(
-                "Expected None for endIndex and length on static array"
-            )
-
         return super().decode(
             encoded,
             startIndex=startIndex,
--- a/pyteal/ast/abi/array_static.py
+++ b/pyteal/ast/abi/array_static.py
@@ -152,7 +152,7 @@ class StaticArray(Array[T], Generic[T, N]):
         Returns:
             A PyTeal expression that represents the static array length.
         """
-        return Int(int(self.type_spec().length_static()))
+        return Int(self.type_spec().length_static())
 
     def __getitem__(self, index: Union[int, Expr]) -> "ArrayElement[T]":
         if type(index) is int and index >= self.type_spec().length_static():
--- a/pyteal/ast/abi/array_static.py
+++ b/pyteal/ast/abi/array_static.py
@@ -78,6 +78,21 @@ class StaticArray(Array[T], Generic[T, N]):
     def type_spec(self) -> StaticArrayTypeSpec[T, N]:
         return cast(StaticArrayTypeSpec[T, N], super().type_spec())
 
+    def decode(
+        self,
+        encoded: Expr,
+        *,
+        startIndex: Expr = None,
+        endIndex: Expr = None,
+        length: Expr = None,
+    ) -> Expr:
+        if endIndex is not None or length is not None:
+            raise TealInputError(
+                "Expected None for endIndex and length on static array"
+            )
+
+        return super().decode(encoded, startIndex=startIndex, length=self.length())
+
     def set(
         self,
         values: Union[
@@ -127,7 +142,7 @@ class StaticArray(Array[T], Generic[T, N]):
         Returns:
             A PyTeal expression that represents the static array length.
         """
-        return Int(self.type_spec().length_static())
+        return Int(int(self.type_spec().length_static()))
 
     def __getitem__(self, index: Union[int, Expr]) -> "ArrayElement[T]":
         if type(index) is int and index >= self.type_spec().length_static():
--- a/pyteal/ast/abi/array_static.py
+++ b/pyteal/ast/abi/array_static.py
@@ -92,7 +92,7 @@ class StaticArray(Array[T], Generic[T, N]):
                 encoded, startIndex=startIndex, endIndex=endIndex, length=length
             )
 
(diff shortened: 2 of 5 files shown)


pyteal/ast/abi/address_test.py
```diff
@@ -68,15 +68,15 @@ def test_Address_decode():
...
+                expr = value.decode(encoded, startIndex=startIndex)
                 assert expr.type_of() == pt.TealType.none
                 assert expr.has_return() is False
 
                 expectedExpr = value.stored_value.store(
                     substringForDecoding(
-                        encoded, startIndex=startIndex, endIndex=endIndex, length=length
+                        encoded,
+                        startIndex=startIndex,
+                        length=value.length(),
```
tzaffi, line 79: I'm getting different compilation results for this branch compared to mine. For example, when compiling [this PyTEAL](https://github.com/algorand/pyteal/blob/1ee615117da7617078fa2d68ad8546966a734eb7/tests/unit/blackbox_test.py#L124):

```python
@Blackbox(input_types=[None, pt.TealType.uint64, None])
@pt.ABIReturnSubroutine
def fn_3mixed_args_0ret(
    a: pt.abi.Uint64, b: pt.ScratchVar, C: pt.abi.StaticArray[pt.abi.Byte, Literal[10]]
) -> pt.Expr:
    return pt.Return()
```

I get the following change in compilation output:
```
#pragma version 6
txna ApplicationArgs 0
btoi
store 3
txna ApplicationArgs 1
btoi
store 4
txna ApplicationArgs 2
extract 0 10          <----------- NEW ADDITIONAL LINE (everything else the same)
store 5
load 3
int 4
load 5
callsub fn3mixedargs0ret_0
int 1
return

// fn_3mixed_args_0ret
fn3mixedargs0ret_0:
store 2
store 1
store 0
retsub
```

tzaffi: Not sure why yet, or the impact, but I wanted to bring it to your attention.

tzaffi (commented):
I left a comment regarding an observed extra (probably correct) line of teal that is generated due to these changes. Even if correct, it's probably unnecessary to have this extra line and so this should be avoided if possible, or at least understood within a greater context that justifies the additional op.

pyteal/ast/abi/address_test.py
barnjamin: I mentioned this in the summary above. It may make sense to put this on the dev to specify a length if necessary but also seems incorrect to allow `decode` to read in more bytes than should be allowed given the types.

tzaffi: Ok, I see. I'll have to think about the implications some more. Meanwhile, some of the roundtrip graviton tests on my branch are breaking. These might be fixed if I specify the extract length, but my first hunch is that I was just doing vanilla operations and am surprised by the break. It's hard to show the "pyteal" because it's auto generated, but here's the teal (approximately) that it generated, and that teal gives an error that indicates that the expected log wasn't generated.
```
00:'#pragma version 6'
01:'txna ApplicationArgs 0'
02:'extract 0 1'
03:'store 10'
04:'load 10'
05:'callsub roundtripper_2'
06:'store 0'
07:'byte 0x151F7C75'
08:'load 0'
09:'concat'
10:'log'
11:'int 1'
12:'return'
13:''
14:'// numerical_comp'
15:'numericalcomp_0:'
16:'store 8'
17:'int 18446744073709551615'
18:'load 8'
19:'-'
20:'store 9'
21:'load 9'
22:'retsub'
```

pyteal/ast/abi/array_static.py
```diff
@@ -78,6 +78,21 @@ def __init__(self, array_type_spec: StaticArrayTypeSpec[T, N]) -> None:
...
+        startIndex: Expr = None,
+        endIndex: Expr = None,
+        length: Expr = None,
+    ) -> Expr:
+        if endIndex is not None or length is not None:
+            raise TealInputError(
+                "Expected None for endIndex and length on static array"
+            )
+
+        return super().decode(encoded, startIndex=startIndex, length=self.length())
```
tzaffi, line 94:
```suggestion
        return super().decode(encoded, startIndex=startIndex, length=self.byte_length())
```

Maybe this fixes the bug I observed?

pyteal/ast/abi/address_test.py
barnjamin: Ah, length is just number of elements not byte length I guess?

pyteal/ast/abi/array_static.py
tzaffi: sorry `byte_length_static`

tzaffi: ...of the associated typespec

pyteal/ast/abi/address_test.py
tzaffi: yep

tzaffi: When I fixed locally on my branch all the graviton tests pass, except the most ambitious. I get the following error:

```
>           raise TealInputError(
                "Expected None for endIndex and length on static array"
            )
E           pyteal.TealInputError: Expected None for endIndex and length on static array

pyteal/ast/abi/array_static.py:88: TealInputError
```
Not sure if I'm running out of gas in the dry run, or there is some deeper issue.

pyteal/ast/abi/array_static.py
```diff
@@ -78,6 +78,21 @@ def __init__(self, array_type_spec: StaticArrayTypeSpec[T, N]) -> None:
...
+        startIndex: Expr = None,
+        endIndex: Expr = None,
+        length: Expr = None,
+    ) -> Expr:
+        if endIndex is not None or length is not None:
+            raise TealInputError(
+                "Expected None for endIndex and length on static array"
+            )
+
+        return super().decode(encoded, startIndex=startIndex, length=self.byte_length())
```
ahangsu, line 94: nit: let's consider a case where static array has elements that are dynamic, then this static byte length trick cannot work, so I propose we should add a condition check here for dynamic array element.

pyteal/ast/abi/address_test.py
ahangsu: I think there are two things in here, `length` is more likely to be static array element number, while `byte_length` might be the static length of the byte encoding (can precompute for static type).

tzaffi: I create a temp branch showing the [failing test](https://github.com/algorand/pyteal/runs/6479500080?check_suite_focus=true#step:8:100).

pyteal/ast/abi/array_static.py
barnjamin: I added a check and a couple tests in the DYNAMIC_TYPES list in array_base_test.py

pyteal/ast/abi/address_test.py
barnjamin: @tzaffi can you give it another shot with the latest changes?

tzaffi: It's still failing, but this time I think I know why. Consider the following ABI type which is failing:

```python
        abi.DynamicArray[
            abi.Tuple4[
                abi.StaticArray[abi.Byte, Literal[4]],
                abi.Tuple2[abi.Bool, abi.Bool],
                abi.Uint64,
                abi.Address,
            ]
        ],
```
In particular, we have a `Tuple4[StaticArray[...]]`.  If we have instance `w, x: ...` of this type and we set one to the other `w.set(x)` eventually we hit [this line in tuple::indexTuple()](https://github.com/algorand/pyteal/blob/d21678e7444cd4e4481c7ccd0e7f3c641fb269f7/pyteal/ast/abi/tuple.py#L201) which supplies a non-empty `length` to `x.decode()` and results in the error above [at this line](https://github.com/algorand/pyteal/blob/4c1d4cd46d1eb5d258b9bd11db2218f42a5277c4/pyteal/ast/abi/array_static.py#L94). 

So I suspect that the check for `endIndex == None == length` isn't quite right.

pyteal/ast/abi/array_static.py
ahangsu: seems correct to me, resolving

pyteal/ast/abi/array_static.py
```diff
@@ -127,7 +152,7 @@ def length(self) -> Expr:
         Returns:
             A PyTeal expression that represents the static array length.
         """
-        return Int(self.type_spec().length_static())
+        return Int(int(self.type_spec().length_static()))
```
ahangsu, line 155: nit: do we need this `int` specify? or it is for python failing to infer that it is `StaticArrayTypeSpec`?

pyteal/ast/abi/address_test.py
ahangsu: oh... then there is indeed something magical behind these small code changes...

ahangsu: I have a minor conjecture (or suggestion):

what if we don't enforce that `decode` in `StaticArray` to raise error in `endIndex is None or length is None`, then we just do the decoding like its parent `ArrayBase` and I guess the behavior is preserved?

barnjamin: I can remove the error thrown on endIndex and pass it into decode but I'm torn on the length check since we set it directly it'd seem strange to allow length argument but not use it?

jasonpaulos:
> > Is the purpose of this change isn't clear to me, could you help me understand that?
> 
> Currently if I try to use something like `(abi.Address()).decode(rawbytes)` it'll just read until the end of the byte array. Whereas if I do `(abi.Uint8()).decode(rawbytes)` it'll read only that integer value.
> 
> This difference in behavior confused me for a while and since we should know ahead of time the length of the static array it seems reasonable to read up to that length

Ah, I see, but I'd argue that the `Uint8` decoding is in the wrong here. Check out the docstring for `abi.BaseType.decode`: https://github.com/algorand/pyteal/blob/4f092a4e415dbd06f0c446ba7c34f5bc7c528b09/pyteal/ast/abi/type.py#L104-L134

Uint decoding should _probably_ use `btoi` and assert that the length of the string to decode is correct if `startIndex` isn't present, that way it will adhere to the spec more.

barnjamin:
@jasonpaulos sorry, you're right I _do_ specify the startIndex in the byte array but the length still needs to be specified for static arrays

https://github.com/algorand-devrel/wormhole-demo/blob/main/demos/portal_transfer/vaa.py#L78-L80

barnjamin:
Closing as working as intended

PR #347, opened 2022-05-17
Optionally refactor handling of array_length int subclasses in StaticArrayTypeSpec

michaeldiamant:
Optionally proposes a refactoring to the question raised in https://github.com/algorand/pyteal/pull/322#discussion_r873197158.

Rationale:  It's less error prone to centralize the subclass check in `StaticArrayTypeSpec` rather than have the downstream user remember to cast.

Assumption:  It's _safe_ to treat any `int` subclass as int.  The intuition feels reasonable though it should be double checked.

### DIFF
--- a/pyteal/ast/abi/address.py
+++ b/pyteal/ast/abi/address.py
@@ -22,7 +22,7 @@ AddressLength.__module__ = "pyteal"
 
 class AddressTypeSpec(StaticArrayTypeSpec):
     def __init__(self) -> None:
-        super().__init__(ByteTypeSpec(), int(AddressLength.Bytes))
+        super().__init__(ByteTypeSpec(), AddressLength.Bytes)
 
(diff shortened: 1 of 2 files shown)


tzaffi (approved):
niceeee!

tzaffi:
> Optionally proposes a refactoring to the question raised in [#322 (comment)](https://github.com/algorand/pyteal/pull/322#discussion_r873197158).
> 
> Rationale: It's less error prone to centralize the subclass check in `StaticArrayTypeSpec` rather than have the downstream user remember to cast.
> 
> Assumption: It's _safe_ to treat any `int` subclass as int. The intuition feels reasonable though it should be double checked.

Subclassing `int` is an unlikely situation. We should discourage it in this repo except for a really compelling reason.

PR #348, opened 2022-05-17
DO NOT MERGE

tzaffi:


ahangsu:
this seems to be testing Ben's previous static decoding PR, which he closed this week, do we want to keep this one open?

PR #349, opened 2022-05-18
Refactor ABIRoundtrip to always use annotation_type

michaeldiamant:
Optionally refactors `ABIRoundtrip` to remove branching during construction.

The PR stems from my attempt to understand why #322 requires `annotation_type`.  As a by-product of the investigation, I realized `ABIRoundtrip` branches because unlike all other usages, `tests/integration/abi_roundtrip_test.py` supplies `GenericAlias` instead of an equivalent instance.

From my review, I felt removing the branching produces a better tradeoff.  Though I present the PR optionally in case I've miss understood usage patterns.

### DIFF
--- a/tests/abi_roundtrip.py
+++ b/tests/abi_roundtrip.py
@@ -1,4 +1,4 @@
-from typing import Generic, TypeVar, cast
+from typing import Generic, TypeVar
 
 import pyteal as pt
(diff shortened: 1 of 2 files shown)


tzaffi (approved):
Great improvement!

PR #356, opened 2022-05-20
WIP: test will's docker image

tzaffi:


docker-compose.yml
```diff
@@ -2,8 +2,18 @@ version: '3'
 
 services:
   algod:
+    image: wwinder/algod:latest
```
ahangsu, line 5: Hey, I was reading this PR. Not knowing much about docker config, but it seems we are going to replace docker image from `makerxau/algorand-sandbox-dev` to `wwinder/algod`?

It seems to me that, both are working under `DEV_MODE`, seemingly not too much perf change. Is it because Will is maintaining these image regularly?

docker-compose.yml
```diff
@@ -2,8 +2,18 @@ version: '3'
...
+    ports:
+      - "4001:8080" # algod
+      - "4160:4160" # gossip
+      - "9100:9100" # prometheus
+    environment:
+      - DEV_MODE
+      - TOKEN=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
+
+  # TODO: do not merge with this kmd service. This is only needed cause I don't want to
+  # stand-up a seperate service locally
```
tzaffi, line 15: ^^^ yep: DO NOT MERGE ^^^

docker-compose.yml
tzaffi: If we're going to write true multi-round e2e tests in addition to dry run tests, then yes, a `DEV_MODE` version would be useful. 

@algojohnlee @winder - is having a dev mode docker image variant on the roadmap as well?

winder: It is supported, set `DEV_MODE: true` to enable devmode, see the environment variables section: https://hub.docker.com/r/wwinder/algod

I believe the "latest" tag is stale because I don't delete it before adding the new tags, I should be able to fix that.

winder: I see that you're already using dev mode, does it not work the way you expect it?

tzaffi: Thanks for clarifying!!! It's working well, but the _theoretical_ issue remains that fingers are crossed that the image will be ready in time for when it's used in tests.

ahangsu: > I believe the "latest" tag is stale because I don't delete it before adding the new tags

Just one question, if we want to specify the latest version of your image, what tag should we specify? Or we have to pin a SHA?

winder: You're specifying the correct tag, I just haven't implemented it properly

docker-compose.yml
```diff
@@ -2,8 +2,18 @@ version: '3'
...
+  kmd:
     image: makerxau/algorand-sandbox-dev:latest
```
winder, line 17: Does this even work? I would think you want kmd to talk /w algod. It could be added to our docker container.

tzaffi: Didn't run the full test on it so it may not work (and it was only temporary). But yeah, adding KMD to the docker would be great.

tzaffi: to clarify - current pyteal tests are dry run only so not using KMD

PR #358, opened 2022-05-21
adding sizeof util

barnjamin:
Should be a useful utility especially when writing large structures to boxes


jasonpaulos (approved):
Makes sense

tzaffi (approved):
LGTM

PR #359, opened 2022-05-22
Fix incorrect reference to global account in local state docs

Moataz-E:


ahangsu (approved):
lgtm, thanks for the catch

PR #361, opened 2022-05-25
Reference Types

barnjamin:
Adding reference types to ABI types.

Note: Several skip cases exist in the round trip tests until we have support for them in the SDK


pyteal/ast/subroutine.py
```diff
@@ -589,6 +589,10 @@ def __call__(
     def name(self) -> str:
         return self.subroutine.name()
 
+    def method_signature(self) -> str:
+        args = [str(v) for v in self.subroutine.abi_args.values()]
+        return f"{self.name()}({','.join(args)}){self.type_of()}"
+
```
barnjamin, line 595: Not sure how this made it in 🤔, should I just leave it? @ahangsu guessing this will take some manual wiggling when feature/abi gets merged back into router

tests/integration/abi_roundtrip_test.py
```diff
@@ -11,6 +11,19 @@
 from tests.abi_roundtrip import ABIRoundtrip
 from tests.compile_asserts import assert_teal_as_expected
 
+GAI_ISSUE_2050 = "https://github.com/algorand/go-algorand-internal/issues/2050"
+BAD_TEALS = {
+    "()": GAI_ISSUE_2050,
+}
```
barnjamin, line 17: Just moved these to the top so the issues are easier to see

tests/integration/abi_roundtrip_test.py
```diff
@@ -101,6 +117,7 @@ def roundtrip_setup(abi_type):
 
     return (
         abi_type,
+        str(abi.type_spec_from_annotation(abi_type)),
```
barnjamin, line 120: @tzaffi added this to the return tuple rather than writing another method to just get the string after doing the tuple check

pyteal/ast/subroutine.py
ahangsu: confirmed it is from abi-router branch, is it needed in this branch? If so we can open a minor PR to include the change to feature/abi?

barnjamin: not necessary, just removed it

pyteal/ast/abi/reference_type_test.py
```diff
@@ -0,0 +1,262 @@
...
+
+def test_ApplicationTypeSpec_eq():
+    assert abi.ApplicationTypeSpec() == abi.ApplicationTypeSpec()
+
+    for otherType in (
+        abi.ByteTypeSpec(),
+        abi.Uint8TypeSpec(),
+        abi.AddressTypeSpec(),
+    ):
+        assert abi.StringTypeSpec() != otherType
```
ahangsu, line 199: same here

PR #362, opened 2022-05-25
Refactor #170 to reduce visibility and mutation

michaeldiamant:
Optionally presents a couple of refactorings considered while reviewing #170.  Feel welcomed to close if you prefer to keep as is.

Proposed changes:
* Refactor partitioning of `OnComplete` states via https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.partition.
  * Intent:  Reduce mutation and more succinctly express the intended operations.  While I stylistically prefer the PR's approach, I do _not_ feel strongly about the change.  
* Designate `oc_under_call_config` as private by renaming to `_oc_under_call_config`.
  * Intent:  Minimize end-user API surface area.  From my review, it _appears_ an end-user does _not_ need to invoke `oc_under_call_config`.  Of the 2 changes, I feel more strongly about limiting visibility.

### DIFF
--- a/pyteal/ast/router.py
+++ b/pyteal/ast/router.py
@@ -88,7 +88,7 @@ class CallConfigs:
             and self.delete_application == CallConfig.NEVER
         )
 
-    def oc_under_call_config(self, call_config: CallConfig) -> list[EnumInt]:
+    def _oc_under_call_config(self, call_config: CallConfig) -> list[EnumInt]:
         if not isinstance(call_config, CallConfig):
             raise TealInputError(
                 "generate condition based on OCMethodCallConfigs should be based on OCMethodConfig"
@@ -437,7 +437,7 @@ class Router:
         def partition(cc: CallConfig) -> Tuple[bool, list[EnumInt]]:
             (not_clear_states, clear_states) = more_itertools.partition(
                 lambda x: str(x) == str(OnComplete.ClearState),
-                call_configs.oc_under_call_config(cc),
+                call_configs._oc_under_call_config(cc),
             )
             return (len(list(clear_states)) > 0, list(not_clear_states))
 


requirements.txt
```diff
@@ -2,6 +2,7 @@ black==22.3.0
 flake8==4.0.1
 flake8-tidy-imports==4.6.0
 graviton@git+https://github.com/algorand/graviton@v0.3.0
+more-itertools==8.13.0
```
tzaffi, line 5: I think this wasn't used in the end...

pyteal/ast/router.py
```diff
@@ -427,15 +428,21 @@ def add_method_handler(
...
 
         wrapped = Router.wrap_handler(True, method_call)
 
-        if any(str(OnComplete.ClearState) == str(x) for x in oc_create):
+        def partition(cc: CallConfig) -> Tuple[bool, list[EnumInt]]:
+            (not_clear_states, clear_states) = more_itertools.partition(
+                lambda x: str(x) == str(OnComplete.ClearState),
+                call_configs._oc_under_call_config(cc),
+            )
+            return (len(list(clear_states)) > 0, list(not_clear_states))
```
tzaffi, line 442:
```suggestion
            return (bool(clear_states), not_clear_states)
```

I believe this is equivalent

pyteal/ast/router.py
```diff
@@ -427,15 +428,21 @@ def add_method_handler(
...
-        oc_create: list[EnumInt] = call_configs.oc_under_call_config(CallConfig.CREATE)
-        oc_call: list[EnumInt] = call_configs.oc_under_call_config(CallConfig.CALL)
         if method_signature in self.added_method_sig:
             raise TealInputError(f"re-registering method {method_signature} detected")
         self.added_method_sig.add(method_signature)
 
         wrapped = Router.wrap_handler(True, method_call)
 
-        if any(str(OnComplete.ClearState) == str(x) for x in oc_create):
+        def partition(cc: CallConfig) -> Tuple[bool, list[EnumInt]]:
```
tzaffi, line 437:
```suggestion
        def partition(cc: CallConfig) -> tuple[bool, list[EnumInt]]:
```

3.10 style type-hints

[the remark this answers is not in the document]
ahangsu: This function can live as a method under `CallConfigs`, so we can eliminate nested local function.

PR #365, opened 2022-05-25
Graviton last rows bugfix

tzaffi:
This bug never got detected because the incorrect param `last_rows` in `inspector.report()` never gets called, except for a failing test.


ahangsu (approved):
"lgtm"

PR #366, opened 2022-05-26
Adding some improved trace messaging

barnjamin:
I think if we change TealCompileError to accept the compile options we can inspect the current subroutine implementation for its source information.

This would require changing all the current TealCompileErrors in methods other than `__teal__` to be converted to some other error type

### DIFF
--- a/pyteal/ast/return_.py
+++ b/pyteal/ast/return_.py
@@ -37,6 +37,7 @@ class Return(Expr):
                 "TEAL version too low to use subroutines",
             )
             returnType = options.currentSubroutine.returnType
+
             if returnType == TealType.none:
                 if self.value is not None:
                     raise TealCompileError(
--- a/pyteal/ast/return_.py
+++ b/pyteal/ast/return_.py
@@ -42,13 +42,13 @@ class Return(Expr):
                 if self.value is not None:
                     raise TealCompileError(
                         "Cannot return a value from a subroutine with return type TealType.none",
-                        self,
+                        options,
                     )
             else:
                 if self.value is None:
                     raise TealCompileError(
                         "A subroutine declares it returns a value, but no value is being returned",
-                        self,
+                        options,
                     )
                 actualType = self.value.type_of()
                 if not types_match(actualType, returnType):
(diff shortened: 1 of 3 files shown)


tzaffi:
So it looks like this PR is about to turn `Expr.getDefinitionTrace()` into dead code as well as the field `Expr.trace`. So the question that I'd like to understand is are there still situations when the original approach was better than the new one?

@jasonpaulos - as the original author of `getDefinitionTrace()` you may have some insights.

barnjamin:
stray comma makes python think its meant to be a tuple
```
  File "/home/ben/vex/vex.py", line 119, in <module>
    approval, clear, spec = router.compile_program(
  File "/home/ben/pyteal/pyteal/ast/router.py", line 562, in compile_program
    ap_compiled = compileTeal(
  File "/home/ben/pyteal/pyteal/compiler/compiler.py", line 264, in compileTeal
    compileSubroutine(
  File "/home/ben/pyteal/pyteal/compiler/compiler.py", line 197, in compileSubroutine
    subroutine.get_declaration(),
  File "/home/ben/pyteal/pyteal/ast/subroutine.py", line 242, in get_declaration
    self.declaration = evaluate_subroutine(self)
  File "/home/ben/pyteal/pyteal/ast/subroutine.py", line 742, in evaluate_subroutine
    subroutine_body = subroutine.implementation(*loaded_args, **abi_output_kwargs)
  File "/home/ben/vex/vex.py", line 67, in bootstrap
    return Seq(
  File "/home/ben/pyteal/pyteal/ast/seq.py", line 48, in __init__
    raise TealInputError("{} is not a pyteal expression.".format(expr))
pyteal.TealInputError: (<pyteal.BoxCreate object at 0x7f7c6b70a350>,) is not a pyteal expression.
```

jasonpaulos:
This PR doesn't make a lot of sense to me. I understand the motivation, but showing users the file that defines the subroutine their error happens to be in seems like a step down from the current error messaging.

The current system works by showing the stack frames leading up to where the problematic `Expr` is defined. Since in _most_ cases, the user creates the expression, this works pretty well. However, you've pointed out one such case where the user doesn't define the expression, the PyTeal compiler does.

To remedy that, I think a more appropriate fix would be for the compiler to adjust the trace on the new expressions it creates. For example:

```python

# current code
ast = Seq([ast, Return()])

# proposed code
ret_expr = Return()
ret_expr.trace = ast.trace
seq_expr = Seq([ast, ret_expr])
seq_expr.trace = ast.trace

ast = seq_expr
```

By copying the original trace, the compiler's making its new expressions transparent. I've not tested any of this, but it you find it satisfactory, it would certainly be a less-invasive way to improve this this.

PR #370, opened 2022-05-27
An attempt to add decorator syntax to abi-router

ahangsu:
## Description

`method` decorator is adapted from `dataclasses`: https://github.com/python/cpython/blob/3.10/Lib/dataclasses.py#L1157-L1185, which allows one to register a method in "fast-API"-like way in following 3 forms of syntax:

```python3
@router.method
def bootstrap():
    return Seq(vex.initialize_globals(), ask_pq.initialize(), bid_pq.initialize())

@router.method(overriding_name = "newName")
def bootstrap():
    return Seq(vex.initialize_globals(), ask_pq.initialize(), bid_pq.initialize())

@router.method(overriding_name = "newNameAndOC", opt_in = CallConfig.ALL)
def bootstrap():
    return Seq(vex.initialize_globals(), ask_pq.initialize(), bid_pq.initialize())
```

### DIFF
--- a/pyteal/ast/router.py
+++ b/pyteal/ast/router.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass, field, fields, astuple
-from typing import cast, Optional
+from typing import cast, Optional, Callable
 from enum import IntFlag
 
 from algosdk import abi as sdk_abi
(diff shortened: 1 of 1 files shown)


cusma:
@barnjamin quick note on `Router.method`:

### EXPECTED USAGE
```python
@my_router.method(opt_in=CallConfig.ALL)
def my_abi_optin_method(asset_id: abi.Asset) -> Expr:
    return Seq(
        ...
    )
```

### WORKING USAGE
```python
def my_abi_optin_method(asset_id: abi.Asset) -> Expr:
    return Seq(
        ...
    )
my_router.method(my_abi_optin_method, opt_in=CallConfig.ALL)
```

ahangsu:
good catch, will fix soon

ahangsu:
fixed https://github.com/algorand/pyteal/pull/170/commits/f66b7e4b1d449ca849e742e8d57bc323ddfbddf3

cusma:
@ahangsu one note: in order to identify a method as `optin` the following doesn’t work:
```python
my_abi.method(optin_method, opt_in=CallConfig.ALL)
```
this one works instead:
```python
my_abi.method(optin_method, no_op=CallConfig.NEVER, opt_in=CallConfig.ALL)
```

Would be useful to override `no_op=` if `opt_in=` is passed.

PR #372, opened 2022-05-31
Reference Type  `deref`

barnjamin:
Adding a reference type method `deref` to allow easily getting the underlying value referred to by the reference type.

### DIFF
--- a/pyteal/ast/abi/reference_type.py
+++ b/pyteal/ast/abi/reference_type.py
@@ -1,4 +1,6 @@
 from pyteal.ast.abi.uint import Uint, UintTypeSpec
+from pyteal.ast.expr import Expr
+from pyteal.ast.txn import Txn
 
 
 class AccountTypeSpec(UintTypeSpec):
@@ -24,6 +26,9 @@ AccountTypeSpec.__module__ = "pyteal"
 class Account(Uint):
     def __init__(self) -> None:
         super().__init__(AccountTypeSpec())
+
+    def deref(self) -> Expr:
+        return Txn.accounts[self.stored_value.load()]
 
 
 Account.__module__ = "pyteal"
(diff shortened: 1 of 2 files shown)


ahangsu (commented):
I think the implementation and testing are both correct, just one comment on `deref` as a naming. afaik I don't think it is a correct naming in this scenario, for `deref` is indicating referencing to another obj.

Probably name it as `refer_to` or something else? What do yall folks think?

pyteal/ast/abi/reference_type.py
```diff
@@ -25,6 +27,9 @@ class Account(Uint):
     def __init__(self) -> None:
         super().__init__(AccountTypeSpec())
 
+    def deref(self) -> Expr:
+        return Txn.accounts[self.stored_value.load()]
```
tzaffi, line 31: Probably the following are non-issues and the onus is on the PyTeal developer to understand the value of implicit index 0, but it's worth clarifying. And if that's the case, maybe there's still a way to reduce foot shooting:

* Is this logic going to have an off-by-one problem because index-0 implicitly refers to the self account/app (see [this arc-4 section](https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0004.md#reference-types)) ? 
* For assets the logic would be different?

tzaffi:
> I think the implementation and testing are both correct, just one comment on `deref` as a naming. afaik I don't think it is a correct naming in this scenario, for `deref` is indicating referencing to another obj.
> 
> Probably name it as `refer_to` or something else? What do yall folks think?

Not sure what the perfect naming is. Maybe a more verbose name such as `referenced_object` or `dereference`? The original `deref` seems reasonable to me as it sort of feels like doing a `*ref` in C++ or go.

ahangsu:
> > I think the implementation and testing are both correct, just one comment on `deref` as a naming. afaik I don't think it is a correct naming in this scenario, for `deref` is indicating referencing to another obj.
> > Probably name it as `refer_to` or something else? What do yall folks think?
> 
> Not sure what the perfect naming is. Maybe a more verbose name such as `referenced_object` or `dereference`? The original `deref` seems reasonable to me as it sort of feels like doing a `*ref` in C++ or go.

got it! I will then resolve the comments now for naming, for it's being reasonable to you.

ahangsu (approved):

PR #373, opened 2022-05-31
Txn types

barnjamin:
Starting to add Transaction types from arc4: https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0004.md#transaction-types

feel free to take this over if you've got a better idea for implementation

I think the tricky bit will be in router parsing args and making sure we don't try to decode the Transaction types


pyteal/ast/txn.py
```diff
@@ -28,6 +28,7 @@ class TxnType:
     AssetTransfer = EnumInt("axfer")
     AssetFreeze = EnumInt("afrz")
     ApplicationCall = EnumInt("appl")
+    Any = EnumInt("txn")
```
jasonpaulos, line 31: This isn't right.

This enum matches the output of `txn TypeEnum`, which is an int. Furthermore, it uses the "enum" int ops to load these values to make it more obvious, e.g. `int acfg` will translate to the right integer that matches the `txn TypeEnum` value.

So an "any" value here doesn't make sense for two reasons: `int any` won't work, and `txn TypeEnum` will never return a value for "any"

pyteal/ast/abi/transaction.py
```diff
@@ -0,0 +1,196 @@
...
+        raise TealInputError("A Transaction cannot be decoded")
+
+    def encode(self) -> Expr:
+        raise TealInputError("A Transaction cannot be encoded")
+
+
+Transaction.__module__ = "pyteal"
+
+
+class AnyTransactionTypeSpec(TransactionTypeSpec):
```
jasonpaulos, line 87: Why have `Transaction` and `AnyTransaction`? I think it'd make more sense from an inheritance perspective to get rid of `AnyTransaction` and have `Transaction` be used when it's needed. That way a `PaymentTransaction` and is a `Transaction`.

pyteal/ast/abi/transaction.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+class Transaction(BaseType):
+    def __init__(self, spec: TransactionTypeSpec) -> None:
+        super().__init__(spec)
+
+    def type_spec(self) -> TransactionTypeSpec:
+        return cast(TransactionTypeSpec, super().type_spec())
+
+    def get(self) -> Expr:
```
jasonpaulos, line 50: Suggestion: don't make `get` return the index/offset (I can't tell which the implementation uses yet) -- most people won't care about this. Make `get` return the correct `TxnObject`, e.g. `return Gtxn[<index here>]`. Then the caller can access all the fields on it.

If you want to return the index, a separate `index` method seems more appropriate.

jasonpaulos (commented):
Just some initial thoughts

pyteal/ast/txn.py
barnjamin: I added this here so I could re-use the existing txn types and because we have an `Unknown` type already. I'll redefine them in the abi/transaction.py though for the reasons you mentioned

pyteal/ast/abi/transaction.py
barnjamin: I agree that `get` is better to return some TxnObject since that is how it will be used. 

I think it might be a departure from the other sort of reference types since they offer both `get` for the raw application_arg decoded vs `deref` for dereferencing the reference type. 

As long as this is ok, I'm happy to make this `get` return the TxnObject but I wonder if I should go back to the reference types and make them return the deref by default and provide a different method to get the app arg? 

@ahangsu @tzaffi any thoughts?

pyteal/ast/abi/transaction.py
barnjamin: you're right, I think I just tripped over some python typing thing and punted. Will make this work.

pyteal/ast/abi/transaction.py
tzaffi: I think we should do is aim for consistency. @jasonpaulos argues that most developers will prefer the actual `TxnObject` over its offset. A similar argument could be made for reference types such as `Asset` etc. If we generally expect `get` to give the "raw" data for an abi-type, then we should continue with Ben's original approach here but communicate in our docs as clearly as possible that this is the purpose of `get()`. But my hunch is that Jason's assessment of usage patterns is spot-on, and that `get()` (and similarly `__get_item__()`) should act in the most convenient manner. 

If we all agree with this philosophy around `get`, I recommend adding to this PR changes that make `deref()` become `get()` and then also allow users to retrieve the raw information either with something like `index()`as Jason suggests, or a variant such as `array_index()` etc. (I prefer `index` even if it clashes with usages of `index()` in other types because it makes sense to me to have polymorphic versions of `index()` based on the type's context).

pyteal/ast/abi/transaction.py
```diff
@@ -0,0 +1,239 @@
...
+
+
+Transaction.__module__ = "pyteal"
+
+
+class PaymentTransactionTypeSpec(TransactionTypeSpec):
+    def new_instance(self) -> "Transaction":
+        return PaymentTransaction()
+
+    def annotation_type(self) -> "type[Transaction]":
```
tzaffi, line 104:
```suggestion
    def annotation_type(self) -> "type[PaymentTransaction]":
```

Feeling a bit guilty now for introducing this boiler-plate method... I haven't discovered a clever python trick to eliminate having to define in every type.

pyteal/ast/abi/transaction.py
```diff
@@ -0,0 +1,239 @@
...
+
+
+PaymentTransaction.__module__ = "pyteal"
+
+
+class KeyRegisterTransactionTypeSpec(TransactionTypeSpec):
+    def new_instance(self) -> "Transaction":
+        return KeyRegisterTransaction()
+
+    def annotation_type(self) -> "type[Transaction]":
```
tzaffi, line 126: similarly, here and below, the `annotation_type`'s return type-annotation should be as specific as possible.

pyteal/ast/abi/reference_type.py
```diff
@@ -29,9 +29,12 @@ class Account(Uint):
     def __init__(self) -> None:
         super().__init__(AccountTypeSpec())
 
-    def deref(self) -> Expr:
+    def get(self) -> Expr:
         return Txn.accounts[self.stored_value.load()]
```
tzaffi, line 33:
```suggestion
        return Txn.accounts[self.index()]
```
Can make the meaning explicit.

pyteal/ast/abi/reference_type.py
```diff
@@ -60,9 +63,12 @@ class Asset(Uint):
     def __init__(self) -> None:
         super().__init__(AssetTypeSpec())
 
-    def deref(self) -> Expr:
+    def get(self) -> Expr:
```
tzaffi, line 66: Recommend using `index()` here and in other reference `get()`'s

pyteal/ast/abi/transaction.py
```diff
@@ -0,0 +1,239 @@
...
+
+    def decode(
+        self,
+        encoded: Expr,
+        *,
+        startIndex: Expr = None,
+        endIndex: Expr = None,
+        length: Expr = None,
+    ) -> Expr:
+        raise TealInputError("A Transaction cannot be decoded")
```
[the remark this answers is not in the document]
barnjamin: the TealTypeError wants 2 args: actual and expected. In this case I'm not sure what I'd set for either of them?

pyteal/ast/abi/reference_type.py
```diff
@@ -89,8 +97,18 @@ class Application(Uint):
     def __init__(self) -> None:
         super().__init__(ApplicationTypeSpec())
 
-    def deref(self) -> Expr:
-        return Txn.applications[self.stored_value.load()]
+    def get(self) -> Expr:
+        return Txn.applications[self.index()]
+
+    def index(self) -> Expr:
+        return self.stored_value.load()
```
tzaffi, line 104: :+1

pyteal/ast/abi/transaction.py
barnjamin: I've tweaked both ref types and txn types here to return what most people would probably expect

tests/abi_roundtrip.py
```diff
@@ -65,6 +67,13 @@ def bool_comp(x: abi.Bool, *, output: abi.Bool):
     def max_int(cls, bit_size):
         return (1 << bit_size) - 1
 
+    def transaction_comp_factory(self) -> pt.ABIReturnSubroutine:
+        @pt.ABIReturnSubroutine
+        def transaction_comp(x: self.annotation, *, output: abi.Uint64):  # type: ignore[name-defined]
+            return output.set(x.get().amount())
```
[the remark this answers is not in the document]
barnjamin: We can't return the transaction type here so i just punted. If you've got a good idea for how to add testing in, feel free to pr it or just force push :)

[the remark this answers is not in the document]
tzaffi: Ahh right - the test shouldn't exemplify an ARC-4 illegal operation. I'll aim to put something together by tomorrow morning.

barnjamin:
Tests failing on diff between expected/generated teal, I expect this is because I'm pre-generating all the eventual BaseTypes in the method wrapper method and using them later instead of generating them as I notice them, so the slot assignments are a bit different

Should i just overwrite the existing teal with the teal being generated?
```
E           txna ApplicationArgs 15
E         - store 46
E         ?        ^
E         + store 44
E         ?        ^
E         - load 46
E         ?       ^
E         + load 44
E         ?       ^
E           int 0
E           extract_uint64
E         - store 44
E         ?        ^
E         + store 45
E         ?        ^
E         - load 46
E         ?       ^
E         + load 44
E         ?       ^
E           int 8
E           extract_uint64
E         - store 45
E         ?        ^
E         + store 46
E         ?        ^
E           load 30
E           load 31```

PR #375, opened 2022-06-01
Add manual test for 15+ args in ABI router

jasonpaulos:
Created a new manual test to add more confidence that the code generation for 15+ arguments in the ABI router is correct.

Also I noticed the original test uses `pt.TealComponent.Context.ignoreScratchSlotEquality()` without `pt.TealBlock.MatchScratchSlotReferences`, so there was a missed opportunity to check the equivalence of scratch slots in the two programs. I fixed this, and I added docs to `ignoreScratchSlotEquality()` which clarify why this should be done.


ahangsu (approved):
lgtm, now I understand what is "manual test" 😅

PR #376, opened 2022-06-01
Txn types

barnjamin:
Adds Transaction types from arc4: https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0004.md#transaction-types


pyteal/ast/router.py
```diff
@@ -339,52 +339,83 @@ def wrap_handler(
...
+                    cast(abi.Transaction, arg_val).set(
+                        Txn.group_index() - Int(txn_relative_pos - idx)
+                    )
+                    for idx, arg_val in enumerate(txn_arg_vals)
+                ]
+
+                decode_instructions += txn_decode_instructions
+
+            # de-tuplify into specific values using `store_into` on
+            # each element of the tuple'd arguements
```
tzaffi, line 394:
```suggestion
            # each element of the tuple'd arguments
```
typo

ahangsu:
Just one question first, do we want to explicitly disallow embedded transaction types in tuple/array?

barnjamin:
> Just one question first, do we want to explicitly disallow embedded transaction types in tuple/array?

I think initially yes since its not immediately clear how they'd be encoded, especially for dynamic arrays

pyteal/ast/router.py
```diff
@@ -339,52 +339,83 @@ def wrap_handler(
...
+
+            # All subroutine arg values, initialize here and use below instead of
+            # creating new instances on the fly so we dont have to think about splicing
+            # back in the transaction types
+            arg_vals = [typespec.new_instance() for typespec in arg_type_specs]
+
+            # Only args that appear in app args
+            app_arg_vals: list[abi.BaseType] = [
+                ats for ats in arg_vals if not isinstance(ats, abi.Transaction)
             ]
```
ahangsu, line 355: this is filtered only on the assumption that, there are no transaction types existing in underlying composite type spec, right?

ahangsu, line 355: this is filtered only on the assumption that, there are no transaction types existing in underlying composite type spec, right?

pyteal/ast/router.py
```diff
@@ -339,52 +339,83 @@ def wrap_handler(
...
+
+            # "decode" transaction types by setting the relative index
+            if len(txn_arg_vals) > 0:
+                txn_relative_pos = len(txn_arg_vals)
+                txn_decode_instructions: list[Expr] = [
+                    cast(abi.Transaction, arg_val).set(
+                        Txn.group_index() - Int(txn_relative_pos - idx)
+                    )
+                    for idx, arg_val in enumerate(txn_arg_vals)
+                ]
```
ahangsu, line 389: this part is somewhat tricky, could you explain more on relative index and group index relation?

ahangsu, line 389: this part is somewhat tricky, could you explain more on relative index and group index relation?

tests/integration/abi_roundtrip_test.py
```diff
@@ -115,13 +131,15 @@ def roundtrip_setup(abi_type):
...
-        abi_type,
-        str(abi.type_spec_from_annotation(abi_type)),
-        dynamic_length,
-        ABIRoundtrip(abi.make(abi_type), length=dynamic_length).pytealer(),
+    abi_type_str = str(abi.type_spec_from_annotation(abi_type))
+    roundtrip_or_msg = (
+        "ineligible for round trip test"
+        if abi_type_str in BAD_TYPES
+        else ABIRoundtrip(abi.make(abi_type), length=dynamic_length).pytealer()
     )
```
ahangsu, line 139: nit: not knowing much of the code, but it seems the message can be replaced by None? but if it is for the sake of readability, you can ignore me

ahangsu, line 139: nit: not knowing much of the code, but it seems the message can be replaced by None? but if it is for the sake of readability, you can ignore me

pyteal/ast/router.py
barnjamin: correct

tests/integration/abi_roundtrip_test.py
barnjamin: @tzaffi ^?

pyteal/ast/router.py
barnjamin: The transactions should appear in the group in the order they're specified in the method signature and should be relative to the current transaction.

ex:
`doit(axfer,pay,appl)`
would be 4 transactions
current_idx-3 = axfer
current_idx-2 = pay
current_idx-1 = appl
current_idx-0  = the txn that triggered the current eval

since we're iterating in order of the txns appearance in the args we just subtract the current index from the total length to get the offset.

Do you want this explanation in the source file as well as a comment?

ahangsu: great, thanks for explaining and that makes sense.

I think the explanation can live as a chunk of comment in router code, that helps a lot.

tests/integration/abi_roundtrip_test.py
ahangsu: no worries, it is of minor importance

tests/integration/abi_roundtrip_test.py
```diff
@@ -115,13 +131,15 @@ def roundtrip_setup(abi_type):
...
 
+    return (abi_type, abi_type_str, dynamic_length, roundtrip_or_msg)
```
tzaffi, line 141:
```suggestion
    roundtrip_or_none = None
    if abi_type_str not in BAD_TYPES:
        roundtrip_or_none = ABIRoundtrip(abi.make(abi_type), length=dynamic_length).pytealer()

    return (abi_type, abi_type_str, dynamic_length, roundtrip_or_none)
```

here's my suggestion reflecting @ahangsu 's remark

pyteal/ast/abi/transaction_test.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+@dataclass
+class TransactionTypeTest:
+    ts: abi.TransactionTypeSpec
+    t: abi.Transaction
+    s: str
+
+
+TransactionValues: List[TransactionTypeTest] = [
```
ahangsu, line 19: Would be nice

ahangsu (approved):
waiting on Jason's pass

pyteal/ast/abi/reference_type.py
```diff
@@ -27,8 +29,11 @@ class Account(Uint):
     def __init__(self) -> None:
         super().__init__(AccountTypeSpec())
 
-    def deref(self) -> Expr:
-        return Txn.accounts[self.stored_value.load()]
+    def get(self) -> Expr:
```
jasonpaulos, line 32: I like this improvement, but it brings to light some contradictions:
* `Account` inherits from `Unit`, but returning a non-integer here breaks the behavior you'd expect from a `Uint`.
* For `Application` and `Asset`, if you were to inspect their type specs, they claim to return 8 bit integers, but in reality they're really 64 bits.

Would it possible for reference types to not inherit from Uint for these reasons?

pyteal/ast/abi/transaction.py
```diff
@@ -0,0 +1,238 @@
...
+from pyteal.ast.txn import TxnObject
+from pyteal.ast.gtxn import Gtxn
+from pyteal.types import TealType
+from pyteal.errors import TealInputError
+
+T = TypeVar("T", bound=BaseType)
+
+
+class TransactionType(Enum):
+    Transaction = "txn"
```
jasonpaulos, line 15: nit: `Any` might be a clearer name for this one

pyteal/ast/abi/transaction.py
```diff
@@ -0,0 +1,238 @@
...
+
+    def __str__(self) -> str:
+        return TransactionType.Transaction.value
+
+
+TransactionTypeSpec.__module__ = "pyteal"
+
+
+class Transaction(BaseType):
+    def __init__(self, spec: TransactionTypeSpec = TransactionTypeSpec()) -> None:
```
jasonpaulos, line 57: nit: consider `= None` instead and creating the default value in the function body. There can be unintended side effects when you create a mutable object as a default value, since that value is reused for all function calls

pyteal/ast/router_test.py
```diff
@@ -422,35 +429,55 @@ def test_wrap_handler_method_call():
...
-                    sdk_last_arg[i - pt.METHOD_ARG_NUM_CUTOFF + 1].store_into(args[i])
-                )
         else:
             loading = [
                 arg.decode(pt.Txn.application_args[index + 1])
-                for index, arg in enumerate(args)
+                for index, arg in enumerate(app_args)
             ]
 
+        if len(txn_args) > 0:
```
jasonpaulos, line 463: This test code has code that mirrors the implementation very closely, which isn't great. I would have more confidence if you added a test with very little logic, similar to `test_wrap_handler_method_call_many_args` below -- that was added to give us more confidence that were were testing tupling arguments appropriately.

pyteal/ast/subroutine.py
```diff
@@ -597,12 +597,7 @@ def method_signature(self, overriding_name: str = None) -> str:
...
         if isinstance(ret_type, abi.TypeSpec) and abi.contains_type_spec(
-            ret_type,
-            [
-                abi.AccountTypeSpec(),
-                abi.AssetTypeSpec(),
-                abi.ApplicationTypeSpec(),
-            ],
+            ret_type, abi.TransactionTypeSpecs + abi.ReferenceTypeSpecs
         ):
             raise TealInputError("Reference types may not be used as return values")
```
jasonpaulos, line 602: nit: update error message?

pyteal/ast/abi/reference_type.py
barnjamin: you mean create a new ReferenceTypeSpec and ReferenceType and have them inherit from that?

jasonpaulos: Yeah that would work

pyteal/ast/router_test.py
jasonpaulos: Yes, the new feature related to transaction types

pyteal/ast/abi/reference_type.py
barnjamin: I'm not quite sure how to handle this tbh, the over the wire encoding is a uint8 but get returns either uint64 or a bytestring. 

I'm not sure what to do for things like byte_length_static for example, should it be the over the wire encoding or the return type for `get`?

barnjamin: https://github.com/algorand/pyteal/pull/383 moving this set of changes to new pr

tzaffi (approved):
I support the approach of merging as is and addressing outstanding issues in the follow up #383

jasonpaulos (approved):
Looks good! Thanks for adding the additional test and splitting out a new PR for the reference types changes

PR #378, opened 2022-06-02
Change Router contract obj from `Contract` to `Interface`

ahangsu:
Since we are not caring too much about `network` field in `Contract`, returning `Interface` seems to be more appropriate.

### DIFF
--- a/pyteal/ast/router.py
+++ b/pyteal/ast/router.py
@@ -573,7 +573,7 @@ class Router:
             return wrap
         return wrap(func)
 
-    def contract_construct(self) -> sdk_abi.Contract:
+    def contract_construct(self) -> sdk_abi.Interface:
         """A helper function in constructing contract JSON object.
 
         It takes out the method signatures from approval program `ProgramNode`'s,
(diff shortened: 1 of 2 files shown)


jasonpaulos:
I think I prefer to keep this as a contract, the reason being that the router _does_ produce a contract, a concrete application that implements many methods.

What is absent from pyteal is a way to specify interfaces.

jasonpaulos:
@barnjamin if this helps, the idea was that interfaces are sets of method definitions -- by themselves they have no TEAL code. Any program that implements a method with TEAL code must therefore be a contract.

And yes, there is a spot for app IDs in the contract definition, but this is optional. Since at the point of compilation, the contract isn't deployed yet, pyteal can't provide a value for this. But my hope is that you can eventually deploy the TEAL code pyteal generates for you, and once that happens you can easily place the app ID in the contract object that was generated for you as well.

ahangsu:
hmmm, by the discussion I should probably close this pr, and we should keep in mind for further improvement about specifying interface?

PR #380, opened 2022-06-03
Adding method_spec to ABIReturnSubroutine

barnjamin:
Adding a method to ABIReturnSubroutine with names of args and placeholder for description

### DIFF
--- a/pyteal/ast/router.py
+++ b/pyteal/ast/router.py
@@ -6,7 +6,10 @@ from algosdk import abi as sdk_abi
 from algosdk import encoding
 
 from pyteal.config import METHOD_ARG_NUM_CUTOFF
-from pyteal.errors import TealInputError, TealInternalError
+from pyteal.errors import (
+    TealInputError,
+    TealInternalError,
+)
 from pyteal.types import TealType
 from pyteal.compiler.compiler import compileTeal, DEFAULT_TEAL_VERSION, OptimizeOptions
 from pyteal.ir.ops import Mode
@@ -447,6 +450,7 @@ class Router:
         self.approval_ast = ASTBuilder()
         self.clear_state_ast = ASTBuilder()
 
+        self.methods: list[ABIReturnSubroutine] = []
         self.method_sig_to_selector: dict[str, bytes] = dict()
         self.method_selector_to_sig: dict[bytes, str] = dict()
 
@@ -494,6 +498,9 @@ class Router:
                 f"re-registering method {method_signature} has hash collision "
                 f"with {self.method_selector_to_sig[method_selector]}"
             )
+
+        self.methods.append(method_call)
+
         self.method_sig_to_selector[method_signature] = method_selector
         self.method_selector_to_sig[method_selector] = method_signature
 
@@ -583,12 +590,10 @@ class Router:
             contract: a dictified `Contract` object constructed from
                 approval program's method signatures and `self.name`.
         """
-        method_collections = [
-            sdk_abi.Method.from_signature(sig)
-            for sig in self.method_sig_to_selector
-            if isinstance(sig, str)
-        ]
-        return sdk_abi.Contract(self.name, method_collections)
+
+        methods = [sdk_abi.Method.undictify(mc.method_spec()) for mc in self.methods]
+
+        return sdk_abi.Contract(self.name, methods)
 
     def build_program(self) -> tuple[Expr, Expr, sdk_abi.Contract]:
         """
--- a/pyteal/ast/router.py
+++ b/pyteal/ast/router.py
@@ -477,6 +477,7 @@ class Router:
         method_call: ABIReturnSubroutine,
         overriding_name: str = None,
         method_config: MethodConfig = None,
+        description: str = None
     ) -> None:
         if not isinstance(method_call, ABIReturnSubroutine):
             raise TealInputError(
@@ -499,7 +500,10 @@ class Router:
                 f"with {self.method_selector_to_sig[method_selector]}"
             )
 
-        self.methods.append(method_call.method_spec())
+        meth = method_call.method_spec()
+        if description is not None:
+            meth.desc = description
+        self.methods.append(meth)
 
         self.method_sig_to_selector[method_signature] = method_selector
         self.method_selector_to_sig[method_selector] = method_signature
@@ -525,6 +529,7 @@ class Router:
         clear_state: CallConfig = None,
         update_application: CallConfig = None,
         delete_application: CallConfig = None,
+        description: str = None,
     ):
         """
         A decorator style method registration by decorating over a python function,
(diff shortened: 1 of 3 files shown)


pyteal/ast/subroutine.py
```diff
@@ -611,6 +611,23 @@ def method_signature(self, overriding_name: str = None) -> str:
...
+                "name": name,
+            }
+            for name, val in self.subroutine.annotations.items()
+            if name not in skip_names
+        ]
+        return {
+            "name": self.name(),
+            "args": args,
+            "returns": {"type": str(self.type_of())},
+            "desc": "todo",  # Allow decorator to accept desc? Repurpose doc string?
```
barnjamin, line 628: what do?

pyteal/ast/router_test.py
```diff
@@ -262,74 +262,74 @@ def test_call_config():
...
-                        )
-                    case 0:
-                        continue
-                    case 1:
-                        list_of_expressions.append(pt.Txn.on_completion() == oc)
-            with pt.TealComponent.Context.ignoreExprEquality():
-                assert assemble_helper(mc.approval_cond()) == assemble_helper(
-                    pt.Or(*list_of_expressions)
-                )
+# def test_method_config():
```
barnjamin, line 265: I'll put this back before merging

pyteal/ast/subroutine.py
ahangsu: read the doc string from the python function?

tzaffi: We could have the best of both worlds for that comment - if it's not too much work
Eg, consider:

```python
@router.method
def foo():
    """
    This is a wonderful method
    """
    ...
```
It could generate

```python
    "description": "This is a wonderful method"
```

and similarly 

```python
@router.method(decription="This is a wonderful method")
def foo():
    ...
```

would generate the same. I guess we'd need to decide what do about

```python
@router.method(decription="This is a wonderful method")
def foo():
    "this is a terrible method - don't call it!"
    ...
```

As a sidenote, I advocate using `description` as opposed to `desc`

barnjamin: i think if specified in decorator it should override doc string

noted about the desc/description 👍

ahangsu: > i think if specified in decorator it should override doc string

That makes sense.

pyteal/ast/subroutine.py
```diff
@@ -611,6 +611,23 @@ def method_signature(self, overriding_name: str = None) -> str:
             overriding_name = self.name()
         return f"{overriding_name}({','.join(args)}){self.type_of()}"
 
+    def method_spec(self) -> dict:
```
jasonpaulos, line 614: nit: it would be nice to return the result of `sdk_abi.Method.undictify`, i.e. a `Method` object instead. That's the ultimate target of this function anyway, and it makes the types clearer

pyteal/ast/router.py
```diff
@@ -447,6 +450,7 @@ def __init__(
         self.approval_ast = ASTBuilder()
         self.clear_state_ast = ASTBuilder()
 
+        self.methods: list[ABIReturnSubroutine] = []
```
ahangsu, line 453: nit: would be better to maintain a list of `sdk_abi.Method` than a list of `ABIReturnSubroutine`.

pyteal/ast/subroutine.py
barnjamin: @tzaffi `description` doesnt appear to work: https://github.com/algorand/py-algorand-sdk/blob/master/algosdk/abi/method.py#L212

tzaffi: Since Abi hasn't been merged I advocate for changing desc to description everywhere in ABI

barnjamin: Looks like its the same  `desc` across all of these, unless we have a really good reason to change it across all the SDKs I'm inclined to leave it as is

https://github.com/algorand/go-algorand-sdk/blob/develop/abi/interactions.go#L140
https://github.com/algorand/js-algorand-sdk/blob/develop/src/abi/method.ts#L149
https://github.com/algorand/java-algorand-sdk/blob/develop/src/main/java/com/algorand/algosdk/abi/Method.java#L80

pyteal/ast/subroutine.py
```diff
@@ -611,6 +613,29 @@ def method_signature(self, overriding_name: str = None) -> str:
...
+        ]
+
+        spec = {
+            "name": self.name(),
+            "args": args,
+            "returns": {"type": str(self.type_of())},
+        }
+
+        if self.subroutine.implementation.__doc__ is not None:
+            spec["desc"] = self.subroutine.implementation.__doc__
```
ahangsu, line 635: nit: a small improvement for preprocessing `__doc__`:
```python3
spec["desc"] = " ".join([i.strip() for i in spec["desc"].split("\n") if not (i.isspace() or len(i) == 0)])
```
and an example would be
```python3
>>> def func(a):
...     """
...     this is a docstring
...     with breaklines
...     """
...
>>> " ".join([i.strip() for i in func.__doc__.split("\n") if not (i.isspace() or len(i) == 0)])
'this is a docstring with breaklines'
```

pyteal/ast/subroutine.py
tzaffi: NVM. I see `desc` is built-in to ARC-4. Resolving.

pyteal/ast/router.py
```diff
@@ -494,6 +499,12 @@ def add_method_handler(
                 f"re-registering method {method_signature} has hash collision "
                 f"with {self.method_selector_to_sig[method_selector]}"
             )
+
+        meth = method_call.method_spec()
```
ahangsu, line 503: lmao love this name, Heisenberg's stuff is tight!

ahangsu (approved):

tzaffi (commented):
Looks good, but would love to have some unit tests.

barnjamin:
> Looks good, but would love to have some unit tests.

Added another set of asserts in the existing test to compare against the method generated from sig && the one we're outputing from ABIReturnSubroutine

tzaffi:
> > Looks good, but would love to have some unit tests.
> 
> Added another set of asserts in the existing test to compare against the method generated from sig && the one we're outputing from ABIReturnSubroutine

Some more end-to-end'ish tests showing the different behavior when using `description` as an argument of `@router.method` vs what's picked up from the doc-string would be great too.

pyteal/ast/router_test.py
```diff
@@ -528,8 +530,29 @@ def test_contract_json_obj():
...
+
+        ms = subroutine.method_spec()
+
+        # Manually replace it since the override is applied in the method handler
+        # not attached to the ABIReturnSubroutine itself
+        ms.desc = desc if desc is not None else ms.desc
+
+        sig_method = sdk_abi.Method.from_signature(subroutine.method_signature())
+
+        assert ms.name == sig_method.name
```
tzaffi, line 549: shouldn't there be an assertion involving  `sig_method.desc` ?

barnjamin: sig_method doesnt get a description since it just parses out the `methodname(arg,arg)`

tzaffi (approved):
LGTM

jasonpaulos (approved):
Makes sense

PR #382, opened 2022-06-06
Implementation of signed integers and fixed point numbers

CiottiGiorgio:
In this PR, we should discuss the implementation of the types.

My main concerns are:

- How to make sure that creating a custom ABI type starting from native ABI types makes sense.
- How to define operations on ABI Types and make sure that it's not possible to use operations with Types that don't respect operation signature. (I'm using the term Operation because I don't want to fix it to a function, class method, Subroutine, ...)
- What properties to test on custom ABI Types.

At this specific moment, only a primitive implementation of signed integers and fixed point numbers is present. That's because the actual heart of this project relies on implementing arithmetic that cannot be mapped to native TEAL. It's not 100% clear to me how to (safely) implement arithmetic in a way that feels native to PyTeal but not to TEAL. My attempt was to leverage Uint machinery, reduce new code to the minimum, avoid duplicated features.


barnjamin:
make sure to run run `black` to format and `flake8 pyteal` to  lint and `mypy pyteal` for type checking stuff

barnjamin:
It might seem against DRY but I think you can copy/paste the methods instead of inheriting. A lil pasta is better than a lil dependency in this case 🤝

tzaffi:
> It might seem against DRY but I think you can copy/paste the methods instead of inheriting. A lil pasta is better than a lil dependency in this case 🤝

I love 🍝 (that's pasta _not_ spaghetti). But seriously, even on theoretical grounds it's arguable that `SInt` is not a `Uint` (for example the invariant $x \geq 0$ fails for SInt).

pyteal/ast/abi/int.py
```diff
@@ -0,0 +1,185 @@
...
+        return SInt64()
+
+    def annotation_type(self) -> "type[SInt64]":
+        return SInt64
+
+
+SInt64TypeSpec.__module__ = "pyteal"
+
+
+T = TypeVar("T", bound="Int")
```
jasonpaulos, line 88: I think this should be `SInt`
```suggestion
T = TypeVar("T", bound="SInt")
```

jasonpaulos (commented):
Another option is to create a new abstract subclass of `BaseType` that both `UInt` and `SInt` can inherit from. It could be called `NumericType` or similar and contain all fields/methods common to both types, such as bit size.

This would be more involved since it touches `Uint` code as well, but it might be the best choice long term.

FrankSzendzielarz:
Signed arithmetic results in a lot of opcodes for things Remainder (modulo) operation. It might be worth considering compiler flags too for inlining this stuff or not.

CiottiGiorgio:
Closing this branch as this topic will be picked up later in the future

PR #383, opened 2022-06-06
define reference type spec

barnjamin:
Define ReferenceTypeSpec and ReferenceType

Each ReferenceType implementation provides params and a way to access the underlying identifier (asset id, app id, address) as well as reference index (thanks @jasonpaulos !)


jasonpaulos (approved):
My approval doesn't count for much since I worked on a child PR that went into this, but I'm giving it anyway

pyteal/ast/acct.py
```diff
@@ -55,3 +57,33 @@ def authAddr(cls, acct: Expr) -> MaybeValue:
...
+
+
+class AccountParamObject:
+    """Represents information about an account"""
+
+    def __init__(self, account: Expr) -> None:
+        """Create a new AccountParamObject for the given account.
+
+        Args:
+            app: An index into Txn.accounts that corresponds to the application to check or an
```
tzaffi, line 69:
```suggestion
            account: An index into Txn.accounts that corresponds to the application to check or an
```
Not sure if this should be changed or the parameter should be renamed to accomodate the comment

tzaffi: Reading more of the PR, it seems like this is just a copy/paste error and `account` makes sense here.

pyteal/ast/app.py
```diff
@@ -356,3 +356,56 @@ def address(cls, app: Expr) -> MaybeValue:
...
+
+    def global_num_unit(self) -> MaybeValue:
+        """Get the number of uint64 values allowed in Global State for the application."""
+        return AppParam.globalNumUnit(self._app)
+
+    def global_num_byte_slice(self) -> MaybeValue:
+        """Get the number of byte array values allowed in Global State for the application."""
+        return AppParam.globalNumByteSlice(self._app)
+
+    def local_num_unit(self) -> MaybeValue:
```
tzaffi, line 390:
```suggestion
    def global_num_uint(self) -> MaybeValue:
        """Get the number of uint64 values allowed in Global State for the application."""
        return AppParam.globalNumUnit(self._app)

    def global_num_byte_slice(self) -> MaybeValue:
        """Get the number of byte array values allowed in Global State for the application."""
        return AppParam.globalNumByteSlice(self._app)

    def local_num_uint(self) -> MaybeValue:
```

Unfortunately, these method names are misspelled in the `AppParam` class, but there's not need to perpetuate the type in a brand new class.

I'm suggesting `uint` (which aligns with **algod**) as opposed to `int` (which aligns with **goal**)

tzaffi (changes requested):
LGTM except for a suggested typo fix.

pyteal/ast/app.py
```diff
@@ -356,3 +356,56 @@ def address(cls, app: Expr) -> MaybeValue:
...
+    def __init__(self, app: Expr) -> None:
+        """Create a new AppParamObject for the given application.
+
+        Args:
+            app: An identifier for the app. It must be an index into Txn.ForeignApps that
+                corresponds to the app to check, or since v4, an application ID that appears in
+                Txn.ForeignApps or is the CurrentApplicationID. In either case, it must evaluate to
+                uint64.
+        """
+        self._app: Final = app
```
ahangsu, line 373: nit: do we want to type check here, for `_app` must be evaluated to `uint64` by comment?

pyteal/ast/asset.py
```diff
@@ -51,6 +53,39 @@ def frozen(cls, account: Expr, asset: Expr) -> MaybeValue:
...
+            asset: An identifier for the asset. It must be an index into Txn.ForeignAssets that
+                corresponds to the asset to check, or since v4, an asset ID that appears in
+                Txn.ForeignAssets. In either case, it must evaluate to uint64.
+            account: An identifier for the account. It must be an index into Txn.Accounts that
+                corresponds to the account to check (in which case it must evaluate to uint64), or
+                since v4, an account address that appears in Txn.Accounts or is Txn.Sender (in which
+                case it must evaluate to bytes).
+        """
+        self._asset: Final = asset
+        self._account: Final = account
```
ahangsu, line 72: if the previous assumption on type check is correct, I think this place also need some type check?

pyteal/ast/asset.py
```diff
@@ -259,3 +294,70 @@ def creator(cls, asset: Expr) -> MaybeValue:
...
+
+    def __init__(self, asset: Expr) -> None:
+        """Create a new AssetParamObject for the given asset.
+
+        Args:
+            asset: An identifier for the asset. It must be an index into Txn.ForeignAssets that
+                corresponds to the asset to check, or since v4, an asset ID that appears in
+                Txn.ForeignAssets. In either case, it must evaluate to uint64.
+        """
+        self._asset: Final = asset
```
ahangsu, line 310: same here?

ahangsu (commented):
just a few comments on type checking

ahangsu (approved):

PR #384, opened 2022-06-06
Add handler for allowing Txna in InnerTxnBuilder

barnjamin:
Often folks will want to just pass the full list included in a Transaction to an inner transaction, this should be easier than writing a for loop themselves

This should allow:
```py
InnerTxnBuilder.SetFields({
  # ...
  TxnField.accounts: Txn.accounts,
  # ...
})
```

Alternatively, if they know how many accounts it'll be exactly, we should be able to unroll the loop


pyteal/ast/itxn.py
```diff
@@ -158,33 +168,42 @@ def SetField(cls, field: TxnField, value: Expr | list[Expr]) -> Expr:
...
+                        InnerTxnFieldExpr(field, cast(Expr, valueIter))
+                        for valueIter in value
+                    ]
+                )
+            else:
+                return For(
+                    (i := ScratchVar()).store(Int(0)),
+                    i.load() < cast(TxnArray, value).length(),
+                    i.store(i.load() + Int(1)),
+                ).Do(InnerTxnFieldExpr(field, i.load()))
```
ahangsu, line 206: hold up, these lines seem off: I assume you want to put everything in `value`, which is a `TxnArray`, in `itxn_field` setting. But it looks like you are setting 0,1,2... to itxn_field, which I am not sure if it is making sense.

barnjamin: i'll need to actually test it against the avm but i _think_ you can reference the relevant array by index and it knows where to look. I'm not positive though

barnjamin: > algosdk.error.AlgodHTTPError: 169: itxn_field Accounts arg 0 wanted type []byte got uint64

no worky :(

ahangsu: yikes

barnjamin:
Just for fun, heres the test i wrote w/ beaker
```py
from pyteal import *
from beaker import *
from beaker.application import get_method_selector

class Testy(Application):

    @handler
    def handle_outer(self, acct1: abi.Account, acct2: abi.Account, app: abi.Application):
        return Seq(
            InnerTxnBuilder.Begin(),
            InnerTxnBuilder.SetFields({
                TxnField.type_enum: TxnType.ApplicationCall,
                TxnField.application_id: app.application_id(),
                TxnField.accounts: Txn.accounts,
                TxnField.application_args: [Bytes(get_method_selector(self.handle_inner))],
            })
        )

    @handler
    def handle_inner():
        return Assert(Txn.accounts.length()>Int(0))

if __name__ == "__main__":
    from algosdk.atomic_transaction_composer import AccountTransactionSigner
    from beaker import sandbox, client


    accts = sandbox.get_accounts()
    addr, sk = accts.pop()
    signer = AccountTransactionSigner(sk)

    addr1, _ = accts.pop() 
    addr2, _ = accts.pop()

    a1 = Testy()
    outer_client = client.ApplicationClient(sandbox.get_client(), a1, signer=signer)
    outer_client.create()

    a2 = Testy()
    inner_client = client.ApplicationClient(sandbox.get_client(), a2, signer=signer)
    inner_client.create()

    result = outer_client.call(a1.handle_outer, acct1=addr1, acct2=addr2, app=inner_client.app_id)
    print(result.return_value)
```

ahangsu:
looks cool, but the test script seems not have submitted the itxn?

barnjamin:
> looks cool, but the test script seems not have submitted the itxn?
lol, updated.

Also if i took the time point my SB to master go-algorand i think I could omit the pay and instead set suggested params to flat fee and fee = 2*millialgo

ahangsu:
> Also if i took the time point my SB to master go-algorand i think I could omit the pay and instead set suggested params to flat fee and fee = 2*millialgo

That surprises me, wow we can actually do the pooled payment with suggested-params.

ahangsu (approved):

barnjamin:
> That surprises me, wow we can actually do the pooled payment with suggested-params

Just confirmed it does work:

```py

if __name__ == "__main__":
    from algosdk.atomic_transaction_composer import AccountTransactionSigner
    from algosdk.future import transaction
    from beaker import sandbox, client, consts

    algod_client = sandbox.get_client()

    accts = sandbox.get_accounts()
    addr, sk = accts.pop()
    signer = AccountTransactionSigner(sk)

    addr1, _ = accts.pop()
    addr2, _ = accts.pop()

    # Create first app
    a1 = Testy()
    outer_client = client.ApplicationClient(algod_client, a1, signer=signer)
    _, app_addr, _ = outer_client.create()

    # Create second app
    a2 = Testy()
    inner_client = client.ApplicationClient(algod_client, a2, signer=signer)
    inner_client.create()

    # try to call second from first
    sp = algod_client.suggested_params()
    sp.flat_fee = True
    sp.fee = 2 * consts.milli_algo
    result = outer_client.call(
        a1.handle_outer, suggested_params=sp, acct1=addr1, acct2=addr2, app=inner_client.app_id
    )
    print("works!")
    
  ```

**Comment by ahangsu** [comment]
the merge master issue also apply here

File: pyteal/ast/itxn_test.py [hunk 2]
```diff
@@ -128,7 +138,7 @@ def test_InnerTxnBuilder_SetFields():
         actual.addIncoming()
         actual = pt.TealBlock.NormalizeBlocks(actual)
 
-        with pt.TealComponent.Context.ignoreExprEquality():
+        with pt.TealComponent.Context.ignoreScratchSlotEquality(), pt.TealComponent.Context.ignoreExprEquality():
```
jasonpaulos, line 141: If you compare with `TealComponent.Context.ignoreScratchSlotEquality()`, please follow with a call to `TealBlock.MatchScratchSlotReferences`.

Example: https://github.com/algorand/pyteal/blob/79df68ba0d02fd56c242cdaaf84929fe7908b003/pyteal/ast/abi/array_base_test.py#L157-L160

jasonpaulos (commented):
Functionality looks good.

Can you add the new behavior to the `SetField` and `SetFields` docstrings?

pyteal/ast/itxn_test.py
```diff
@@ -152,7 +167,7 @@ def test_InnerTxnBuilder_Execute():
...
```
jasonpaulos, line 170: Same comment here about `TealBlock.MatchScratchSlotReferences` 🙂

jasonpaulos (commented):
Just one holdup

pyteal/ast/itxn_test.py
barnjamin: doh!

jasonpaulos (approved):

jasonpaulos:
I already approved, but if you'd like to add a CHANGELOG entry, that would be appreciated

PR #385, opened 2022-06-06
Create params and holding objects for references

jasonpaulos:
Child PR of #383

I rewrote part of the reference type implementation to add these methods:
* Common to all:
    * `referenced_index()` - get the foreign array index
* `Account`:
    * `address()` - get the address of the account
    * `params()` - get information about the account, returns the new `AccountParamObject`
    * `asset_holding(asset)` - get information about an asset held by this account, returns the new `AssetHoldingObject`
* `Asset`:
    * `asset_id()` - get the ID of the asset
    * `holding(account)` - get information about this asset held by an account, returns the new `AssetHoldingObject`
    * `params()` - get information about the asset's parameters, returns the new `AssetParamObject`
* `Application`:
    * `application_id()` - get the ID of the app
    * `params()` - get information about the app's parameters, returns the new `AppParamObject`

`AccountParamObject`, `AssetHoldingObject`,  `AssetParamObject`, and `AppParamObject` are all classes which expose the same fields as `AccountParam`, `AssetHolding`, `AssetParam`, and `AppParam`, respectively. The difference is that they take their input during class creation, not when calling each field.

Here's an example of the new field usage:

```python
@ABIReturnSubroutine
def is_sender_the_asset_creator_1(asset: abi.Asset, *, output: abi.Bool) -> Expr:
    # using the new params() method, which returns an AssetParamObject
    return asset.params().creator_address().outputReducer(lambda value, has_value: output.set(value == Txn.sender()))

@ABIReturnSubroutine
def is_sender_the_asset_creator_2(asset: abi.Asset, *, output: abi.Bool) -> Expr:
    # using the existing AssetParam class
    return AssetParam.creator(asset.asset_id()).outputReducer(lambda value, has_value: output.set(value == Txn.sender()))
```

The benefit of the new fields is that this is a more object oriented approach, and it should help users become aware of how reference types can be used.


barnjamin:
I like it

Can we do just `id` instead of `Application.application_id()`? same for asset?

cusma:
I like that too! 😎

pyteal/ast/abi/reference_type.py
```diff
@@ -57,37 +46,16 @@ def __init__(self, spec: ReferenceTypeSpec) -> None:
     def type_spec(self) -> ReferenceTypeSpec:
         return cast(ReferenceTypeSpec, super().type_spec())
 
-    def get(self) -> Expr:
-        return self.stored_value.load()
+    def referenced_index(self) -> Expr:
+        """Get the reference index for this value.
 
-    def set(
```
jasonpaulos, line 63: I've chosen to remove the `set` method for now because it's not used or needed currently. Reference types aren't supposed to be user-created, so if we want to change that we'll have to revisit.

barnjamin (approved):

PR #386, opened 2022-06-07
Changes to router clear state

jasonpaulos:
This PR introduces the following changes to the `Router`'s handling of clear state actions:
* `CallConfig.CREATE` is now an invalid config for clear state, since the clear state program can never be invoked during creation.
* `CallConfig.ALL` and `CallConfig.CALL` now mean the same thing for clear state. Notably, the router does not generate checks to ensure that `Txn.application_id() != Int(0)` anymore, since this will always be true during clear state.
* I found it odd that the router does not allow programs to be built that specify no clear state actions or methods, so I added a default action of `Reject()` if no actions or methods have been added to the router. This affects approval programs as well.

### DIFF
--- a/pyteal/ast/router.py
+++ b/pyteal/ast/router.py
@@ -29,7 +29,7 @@ from pyteal.ast.seq import Seq
 from pyteal.ast.methodsig import MethodSignature
 from pyteal.ast.naryexpr import And, Or
 from pyteal.ast.txn import Txn
-from pyteal.ast.return_ import Approve
+from pyteal.ast.return_ import Approve, Reject
 
 
 class CallConfig(IntFlag):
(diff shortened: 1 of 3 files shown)


pyteal/ast/router.py
```diff
@@ -475,7 +479,7 @@ def add_method_to_ast(
 
     def program_construction(self) -> Expr:
         if not self.conditions_n_branches:
-            raise TealInputError("ABIRouter: Cannot build program with an empty AST")
+            return Reject()
```
tzaffi, line 482: I recommend adding commentary to `build_program()` and/or `compile_program()` which points out that if there is no AST or alternately if no methods or bare calls have been added, a trivial rejecting program will be constructed.

tzaffi (approved):
LGTM

ahangsu (approved):

PR #387, opened 2022-06-07
Method call itxn

barnjamin:
Not sure of the best way to handle Transactions as arguments and whether we should force them to be passed in to the argument list or require they are _not_ passed in. Both are possible but looking for best ux for devs

Had to move method_return for circular import stuff?

Added some tests, will add more as needed


pyteal/ast/abi/util.py
```diff
@@ -379,3 +389,87 @@ def algosdk_from_type_spec(t: TypeSpec) -> algosdk.abi.ABIType:
...
+    from pyteal.ast.abi.uint import (
+        ByteTypeSpec,
+        Uint8TypeSpec,
+        Uint16TypeSpec,
+        Uint32TypeSpec,
+        Uint64TypeSpec,
+    )
+
+    match t:
+        case str():
```
ahangsu, line 414: nit: would be better if you add a comment about sdk do not currently support reference types and transaction types, and they are handled only by `str`

pyteal/ast/itxn.py
```diff
@@ -200,6 +206,104 @@ def SetFields(cls, fields: Dict[TxnField, Union[Expr, List[Expr]]]) -> Expr:
...
+
+                    case abi.ApplicationTypeSpec():
+                        if isinstance(arg, Expr):
+                            apps.append(arg)
+                        elif isinstance(arg, abi.Application):
+                            apps.append(arg.application_id())
+                        else:
+                            raise TealTypeError(arg, Union[abi.Application, Expr])
+
+                        app_args.append(Suffix(Itob(Int(len(apps))), Int(7)))
```
ahangsu, line 269: nit: would be good to explain this suffix takes the last byte, which is a uint8 value for reference type value encoding

ahangsu (commented):
looks largely good to me, just some minor comments

pyteal/ast/abi/util.py
```diff
@@ -1,4 +1,14 @@
...
+from typing import (
+    Sequence,
+    TypeVar,
+    Any,
+    Literal,
+    List,
+    Union,
+    get_origin,
+    get_args,
+    cast,
```
tzaffi, line 10:
```suggestion
    Any,
    List,
    Literal,
    Sequence,
    TypeVar,
    Union,
    cast,
    get_args,
    get_origin,
```

nit: ocd request to sort the imports

pyteal/ast/abi/util_test.py
```diff
@@ -593,3 +594,11 @@ def test_abi_type_translation(algosdk_abi, abi_string, pyteal_abi_ts, pyteal_abi
     )
     assert algosdk_abi == abi.algosdk_from_type_spec(pyteal_abi_ts)
     assert algosdk_abi == abi.algosdk_from_annotation(pyteal_abi)
+
+
+def test_sdk_abi_translation():
+    for case in ABI_TRANSLATION_TEST_CASES:
+        # Errors are strings in the 0th element
+        if type(case[0]) == str:
```
tzaffi, line 602:
```suggestion
        if type(case[0]) is str:
```
Types are singleton's so prefer `is`

pyteal/ast/abi/util_test.py
```diff
@@ -593,3 +594,11 @@ def test_abi_type_translation(algosdk_abi, abi_string, pyteal_abi_ts, pyteal_abi
     )
     assert algosdk_abi == abi.algosdk_from_type_spec(pyteal_abi_ts)
     assert algosdk_abi == abi.algosdk_from_annotation(pyteal_abi)
+
+
+def test_sdk_abi_translation():
+    for case in ABI_TRANSLATION_TEST_CASES:
```
tzaffi, line 600:
```suggestion
@pytest.mark.parametrize("case", ABI_TRANSLATION_TEST_CASES)
def test_sdk_abi_translation(case):
```
indentation will be a bit off, but by separating out the tests this way, it will be easier to tell in CI which particular test case failes

ahangsu: that's a good one ☝️

pyteal/ast/abi/util.py
```diff
@@ -379,3 +389,88 @@ def algosdk_from_type_spec(t: TypeSpec) -> algosdk.abi.ABIType:
...
+                    return StringTypeSpec()
+                case algosdk.abi.AddressType():
+                    return AddressTypeSpec()
+                case algosdk.abi.UfixedType():
+                    raise TealInputError("Ufixed not supported")
+
+    raise TealInputError(f"Invalid Type: {t}")
+
+
+def type_specs_from_signature(sig: str) -> tuple[List[TypeSpec], TypeSpec]:
```
tzaffi, line 472:
```suggestion
def type_specs_from_signature(sig: str) -> tuple[list[TypeSpec], TypeSpec]:
```
nit: prefer 3.10 usage

tzaffi: IMO this one deserves its own unit test with some actual method signature examples. I suspect it might barf on "void" but a unit test could prove that I'm wrong.

pyteal/ast/router.py
```diff
@@ -4,6 +4,7 @@
 
 from algosdk import abi as sdk_abi
 from algosdk import encoding
+from pyteal.ast.method_return import MethodReturn
 
 from pyteal.config import METHOD_ARG_NUM_CUTOFF
```
tzaffi, line 9:
```suggestion
from algosdk import encoding

from pyteal.ast.method_return import MethodReturn
from pyteal.config import METHOD_ARG_NUM_CUTOFF
```
nit: import grouping

pyteal/ast/itxn.py
```diff
@@ -200,6 +206,111 @@ def SetFields(cls, fields: Dict[TxnField, Union[Expr, List[Expr]]]) -> Expr:
         fieldsToSet = [cls.SetField(field, value) for field, value in fields.items()]
         return Seq(fieldsToSet)
 
+    @classmethod
+    def MethodCall(
+        cls,
+        *,
+        app_id: Expr,
```
tzaffi, line 213: Are any of these arguments mandatory? It ought to be the case that `app_id` is since how else would you know which app to call and re-entry is disallowed.

pyteal/ast/itxn.py
```diff
@@ -200,6 +206,111 @@ def SetFields(cls, fields: Dict[TxnField, Union[Expr, List[Expr]]]) -> Expr:
         fieldsToSet = [cls.SetField(field, value) for field, value in fields.items()]
         return Seq(fieldsToSet)
 
+    @classmethod
+    def MethodCall(
```
tzaffi, line 210: Worthwhile having some kind of unit test or even a graviton test that shows how to use it. It could be based on: 
image

I suspect that graviton will fall flat because of inner txn's, but would love to be proved wrong or at least have a good example to aim for when inner txn's are supported.

pyteal/ast/itxn.py
tzaffi: NVM - this is already being enforced

pyteal/ast/itxn.py
```diff
@@ -200,6 +206,111 @@ def SetFields(cls, fields: Dict[TxnField, Union[Expr, List[Expr]]]) -> Expr:
...
+    ) -> Expr:
+
+        # Default, always need these
+        fields_to_set = [
+            cls.SetField(TxnField.type_enum, TxnType.ApplicationCall),
+            cls.SetField(TxnField.application_id, app_id),
+        ]
+
+        # We only care about the args
+        arg_type_specs, _ = type_specs_from_signature(method_signature)
```
tzaffi, line 226:
```suggestion
        (arg_type_specs: list[TypeSpec], _: TypeSpec) = type_specs_from_signature(method_signature)
```
something like this ought to work, and then there'll be consistency with typing local var's

tzaffi: Probably need to define on a separate line:

```python
arg_type_specs: list[TypeSpec]
```

pyteal/ast/itxn.py
```diff
@@ -200,6 +206,111 @@ def SetFields(cls, fields: Dict[TxnField, Union[Expr, List[Expr]]]) -> Expr:
...
+        # Reference types are encoded as a uint8 index into
+        # their respective arrays
+        def get_uint8(i):
+            return Suffix(Itob(Int(i)), Int(7))
+
+        for idx, arg_ts in enumerate(arg_type_specs):
+            arg = args[idx]
+
+            if arg_ts in abi.TransactionTypeSpecs:
+                if not isinstance(arg, Dict):
```
tzaffi, line 248:
```suggestion
                if not isinstance(arg, dict):
```
just a nit

pyteal/ast/itxn.py
```diff
@@ -200,6 +206,111 @@ def SetFields(cls, fields: Dict[TxnField, Union[Expr, List[Expr]]]) -> Expr:
...
+        # Start app args with the method selector
+        app_args: List[Expr] = [MethodSignature(method_signature)]
+
+        # Transactions are not included in the App Call
+        txns_to_pass: List[Expr] = []
+
+        # Reference Types are treated specially
+        accts: List[Expr] = []
+        apps: List[Expr] = []
+        assets: List[Expr] = []
```
tzaffi, line 237:
```suggestion
        app_args: list[Expr] = [MethodSignature(method_signature)]

        # Transactions are not included in the App Call
        txns_to_pass: List[Expr] = []

        # Reference Types are treated specially
        accts: list[Expr] = []
        apps: list[Expr] = []
        assets: list[Expr] = []
```

nit 3.10 style

pyteal/ast/itxn.py
```diff
@@ -200,6 +206,111 @@ def SetFields(cls, fields: Dict[TxnField, Union[Expr, List[Expr]]]) -> Expr:
         fieldsToSet = [cls.SetField(field, value) for field, value in fields.items()]
         return Seq(fieldsToSet)
 
+    @classmethod
+    def MethodCall(
+        cls,
+        *,
+        app_id: Expr,
+        method_signature: str,
+        args: List[Union[abi.BaseType, Expr, Dict[TxnField, Union[Expr, List[Expr]]]]],
```
tzaffi, line 215:
```suggestion
        args: list[abi.BaseType | Expr | dict[TxnField, (Expr | list[Expr])]]]],
```
not sure about the last `(Expr | list[Expr])`

3.10 suggestion

pyteal/ast/itxn.py
```diff
@@ -200,6 +206,111 @@ def SetFields(cls, fields: Dict[TxnField, Union[Expr, List[Expr]]]) -> Expr:
...
+                        elif isinstance(arg, abi.Application):
+                            apps.append(arg.application_id())
+                        else:
+                            raise TealTypeError(arg, Union[abi.Application, Expr])
+
+                        app_args.append(get_uint8(len(apps)))
+
+                    # For assets, add to app_args prior to appending to assets array
+                    case abi.AssetTypeSpec():
+                        app_args.append(get_uint8(len(assets)))
```
tzaffi, line 279: Nice ARC-4 nuance to treat `assets` index differently from `accounts` and `applications` !!!

pyteal/ast/itxn.py
```diff
@@ -200,6 +206,111 @@ def SetFields(cls, fields: Dict[TxnField, Union[Expr, List[Expr]]]) -> Expr:
...
+        txns_to_pass: List[Expr] = []
+
+        # Reference Types are treated specially
+        accts: List[Expr] = []
+        apps: List[Expr] = []
+        assets: List[Expr] = []
+
+        # Reference types are encoded as a uint8 index into
+        # their respective arrays
+        def get_uint8(i):
```
tzaffi, line 241: Feel free to ignore. Seems like this would be useful helper function. Maybe a class method of `class Int`? This may be shot done by others as it seems to break our typical patterns, but here's what I could see:

```python
class Int(LeafExpr):
    ...

    @classmethod
    def as_uint8(cls, i: int) -> Expr:
        return Suffix(Itob(cls(i)), Int(7))
```

tzaffi: or maybe just an instance method:

```python
class Int(LeafExpr):
    ...

    def as_uint8(self) -> Expr:
        return Suffix(Itob(self), Int(7))
```

tzaffi (commented):
I think this is a very helpful PR.

I haven't found any problems, but pointed out some small nit's. I do think we could benefit from unit or integration tests that I mentioned specifically.

pyteal/ast/abi/util.py
tzaffi: you may not actually need all of these after converting the `typing` types to concrete types as suggested below and which conforms with 3.10 usage.

pyteal/ast/itxn.py
barnjamin: I'd almost prefer to put this kind of thing as a class method on the ABI type itself.

something like:
```py
def Uint(BaseType):
    ...
    def to_encoded(cls, i: Int)->Expr:
        return Suffix(Itob(i), Int(8-cls.byte_length_static()))
```

Or tweaking/reimplementing the uint_* helper methods in uint.py to not need some scratch var

pyteal/ast/abi/util.py
barnjamin: good call,  but what do for that case 🤔 do we need a new TypeSpec to represent it or?

barnjamin: punted, returning None for now

pyteal/ast/abi/util_test.py
```diff
@@ -593,3 +651,18 @@ def test_abi_type_translation(algosdk_abi, abi_string, pyteal_abi_ts, pyteal_abi
...
+@pytest.mark.parametrize("case", ABI_TRANSLATION_TEST_CASES)
+def test_sdk_abi_translation(case):
+    # Errors are strings in the 0th element
+    if type(case[0]) is str:
+        return
+    assert type_spec_from_algosdk(case[0]) == case[2]
+
+
+@pytest.mark.parametrize("sig_str, sig_args, sig_rets", ABI_SIGNATURE_TYPESPEC_CASES)
+def test_sdk_type_specs_from_signature(sig_str, sig_args, sig_rets):
```
tzaffi, line 665: saweeeeeeet

pyteal/ast/abi/util.py
tzaffi: IMO you've solved the problem (`Optional[TypeSpec]` or equivalently you can use `TypeSpec | None`). So it's not a punt, it's a field goal! 

🏈

pyteal/ast/itxn.py
tzaffi: I don't have a strong opinion on this, so leave at your discretion. Others might have stronger opinions though.

pyteal/ast/abi/util.py
```diff
@@ -379,3 +389,91 @@ def algosdk_from_type_spec(t: TypeSpec) -> algosdk.abi.ABIType:
...
+                    raise TealInputError("Ufixed not supported")
+
+    raise TealInputError(f"Invalid Type: {t}")
+
+
+def type_specs_from_signature(sig: str) -> tuple[list[TypeSpec], Optional[TypeSpec]]:
+    sdk_method = algosdk.abi.Method.from_signature(sig)
+
+    return_type = None
+    if sdk_method.returns.type != "void":
```
tzaffi, line 476:
```suggestion
    if sdk_method.returns.type != algosdk.abi.Returns.VOID:
```

nit: just a bit safer

pyteal/ast/abi/util_test.py
```diff
@@ -552,6 +554,62 @@ def test_size_of():
     ),
 ]
 
+ABI_SIGNATURE_TYPESPEC_CASES = [
```
ahangsu, line 557: would be nice if you add a case returning void

pyteal/ast/itxn_test.py
```diff
@@ -134,4 +135,125 @@ def test_InnerTxnBuilder_SetFields():
                 expr.__teal__(teal4Options)
 
 
+def test_InnerTxnBuilder_method_call():
+    cases = (
```
ahangsu, line 139: nit: would be nice if it is possible to follow `@pytest.mark.parametrize`? (feel free to ignore if it is hard to do)

pyteal/ast/itxn.py
```diff
@@ -200,6 +205,108 @@ def SetFields(cls, fields: Dict[TxnField, Union[Expr, List[Expr]]]) -> Expr:
...
+                    # For assets, add to app_args prior to appending to assets array
+                    case abi.AssetTypeSpec():
+                        app_args.append(uint_encode(8, Int(len(assets))))
+
+                        if isinstance(arg, Expr):
+                            assets.append(arg)
+                        elif isinstance(arg, abi.Asset):
+                            assets.append(arg.asset_id())
+                        else:
+                            raise TealTypeError(arg, abi.Asset | Expr)
```
ahangsu, line 281: that is a smart move

ahangsu (commented):
look largely good to me

ahangsu (approved):
I don't have too much comments out there, so I just approve. But keep an eye on folks, might have good insights and comments.

tzaffi (approved):
LGTM - glad we have this functionality now.

jdtzmn (approved):

pyteal/ast/itxn.py
```diff
@@ -200,6 +205,108 @@ def SetFields(cls, fields: Dict[TxnField, Union[Expr, List[Expr]]]) -> Expr:
...
+                    # app args _after_ appending since 0 is implicitly set
+                    case abi.AccountTypeSpec():
+                        if isinstance(arg, Expr):
+                            accts.append(arg)
+                        elif isinstance(arg, abi.Account):
+                            accts.append(arg.address())
+                        else:
+                            raise TealTypeError(arg, abi.Account | Expr)
+
+                        app_args.append(uint_encode(8, Int(len(accts))))
```
jasonpaulos, line 260: `uint_encode` encodes the argument at runtime, which is less efficient. It would be better to encode the value in Python and wrap it in a `Bytes` expression.

To do the encoding I'd recommend doing something like `algosdk.abi.ABIType.from_string("uint8").encode(len(accts))`

pyteal/ast/itxn.py
```diff
@@ -200,6 +205,108 @@ def SetFields(cls, fields: Dict[TxnField, Union[Expr, List[Expr]]]) -> Expr:
...
+                            assets.append(arg)
+                        elif isinstance(arg, abi.Asset):
+                            assets.append(arg.asset_id())
+                        else:
+                            raise TealTypeError(arg, abi.Asset | Expr)
+            else:
+                if isinstance(arg, Expr):
+                    app_args.append(arg)
+                elif isinstance(arg, abi.BaseType):
+                    app_args.append(arg.encode())
```
jasonpaulos, line 286: I'd like to see more type checking here, specifically making sure `arg`'s type spec matches `arg_ts`

pyteal/ast/itxn.py
```diff
@@ -200,6 +205,108 @@ def SetFields(cls, fields: Dict[TxnField, Union[Expr, List[Expr]]]) -> Expr:
...
+        accts: list[Expr] = []
+        apps: list[Expr] = []
+        assets: list[Expr] = []
+
+        for idx, arg_ts in enumerate(arg_type_specs):
+            arg = args[idx]
+
+            if arg_ts in abi.TransactionTypeSpecs:
+                if not isinstance(arg, dict):
+                    raise TealTypeError(arg, dict[TxnField, Expr | list[Expr]])
```
jasonpaulos, line 244: I'd also like to see more type checking here. We should error if you pass in the wrong type of transaction.

Checking if the dict has a `TxnField.type_enum` key and its value adheres to one of the following might work:
* `value` is an `Int` with the right value
* OR `value` is an `IntEnum` with the right string value

Since we're validating a runtime expression, this won't work in all cases. You could do something strange where the type of the txn isn't a constant, but I'd argue that's not an important scenario to support since it seems reasonable to expect inner txn types to be known at compile time.

pyteal/ast/itxn.py
```diff
@@ -200,6 +205,108 @@ def SetFields(cls, fields: Dict[TxnField, Union[Expr, List[Expr]]]) -> Expr:
...
 
+    @classmethod
+    def MethodCall(
+        cls,
+        *,
+        app_id: Expr,
+        method_signature: str,
+        args: list[abi.BaseType | Expr | dict[TxnField, Expr | list[Expr]]],
+        fields: dict[TxnField, Expr | list[Expr]],
+    ) -> Expr:
```
jasonpaulos, line 216: We'll definitely want a comprehensive docstring explaining how to use this function

jasonpaulos (commented):
Haven't looked at testing yet, but here are my thoughts so far

pyteal/ast/itxn.py
barnjamin: Started something here but I expect there should be more and possible improve formatting for docs

barnjamin:
I've got a weird case as I try to use this. Here the method signature asks for an asset to be passed as the 7th argument but it is optional during evaluation. That is if the payment transaction (6th arg) is an axfer it is consulted, otherwise we treat it as a payment. 

I'm unsure how to pass some undefined or none value here though. If we tried to encode a Uint16(0) in the arguments and skip appending to the assets array, there is a possible ambiguity since it refers to an element in the assets array. 

I guess this should be handled as 2 different transfer methods where the method signature only includes an asset reference type IFF it is an asset transfer as payment?

```py
 InnerTxnBuilder.MethodCall(
            app_id=app.application_id(),
            method_signature="transfer(asset,uint64,account,account,account,txn,asset,uint64)void",
            args=[
                asset,
                amt,
                owner,
                Txn.sender(),
                royalty_acct,
                {
                    TxnField.type_enum: TxnType.Payment,
                    TxnField.amount: pay_txn.get().amount(),
                    TxnField.receiver: app_acct.address(),
                },
                Int(0),
                pay_amt,
            ],
            fields={},
        ),
```

jasonpaulos:
@barnjamin I'd probably agree that two different methods would be the clearest way to solve that issue. Though you could consider keeping the current approach and using `asset[]` to pack 0 or 1 assets.

I hope that a future addition to the ABI will introduce optional types so that you won't need to use a dynamic array for this, and introduce unions of types as well, so you could use `pay | axfer` instead of the generic `txn` type.

barnjamin:
>@barnjamin I'd probably agree that two different methods would be the clearest way to solve that issue. Though you could consider keeping the current approach and using asset[] to pack 0 or 1 assets.

This is what I ended up doing, just split these into different methods to avoid this issue but I expect it'll come up in the future

pyteal/ast/itxn_test.py
```diff
@@ -134,4 +135,126 @@ def test_InnerTxnBuilder_SetFields():
...
+    expected, _ = expectedExpr.__teal__(teal6Options)
+    expected.addIncoming()
+    expected = pt.TealBlock.NormalizeBlocks(expected)
+
+    actual, _ = expr.__teal__(teal6Options)
+    actual.addIncoming()
+    actual = pt.TealBlock.NormalizeBlocks(actual)
+
+    with pt.TealComponent.Context.ignoreScratchSlotEquality(), pt.TealComponent.Context.ignoreExprEquality():
+        assert actual == expected
```
jasonpaulos, line 257: When using `TealComponent.Context.ignoreScratchSlotEquality()`, you'll also want to use `TealBlock.MatchScratchSlotReferences` to ensure the scratch slot references are still compatible.

An example's in this docstring: https://github.com/algorand/pyteal/blob/7304e8f07ffc2572a3fff563a625deaf0231f26e/pyteal/ir/tealcomponent.py#L70-L84

pyteal/ast/itxn_test.py
```diff
@@ -134,4 +135,126 @@ def test_InnerTxnBuilder_SetFields():
...
+        {TxnField.fee: pt.Int(0)},
+        pt.Seq(
+            pt.InnerTxnBuilder.SetFields(
+                {
+                    pt.TxnField.type_enum: TxnType.ApplicationCall,
+                    pt.TxnField.application_id: pt.Int(1),
+                    pt.TxnField.application_args: [
+                        pt.MethodSignature("add(uint64,uint64)void"),
+                        pt.abi.Uint64().encode(),
+                        pt.abi.Uint64().encode(),
```
jasonpaulos, line 172:
```suggestion
        [arg1 := pt.abi.Uint64(), arg2 := pt.abi.Uint64()],
        {TxnField.fee: pt.Int(0)},
        pt.Seq(
            pt.InnerTxnBuilder.SetFields(
                {
                    pt.TxnField.type_enum: TxnType.ApplicationCall,
                    pt.TxnField.application_id: pt.Int(1),
                    pt.TxnField.application_args: [
                        pt.MethodSignature("add(uint64,uint64)void"),
                        arg1.encode(),
                        arg2.encode(),
```

Referencing the same type instance in the expected response will strengthen the test. This is because it enables `TealBlock.MatchScratchSlotReferences` to check that the same underlying slots are used in the actual and expected responses.

Also applies to the next test with reference types.

(Writing this in a browser to convey the general idea, I haven't tested it)

pyteal/ast/itxn.py
```diff
@@ -225,14 +225,17 @@ def MethodCall(
...
+            args: A list of arguments to pass to the application. The values in this list depend on the kind of argument you wish to pass:
+
+                - For basic ABI arguments (not Reference or Transaction types): The ABI type is passed directly and encoded as part of transaction preparation. The type passed _MUST_ match the type specified in the `method_signature` passed.
+
+                - For Reference arguments: Either the Reference type or an Expr that returns the type corresponding to the reference type are allowed (Asset:TealType.uint64, Application:TealType.uint64, Account:TealType.bytes).
+
+                - For Transaction arguments: A dictionary containing TxnField to Expr that describe Transactions to be pre-pended to the transaction group being constructed.  The `TxnField.type_enum` key MUST be set and MUST match the expected transaction type specified in the `method_signature`.
+
+                - For any others: An expression that evaluates to bytes representing a valid ABI encoded type passed directly to the arguments array.  No type checking is performed besides checking that it evaluates to `TealType.bytes`.
```
jasonpaulos, line 236: I think this would be more appropriate under basic ABI arguments, since it doesn't apply to reference or transaction types?

pyteal/ast/itxn_test.py
```diff
@@ -134,4 +135,268 @@ def test_InnerTxnBuilder_SetFields():
...
+        pt.TealTypeError,
+    ),
+    (
+        pt.Int(1),
+        "add(uint64,uint64)void",
+        [pt.abi.String(), pt.abi.Uint64()],
+        None,
+        None,
+        pt.TealTypeError,
+    ),
```
jasonpaulos, line 354: I just thought of 2 more invalid cases to test: too few or too many arguments being passed. Based on the implementation I think too many args is silently accepted, while too few args will error but the message might not be the cleanest, since it'll be a generic array out of bounds error.

pyteal/ast/itxn.py
```diff
@@ -200,6 +206,178 @@ def SetFields(cls, fields: Dict[TxnField, Union[Expr, List[Expr]]]) -> Expr:
...
+
+        Requires Teal version 6 or higher. This operation is only permitted in application mode.
+
+        Args:
+            app_id: An expression that evaluates to a `TealType.uint64` corresponding to the application being called.
+            method_signature: A string representing the method signature of the method we're calling. This is used to do
+                type checking on the arguments passed and to create the method selector passed as the first argument.
+            args: A list of arguments to pass to the application. The values in this list depend on the kind of argument you wish to pass:
+
+                - For basic ABI arguments (not Reference or Transaction types): The ABI type is passed directly and encoded as part of transaction preparation. The type passed _MUST_ match the type specified in the `method_signature` passed.
```
jasonpaulos, line 230: nit: the `_MUST_` in this line doesn't get formatted at all in the docs. Bold & italics are described here: https://thomas-cokelaer.info/tutorials/sphinx/rest_syntax.html#inline-markup-and-special-characters-e-g-bold-italic-verbatim

jasonpaulos (commented):
I think these are the last of my comments -- thanks for quickly addressing everything I've pointed out!

jasonpaulos (approved):
LGTM

PR #389, opened 2022-06-10
Rename decode parameters to abide by PEP8 conventions

michaeldiamant:
Renames parameters in `pyteal.ast.abi.Type.decode`to abide by PEP8 naming conventions.

* The intent is to avoid backwards incompatibilities by renaming _before_ an initial release.  The PR deliberately makes _no_ other changes to keep the changeset minimal. 
* Changes generated via PyCharm renaming facilities.


jasonpaulos (approved):
Looks great!

PR #400, opened 2022-06-16
Document new ABI features

### LINKED ISSUE #367: ABI:  Prep user guide for initial ABI release
## Problem

We're aiming for our 1st ABI release, which brings numerous ABI capabilities to production.  See associated epic:  https://github.com/algorand/pyteal/issues/185.

Since we haven't developed a user guide in parallel with each feature, the story asks for:
* A user guide priming PyTeal developers to leverage ABI functionality:
  * Router method definition + registration 
  * Contract description (JSON) generation
  * ABI type read + write usage 
* Explicit mention for limited level of support mirroring disclaimers around `Router` and `ABIReturnSubroutine`.

jasonpaulos:
Closes #367 

For viewing the docs changes, there are a few options:
* Checkout this branch locally, activate the virtual environment, cd to the `docs` folder, and run `make html`
* OR, you can download the `pyteal.docset` artifact from github actions and use either [Kapeli's Dash](https://kapeli.com/dash) or [zeal](https://zealdocs.org/) to view them locally.


docs/abi.rst
```diff
@@ -0,0 +1,560 @@
...
+       * :any:`Router` usage for defining how to route program invocations.
+    * For general purpose :any:`Subroutine` definition usage, use at your own risk.  Based on feedback, the API and usage patterns will change more freely and with less effort to provide migration paths.
+
+TODO: warning about how ABI types are intended to be used at the moment (for standard serialization at the incoming and outgoing layers of contract) - internal use may be extremely inefficient with ABI types
+TODO: warning about how internal representation of ABI types may change over time
+
+Types
+------
+
+The ABI supports a variety of types whose encodings are standardized.
```
algoanne, line 24: "data types" perhaps?

docs/abi.rst
```diff
@@ -0,0 +1,560 @@
...
+
+Fundamentals
+~~~~~~~~~~~~
+
+Before we introduce these ABI types, it's important to understand the fundamentals of how PyTeal's ABI type system works.
```
algoanne, line 29: the way this is worded ("it's important" without an actual reason) makes me think I can skip this section, but in reality I shouldn't. Maybe something like "We first explain the fundamentals of PyTeal's ABI type system, which includes behavior common to all ABI types."

docs/abi.rst
```diff
@@ -0,0 +1,560 @@
...
+^^^^^^^^^^^^^^^^^^^^
+
+:any:`abi.TypeSpec` is an abstract base class used to describe ABI types. Every child class of :code:`abi.BaseType` also has a companion :code:`abi.TypeSpec` child class. The :code:`abi.TypeSpec` class has a few methods that return information about the type it represents, but one of the class's most important features is the method :any:`abi.TypeSpec.new_instance()`, which creates and returns a new :any:`abi.BaseType` instance of the ABI type it represents.
+
+Static vs Dynamic Types
+^^^^^^^^^^^^^^^^^^^^^^^
+
+An important property of an ABI type is whether it is static or dynamic.
+
+Static types are defined as types whose encoded length does not depend on the value of that type. This property allows encoding and decoding of static types to be more efficient.
```
algoanne, line 50: perhaps you could add "(for example: a boolean)" and "(for example: a string)" to exemplify what is meant in the section.

docs/abi.rst
```diff
@@ -0,0 +1,560 @@
...
+:any:`abi.Byte`                                :code:`byte`           Static                            An 8-bit unsigned integer. This is an alias for :code:`abi.Uint8` that should be used to indicate non-numeric data, such as binary arrays.
+:any:`abi.StaticArray[T,N] <abi.StaticArray>`  :code:`T[N]`           Static if :code:`T` is static     A fixed-length array of :code:`T` with :code:`N` elements
+:any:`abi.Address`                             :code:`address`        Static                            A 32-byte Algorand address. This is an alias for :code:`abi.StaticArray[abi.Byte, Literal[32]]`.
+:any:`abi.DynamicArray[T] <abi.DynamicArray>`  :code:`T[]`            Dynamic                           A variable-length array of :code:`T`
+:any:`abi.String`                              :code:`string`         Dynamic                           A variable-length byte array assumed to contain UTF-8 encoded content. This is an alias for :code:`abi.DynamicArray[abi.Byte]`.
+:any:`abi.Tuple`\*                             :code:`(...)`          Static if all elements are static A tuple of multiple types
```
algoanne, line 139: perhaps "if and only if"

docs/abi.rst
```diff
@@ -0,0 +1,560 @@
...
+Static types are defined as types whose encoded length does not depend on the value of that type. This property allows encoding and decoding of static types to be more efficient.
+
+Likewise, dynamic types are defined as types whose encoded length does in fact depend on the value that type has. Due to this dependency, the code that PyTeal generates to encode, decode, and manipulate dynamic types is more complex and generally less efficient than the code needed for static types.
+
+Because of the difference in complexity and efficiency when working with static and dynamic types, **we strongly recommend using static types over dynamic types whenever possible**.
+
+Instantiating Types
+^^^^^^^^^^^^^^^^^^^
+
+There are a few ways to create an instance of an ABI type. Each method produces the same result, but some may be more convenient than others.
```
algoanne, line 59: I suggest removing ", but some may be more convenient than others." which doesn't really add anything imo.

docs/abi.rst
```diff
@@ -0,0 +1,560 @@
...
+    myUint8 = abi.Uint8()
+    myUint64 = abi.Uint64()
+    myArrayOf12Uint8s = abi.StaticArray(abi.StaticArrayTypeSpec(abi.Uint8TypeSpec(), 12))
+
+For simple types, using the constructor is straightforward and works as you would expect. However, more complex types like :any:`abi.StaticArray` have type-level arguments, so their constructor must take an :any:`abi.TypeSpec` which fully defines all necessary arguments. These types can be created with a constructor, but it's often not the most convenient way to do so.
+
+With an :code:`abi.TypeSpec` Instance
+""""""""""""""""""""""""""""""""""""""
+
+You may remember that :code:`abi.TypeSpec` has a :any:`new_instance() <abi.TypeSpec.new_instance>` method that can be used to instantiate ABI types. This is another way of instantiating ABI types, if you happen to have an :code:`abi.TypeSpec` instance available. For example:
```
algoanne, line 77: could remove "happen to"

docs/abi.rst
```diff
@@ -0,0 +1,560 @@
...
+
+.. code-block:: python
+
+    myUintType = abi.Uint8TypeSpec()
+    myUint8 = myUintType.new_instance()
+
+    myArrayType = abi.StaticArrayTypeSpec(myUintType, 12)
+    myArrayOf12Uint8s = myArrayType.new_instance()
+
+With :code:`abi.make`
```
algoanne, line 87: I agree this syntax is more readable. I don't see an obvious way to solve this, but in reality I was annoyed that you made me read the previous section-and-a-half before telling me the nicest way to instantiate a complex type. 

I wonder if a short "tl;dr" section at the top of this document that literally just shows the right examples would be useful for the lazies of the world. Not saying it's a good idea, just thinking.

docs/abi.rst
```diff
@@ -0,0 +1,560 @@
...
+1. Basic types
+2. Reference types
+3. Transaction types
+
+Each of which is described in detail in the following subsections.
+
+Basic Types
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+Basic types are the most straightforward category of ABI types. These types are used to hold values and they have no other side effects, in contrast to the other categories of types.
```
algoanne, line 119: suggestion: "side effects" -> "properties" (or something other than "side effects" which has a negative connotation and feels imprecise)

docs/abi.rst
```diff
@@ -0,0 +1,560 @@
...
+''''''''''''''''
+
+All basic types have a :code:`set()` method which can be used to assign a value. The arguments for this method differ depending on the ABI type. For convenience, here are links to the docs for each class's method:
+
+* :any:`abi.Uint.set()`, which is used by all :code:`abi.Uint` classes and :code:`abi.Byte`
+* :any:`abi.Bool.set()`
+* :any:`abi.StaticArray.set()`
+* :any:`abi.Address.set()`
+* :any:`abi.DynamicArray.set()`
+* :any:`abi.String.set()`
```
algoanne, line 171: I take it this is _not_ the same as the DynamicArray set method? (unlike the correspondence for Byte using the Uint one)

docs/abi.rst
```diff
@@ -0,0 +1,560 @@
...
+
+    myAddress = abi.make(abi.Address)
+    myBool = abi.make(abi.Bool)
+    myUint64 = abi.make(abi.Uint64)
+    myTuple = abi.make(abi.Tuple3[abi.Address, abi.Bool, abi.Uint64])
+
+    program = Seq(
+        myAddress.set(Txn.sender()),
+        myBool.set(Txn.fee() == Int(0)),
+        myUint64.set(5000),
```
algoanne, line 186: I'm surely wrong because I know nothing and am just pattern matching, but in case I'm not - should this be "Int(5000)"?

docs/abi.rst
```diff
@@ -0,0 +1,560 @@
...
+        myTuple.set(myAddress, myBool, myUint64)
+    )
+
+Getting Values
```
algoanne, line 190: maybe could title it "Getting Single Values"? I was confused at first why the list was missing some items then I had to read more closely

docs/abi.rst
```diff
@@ -0,0 +1,560 @@
...
+* :any:`abi.Tuple.__getitem__`
+
+Be aware that these methods return a :code:`ComputedValue`, TODO link to Computed Value section
+
+A brief example is below. Please consult the documentation linked above for each method to learn more about specific usage and behavior.
+
+.. code-block:: python
+
+    @Subroutine(TealType.none)
+    def ensure_all_values_greater_than_5(array: abi.StaticArray[abi.Uint64, L[10]]) -> Expr:
```
algoanne, line 231: L[10] is the same as Literal[10]?

docs/abi.rst
```diff
@@ -0,0 +1,560 @@
...
+
+The types :code:`abi.StaticArray`, :code:`abi.Address`, :code:`abi.DynamicArray`, :code:`abi.String`, and :code:`abi.Tuple` are compound types, meaning they contain other types whose values can be extracted. The :code:`__getitem__` method, accessible by using square brackets to "index into" an object, can be used to extract these values.
+
+The supported methods are:
+
+* :any:`abi.StaticArray.__getitem__`, used for :code:`abi.StaticArray` and :code:`abi.Address`
+* :any:`abi.Array.__getitem__`, used for :code:`abi.DynamicArray` and :code:`abi.String`
+* :any:`abi.Tuple.__getitem__`
+
+Be aware that these methods return a :code:`ComputedValue`, TODO link to Computed Value section
```
algoanne, line 224: I'm sure there's a good reason for this, but as a user it's unintuitive that it wouldn't just return a value of the type in the array/tuple. So good callout, and I suggest we give a brief explanation for why this is the case

docs/abi.rst
```diff
@@ -0,0 +1,560 @@
...
+
+Limitations
+"""""""""""""""""""""
+
+TODO: explain type size limitations
+
+Reference Types
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+Some AVM operations require specific values to be placed in the "foreign arrays" of the app call transaction. Reference types allow methods to describe these requirements.
```
algoanne, line 248: I suggest giving the reader just a little more here, something like (not in love with how I wrote it, but as an example): 

Many applications require the caller to specify "foreign arrays" when calling the app. These are the blockchain entities (such as accounts, assets, or other applications) that the application will interact with when executing this call. In the ABI, these are called "reference types". Reference types allow methods to describe these requirements.

docs/abi.rst
```diff
@@ -0,0 +1,560 @@
...
+
+Reference types are only valid in the arguments of a method. They may not appear in a method's return value in any form.
```
algoanne, line 250: perhaps remove "in any form"

algoanne: (I'm not sure if you had something in mind that you think people might try to do)

docs/abi.rst
```diff
@@ -0,0 +1,560 @@
...
+
+Definitions
+""""""""""""""""""""""""""""""""""""""""""
+
+PyTeal supports the following reference types:
+
+====================== ====================== ================ =======================================================================================================================================================
+PyTeal Type            ARC-4 Type             Dynamic / Static Description
+====================== ====================== ================ =======================================================================================================================================================
+:any:`abi.Account`     :code:`account`        Static           Represents an additional account that the current transaction can access, stored in the :any:`Txn.accounts <TxnObject.accounts>` array
```
algoanne, line 260: I think the word "additional" is potentially confusing here - additional to what? If you mean in addition to the app's account, then maybe it makes sense, but then I would at least remove it from the assets line.

docs/abi.rst
```diff
@@ -0,0 +1,560 @@
...
+
+.. code-block:: python
+
+    @Subroutine(TealType.none)
+    def referenced_index_example(
+        account: abi.Account, asset: abi.Asset, app: abi.Application
+    ) -> Expr:
+        return Seq(
+            # The accounts array has Txn.accounts.length() + 1 elements in it (the +1 is the txn sender)
+            Assert(account.referenced_index() <= Txn.accounts.length()),
```
algoanne, line 287: it took me a few minutes to understand how `referenced_index()` works, still a bit confused maybe about what the point of it is

algoanne: ok, so this example method requires an account reference, an asset reference, and an app reference as arguments. 

These aren't all at index 0, though, because for accounts and apps there's a default at index 0. 

The comment "The accounts array has Txn.accounts.length() + 1 elements in it" - you're saying the length() method is implemented to actually not give what you'd expect, which is the length of that array, but rather the length of the args put in that array? 

In short: I think you should add a little more explanation and comments here because it feels pretty counterintuitive.  And maybe the example could have 2 args of one of the arrays to help understand.

docs/abi.rst
```diff
@@ -0,0 +1,560 @@
...
+
+Limitations
+""""""""""""""""""""""""""""""""""""""""""
+
+TODO: explain limitations, such as can't be created directly, or used as method return value
+
+Transaction Types
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+Some application calls require that they are invoked as part of a larger transaction group containing specific additional transactions. In order to express these types of calls, the ABI has transaction types.
```
algoanne, line 401: I was thinking about this in the Reference Types section as well, but would you consider capitalizing "Reference Types" and "Transaction Types"? I'm not sure myself, but it feels like lower case is a little harder to parse - it is that the ABI has transaction _types_, or that the ABI has _transaction_ types, or that the ABI has _transaction types_? Since these are both such common words in our day to day, it might be worth capitalizing.

docs/abi.rst
```diff
@@ -0,0 +1,560 @@
...
+
+Every transaction type argument represents a specific, unique, transaction that must appear immediately before the application call.
```
algoanne, line 403: suggest: "specific and unique transaction". 

Also, with respect to "that must appear immediately before the application call": "immediately before" in what context (in the transaction group, I assume)? 
The way you phrased this suggests that there can only be one Transaction Type per app call, but I don't think that's true, so maybe worth rephrasing somehow.

algoanne:
Firstly, this all looks great so far! I think pyteal users are going to be thrilled to have this resource and it will make it very easy for them to use the ABI functionality. 

One more global thought I have: Do we have a separate document explaining the AVM concepts of app calls, method arguments, foreign arrays, and other transactions in the transaction group? 
Reading this user guide, I found myself hoping people have a good understanding of those concepts before they dive into this. I think it's useful to minimally remind people of these concepts in this ABI user guide, but it would also make sense to make sure we have proper documentation for that as well. I thought about maybe a section at the top of this user guide called "AVM Concept Reminders on App Calls" or something.

docs/abi.rst
```diff
@@ -0,0 +1,652 @@
...
+    myUint64 = abi.make(abi.Uint64)
+    myArrayOf12Uint8s = abi.make(abi.StaticArray[abi.Uint8, Literal[12]])
+
+.. note::
+    Since Python does not allow integers to be directly embedded in type annotations, you must wrap any integer arguments in the :code:`Literal` annotation from the :code:`typing` module.
+
+Computed Values
+^^^^^^^^^^^^^^^^^^^^^^^
+
+With the introduction of ABI types, it's only natural for there to be functions and operations which return ABI values. In a conventional language, it would be enough to return an instance of the ABI type directly from the operation, as usual. However, in PyTeal, there operations must actually return two things:
```
michaeldiamant, line 108: @jasonpaulos Few _suggestions_ here:
* Remove _there_ - I think it's a typo.  Possibly, another phrasing was intended.
* Optionally, I'd opt for a more specific term (values) rather than _things_.

```suggestion
With the introduction of ABI types, it's only natural for there to be functions and operations which return ABI values. In a conventional language, it would be enough to return an instance of the ABI type directly from the operation, as usual. However, in PyTeal, operations must actually return two values:
```

docs/abi.rst
```diff
@@ -0,0 +1,652 @@
...
+
+The :any:`abi.ComputedValue[T] <abi.ComputedValue>` abstract base class provides the following methods:
+
+* :any:`abi.ComputedValue[T].produced_type_spec() <abi.ComputedValue.produced_type_spec>`: returns the :any:`abi.TypeSpec` representing the ABI type produced by this object.
+* :any:`abi.ComputedValue[T].store_into(output: T) <abi.ComputedValue.store_into>`: computes the value and store it into the ABI type instance :code:`output`.
+* :any:`abi.ComputedValue[T].use(action: Callable[[T], Expr]) -> Expr <abi.ComputedValue.use>`: computes the value and passes it to the callable expression :code:`action`. This is offered as a convenience over the :code:`store_into(...)` method if you don't want to create a new variable to store the value before using it.
+
+.. note::
+    If you call the methods :code:`store_into(...)` or :code:`use(...)` multiple times, the computation to determine the value will be repeated each time. For this reason, it is recommended to only issue a single call to either of these two method.
```
michaeldiamant, line 124: @jasonpaulos Leaving a _suggestion_ for a perceived minor typo.

```suggestion
    If you call the methods :code:`store_into(...)` or :code:`use(...)` multiple times, the computation to determine the value will be repeated each time. For this reason, it is recommended to only issue a single call to either of these two methods.
```

docs/abi.rst
```diff
@@ -0,0 +1,652 @@
...
+PyTeal Type                         ARC-4 Type             Dynamic / Static Description
+=================================== ====================== ================ =======================================================================================================================================================
+:any:`abi.Transaction`              :code:`txn`            Static           A catch-all for any transaction type
+:any:`abi.PaymentTransaction`       :code:`pay`            Static           A payment transaction
+:any:`abi.KeyRegisterTransaction`   :code:`keyreg`         Static           A key registration transaction
+:any:`abi.AssetConfigTransaction`   :code:`acfg`           Static           An asset configuration transaction
+:any:`abi.AssetTransferTransaction` :code:`axfer`          Static           An asset transfer transaction
+:any:`abi.AssetFreezeTransaction`   :code:`afrz`           Static           An asset freeze transaction
+:any:`abi.AssetTransferTransaction` :code:`appl`           Static           An application call transaction
```
michaeldiamant, line 472: @jasonpaulos Appears to be a copy-paste typo.

```suggestion
:any:`abi.ApplicationCallTransaction` :code:`appl`           Static           An application call transaction
```

docs/abi.rst
```diff
@@ -0,0 +1,652 @@
...
+Subroutines can be created that accept ABI types are arguments and produce ABI types as return values. PyTeal will type check all subroutine calls and ensure that the correct types are being passed to such subroutines and that their return values are used correctly.
+
+There are two different ways to use ABI types in subroutines, depending on whether the return value is an ABI type or a PyTeal :code:`Expr`.
+
+Subroutines that Return Expressions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you'd like to create a subroutine that accepts some or all arguments as ABI types, but whose return value is a PyTeal :code:`Expr`, the normal :any:`@Subroutine <Subroutine>` decorator can be used.
+
+To indicate the type of each argument, Python type annotations are used. Unlike normal usage of Python type annotations which are ignored at runtime, type annotations for subroutines inform the PyTeal compiler about the inputs and outputs of a subroutine. Changing these values has a direct affect on the code PyTeal generates.
```
michaeldiamant, line 541:
```suggestion
To indicate the type of each argument, Python type annotations are used. Unlike normal usage of Python type annotations which are ignored at runtime, type annotations for subroutines inform the PyTeal compiler about the inputs and outputs of a subroutine. Changing these values has a direct effect on the code PyTeal generates.
```

docs/abi.rst
```diff
@@ -0,0 +1,652 @@
...
+    status = abi.make(abi.Tuple2[abi.Uint64, abi.Bool])
+    program = Seq(
+        account.set(Txn.sender()),
+        # NOTE! The return value of get_account_status(account) is actually a ComputedValue[abi.Tuple2[abi.Uint64, abi.Bool]]
+        get_account_status(account).store_into(status),
+    )
+
+Notice that even though the original :code:`get_account_status` function returns an :code:`Expr` object, the :code:`@ABIReturnSubroutine` decorator automatically transforms the function's return value and the :code:`output` variable into a :code:`ComputedValue`. As a result, callers of this subroutine must work with a :code:`ComputedValue`.
+
+The only exception to this transformation is if the subroutine has no return value, in which case a :code:`ComputedValue` is unnecessary, so the subroutine will still return an :code:`Expr` to the caller. In this case the :code:`@ABIReturnSubroutine` decorator acts identically the original :code:`@Subroutine` decorator.
```
michaeldiamant, line 611: @jasonpaulos Minor and optional _suggestions_ based on somewhat subjective preferences for shorter sentences and present tense usage.

```suggestion
The only exception to this transformation is if the subroutine has no return value. Without a return value, :code:`ComputedValue` is unnecessary and the subroutine returns an :code:`Expr` to the caller. In this case, the :code:`@ABIReturnSubroutine` decorator acts identically the original :code:`@Subroutine` decorator.
```

docs/abi.rst
```diff
@@ -0,0 +1,652 @@
...
+--------------------------
+
+Subroutines can be created that accept ABI types are arguments and produce ABI types as return values. PyTeal will type check all subroutine calls and ensure that the correct types are being passed to such subroutines and that their return values are used correctly.
+
+There are two different ways to use ABI types in subroutines, depending on whether the return value is an ABI type or a PyTeal :code:`Expr`.
+
+Subroutines that Return Expressions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you'd like to create a subroutine that accepts some or all arguments as ABI types, but whose return value is a PyTeal :code:`Expr`, the normal :any:`@Subroutine <Subroutine>` decorator can be used.
```
michaeldiamant, line 539: @jasonpaulos Minor and optional:  I think _normal_ can be removed because the two discussed subroutine decorators have unique names.  

```suggestion
If you'd like to create a subroutine that accepts some or all arguments as ABI types, but whose return value is a PyTeal :code:`Expr`, the :any:`@Subroutine <Subroutine>` decorator can be used.
```

docs/abi.rst
ahangsu: yes, I think if we declare `from typing import Literal as L`, then this is the code snippet you are suggesting.

docs/abi.rst
```diff
@@ -588,13 +588,13 @@ The :code:`@ABIReturnSubroutine` decorator should be used with subroutines that
...
     def get_account_status(
-        account: abi.Account, *, output: abi.Tuple2[abi.Uint64, abi.Bool]
+        account: abi.Address, *, output: abi.Tuple2[abi.Uint64, abi.Bool]
     ) -> Expr:
         balance = abi.Uint64()
         is_admin = abi.Bool()
         return Seq(
-            balance.set(App.localGet(account.address(), Bytes("balance"))),
-            is_admin.set(App.localGet(account.address(), Bytes("is_admin"))),
+            balance.set(App.localGet(account.get(), Bytes("balance"))),
```
barnjamin, line 596: I think this needs to stay an account so that the reference works

pyteal/ast/bytes.py
```diff
@@ -15,14 +15,18 @@ class Bytes(LeafExpr):
 
     @overload
     def __init__(self, arg1: Union[str, bytes, bytearray]) -> None:
-        ...
+        pass
```
tzaffi, line 18: 👍

docs/abi.rst
```diff
@@ -0,0 +1,820 @@
...
+            # store the result in the sender's local state too
+            App.localPut(Txn.sender(), Bytes("result", output.get())),
+        )
+
+Compiling a Router Program
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Now that we know how to add bare app call and method call handlers to a :code:`Router`, the next step is to compile the :code:`Router` into TEAL code.
+
+The :any:`Router.compile_program` method exists for this purpose. It combines all registered methods and bare app calls into two ASTs, one for the approval program and one for clear state program, then internally calls :any:`compileTeal` to compile these expressions and create TEAL code.
```
tzaffi, line 793: Consider emphasizing `approval` and `clear`

docs/abi.rst
```diff
@@ -0,0 +1,820 @@
...
+
+In addition to receiving the approval and clear state programs, the :code:`Router.compile_program` method also returns a :code:`Contract` object `from the Python SDK <https://py-algorand-sdk.readthedocs.io/en/latest/algosdk/abi/contract.html#algosdk.abi.contract.Contract>`_. This object represents an `ARC-4 Contract Description <https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0004.md#contract-description>`_, which can be distributed to clients to enable them to call the methods on the contract.
```
tzaffi, line 795:
```suggestion
In addition to receiving the approval and clear state programs, the :code:`Router.compile_program` method also returns a `Python SDK <https://py-algorand-sdk.readthedocs.io/en/latest/algosdk/abi/contract.html#algosdk.abi.contract.Contract>`  :code:`Contract` object. This object represents an `ARC-4 Contract Description <https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0004.md#contract-description>`_, which can be distributed to clients to enable them to call the methods on the contract.
```

tzaffi, line 795:
```suggestion
In addition to receiving the approval and clear state programs, the :code:`Router.compile_program` method also returns a :code:`Contract` object `from the Python SDK <https://py-algorand-sdk.readthedocs.io/en/latest/algosdk/abi/contract.html#algosdk.abi.contract.Contract>`_. This object represents an `ARC-4 Contract Description <https://arc.algorand.foundation/ARCs/arc-0004#contract-description>`_, which can be distributed to clients to enable them to call the methods on the contract.
```

suggest repointing to the foundation instead of github

examples/application/abi/algobank.py
```diff
@@ -0,0 +1,113 @@
+# This example is provided for informational purposes only and has not been audited for security.
+from pyteal import *
```
tzaffi, line 2: nit, consider making explicit all the imports

examples/application/abi/algobank.py
```diff
@@ -0,0 +1,113 @@
...
+import json
+
+
+@Subroutine(TealType.none)
```
tzaffi, line 6: really nice example!!!

examples/application/abi/algobank.py
```diff
@@ -0,0 +1,113 @@
...
+        close_out=OnCompleteAction(
+            action=transfer_balance_to_lost, call_config=CallConfig.CALL
+        ),
+        clear_state=OnCompleteAction(
+            action=transfer_balance_to_lost, call_config=CallConfig.CALL
+        ),
+        # only the creator can update or delete the app
+        update_application=OnCompleteAction(
+            action=assert_sender_is_creator, call_config=CallConfig.CALL
+        ),
```
tzaffi, line 34: Oops, I actually meant to leave this comment on `abi.rst` not on the actual `.py` example

```suggestion
        # approve a creation no-op call
        no_op=OnCompleteAction(action=Approve(), call_config=CallConfig.CREATE),
        
        # approve opt-in calls during normal usage, and during creation as a convenience for the creator
        opt_in=OnCompleteAction(action=Approve(), call_config=CallConfig.ALL),
        
        # move any balance that the user has into the "lost" amount when they close out or clear state
        close_out=OnCompleteAction(
            action=transfer_balance_to_lost, call_config=CallConfig.CALL
        ),
        clear_state=OnCompleteAction(
            action=transfer_balance_to_lost, call_config=CallConfig.CALL
        ),
        
        # only the creator can update or delete the app
        update_application=OnCompleteAction(
            action=assert_sender_is_creator, call_config=CallConfig.CALL
        ),
```

`black` probably disagrees with me, but I find spacing make it a little easier to read

tzaffi (commented):
I focused my attention thus far to `docs/abi.rst`. I hope to look at the other changes as well. Overall, a very solid addition that will be quite helpful to PyTeal dev's who want to start using the ABI and Router functionality.

I left numerous nit's and minor suggestions.

jdtzmn:
Great work on this user guide! After reading, I felt I had a solid understanding of all the relevant components necessary for ABI PyTeal development as well as their limitations. The algobank example was a good idea.

docs/abi.rst
```diff
@@ -0,0 +1,560 @@
...
+:any:`abi.Bool`                                :code:`bool`           Static                            A boolean value that can be either 0 or 1
+:any:`abi.Byte`                                :code:`byte`           Static                            An 8-bit unsigned integer. This is an alias for :code:`abi.Uint8` that should be used to indicate non-numeric data, such as binary arrays.
+:any:`abi.StaticArray[T,N] <abi.StaticArray>`  :code:`T[N]`           Static if :code:`T` is static     A fixed-length array of :code:`T` with :code:`N` elements
+:any:`abi.Address`                             :code:`address`        Static                            A 32-byte Algorand address. This is an alias for :code:`abi.StaticArray[abi.Byte, Literal[32]]`.
+:any:`abi.DynamicArray[T] <abi.DynamicArray>`  :code:`T[]`            Dynamic                           A variable-length array of :code:`T`
+:any:`abi.String`                              :code:`string`         Dynamic                           A variable-length byte array assumed to contain UTF-8 encoded content. This is an alias for :code:`abi.DynamicArray[abi.Byte]`.
```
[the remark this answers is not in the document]
algoanne: but is there any behavior behind the "assumed"? i.e. will it throw up an error if it's not UTF-8

ahangsu (commented):
took a quick pass on code comments, mostly looking good. Still looking abi.rst and algobank.py

tests/unit/compile_test.py
```diff
@@ -1,9 +1,40 @@
 from pathlib import Path
 import pytest
+import json
 
 import pyteal as pt
 
 
+def test_abi_algobank():
```
tzaffi, line 8: nice!

tzaffi (approved):
💯 💯 💯 💯  (for #400)

docs/abi.rst
```diff
@@ -0,0 +1,919 @@
...
+
+* Non-power-of-2 unsigned integers under 64 bits, i.e. :code:`uint24`, :code:`uint48`, :code:`uint56`
+* Unsigned integers larger than 64 bits
+* Fixed point unsigned integers, i.e. :code:`ufixed<N>x<M>`
+
+Limitations
+"""""""""""""""""""""
+
+Due to the nature of their encoding, dynamic container types, i.e. :any:`abi.DynamicArray[T] <abi.DynamicArray>` and :any:`abi.String`, have an implicit limit on the number of elements they may contain. This limit is :code:`2^16 - 1`, or 65535. Do not attempt to store more elements than this in these types.
+
```
barnjamin, line 236: There is a limit on number of bytes the stack can hold implicitly at 4k which I think would get breached before we hit 2^16 elements, worth mentioning?

docs/abi.rst
```diff
@@ -0,0 +1,919 @@
...
+This page will introduce and explain the relevant concepts necessary to build a PyTeal application that adheres to the ARC-4 ABI standards.
+
+Types
+------
+
+The ABI supports a variety of data types whose encodings are standardized.
+
+.. note::
+    Be aware that the ABI type system in PyTeal has been designed specifically for the limited use case of describing a program's inputs and outputs. At the time of writing, we **do not recommend** using ABI types in a program's internal storage or computation logic, as the more basic :any:`TealType.uint64` and :any:`TealType.bytes` :any:`Expr` types are far more efficient for these purposes.
+
```
barnjamin, line 29: This q should _not_ block release, just putting it here for consideration

What is the appropriate way to store things that come in as ABI types? for dynamic types there may be a lot of overhead in the encoded version but static types are more efficient. Should we recommend that _if_ you're storing data, it should be a static type and use `encode`?

barnjamin (approved):
Great work!

docs/abi.rst
jasonpaulos: Great point, I added a sentence to address this

docs/abi.rst
jasonpaulos: Yes, I don't see much of a downside for storing encoded static types.

Unless you want to store single integers in local/global state, in which case you need not convert them to bytes anyway.

For the time being I'd like to retain this warning, but perhaps over time we can clarify this point.

docs/abi.rst
```diff
@@ -0,0 +1,919 @@
...
+These types all inherit from the abstract class :any:`abi.ReferenceType`.
+
+Limitations
+""""""""""""""""""""""""""""""""""""""""""
+
+Because References Types have a special meaning, they should not be directly created, and they cannot be assigned a value by a program.
+
+Additionally, Reference Types are only valid in the arguments of a method. They may not appear in a method's return value.
+
+Note that the AVM has limitations on the maximum number of foreign references an application call transaction may contain. At the time of writing, these limits are:
```
michaeldiamant, line 360: @jasonpaulos Optional:  While it's convenient to list reference limits, I'm concerned it'll grow stale.  How open are you to one of the following alternatives?

* Link to a (or several) Github search(es) helping the user find the current value in `config/consensus.go`.
* Generally pointing the user to `config/consensus.go` to deduce values.

If you'd rather keep as is, then I don't feel strongly enough to pursue further and the thread can be resolved.

docs/abi.rst
```diff
@@ -0,0 +1,919 @@
...
+    )
+
+.. note::
+    When deciding which :any:`CallConfig` value is appropriate for a bare app call or method, consider the question, should it be valid for someone to create my app with this operation? Most of the time the answer will be no, in which case :any:`CallConfig.CALL` should be used.
+
+Registering Methods
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. warning::
+    A challenge for any system exposed to the public, including smart contracts, is input validation. At the moment, PyTeal's :any:`Router` class only performs a limited set of input validation. The largest hole in this validation is compound types, i.e. :any:`abi.StaticArray`, :any:`abi.Address`, :any:`abi.DynamicArray`, and :any:`abi.String`.
```
michaeldiamant, line 756: @jasonpaulos Optional:  I think the _warning_ provides good content, but can be more concise + have fewer long sentences.  If you prefer as is, feel welcomed to resolve.

Here's a proposal for slimming down the warning while hopefully retaining its intent.  If you don't like the proposal, feel welcomed to disregard it.

:any:`Router` does _not_ validate compound types (:any:`abi.StaticArray`, :any:`abi.Address`, :any:`abi.DynamicArray`, and :any:`abi.String`).  **We strongly recommend** methods immediately access and validate compound type parameters _before_ persisting arguments for later transactions.

Notes:
* The recommendation applies to recursively defined compound types.
* :any:`abi.Address` is _not_ guaranteed to have exactly 32 bytes.  To defend against unintended behavior, manually verify the length is 32 bytes.

-----

Aside:  Additionally and also optionally, I'd consider adding _compound types_ to an existing section header.  
* That way the warning can refer to the section defining compound types rather than re-enumerating the types. 
* Candidate section:  _Getting Values at Indexes_ can be named something like _Getting Values at Indexes - Compound types_

docs/abi.rst
```diff
@@ -0,0 +1,919 @@
...
+    ABI support is still taking shape and is subject to backwards incompatible changes.
+        * Based on feedback, the API and usage patterns are likely to change.
+        * For the following use cases, feel encouraged to rely on abstractions. Expect a best-effort attempt to minimize backwards incompatible changes along with a migration path.
+
+            * :any:`ABIReturnSubroutine` usage for ABI Application entry point definition.
+
+            * :any:`Router` usage for defining how to route program invocations.
+        * For general purpose :any:`Subroutine` definition usage, use at your own risk. Based on feedback, the API and usage patterns will change more freely and with less effort to provide migration paths.
+
+    For these reasons, we strongly recommend using :any:`pragma` or the :any:`Pragma` expression to pin the version of PyTeal in your source code.
```
michaeldiamant, line 16: @jasonpaulos Sanity checking:  Is it by-design that the user guide does _not_ provide pragma explanation beyond the API docs?

* I'm asking because we merged https://github.com/algorand/pyteal/pull/429 with the intention to revisit user guide docs here.  And I'm trying to gauge opinion on the topic.
* I'm open to being convinced otherwise, but my starting position is:  The concept is sufficiently important to warrant a 1st class user guide mention.
  * Also flexible:  I can imagine extending https://pyteal.readthedocs.io/en/stable/versions.html or reworking https://pyteal.readthedocs.io/en/stable/compiler_optimization.html to include optimizations + version pragma.

jasonpaulos: Added documentation to the versions page, please re-open if anything is missing

docs/abi.rst
jasonpaulos: Good point. I'm opting to link to https://developer.algorand.org/docs/get-details/parameter_tables/#smart-contract-constraints since it'll be more user-friendly than the actual code

docs/abi.rst
jasonpaulos: Thanks for the suggestions, I've adopted a variation on this

docs/abi.rst
michaeldiamant: @jasonpaulos Re-opening for a final sanity check.  If the PR is in the expected form, then disregard.

From your prior note, I expected the hard-coded values to be removed.  It's unclear to me if you left them deliberately or incidentally.  

If incidental, also consider the same treatment in transaction type limitations (https://github.com/algorand/pyteal/pull/400/files#diff-d5f6d88ad06456a3ba42bcda3171b8d55309a3d87ad250a7e78379a322527d63R510) where max transaction group count is referenced.

michaeldiamant (approved):

docs/abi.rst
```diff
@@ -0,0 +1,916 @@
...
+Additionally, Reference Types are only valid in the arguments of a method. They may not appear in a method's return value.
+
+Note that the AVM has `limitations on the maximum number of foreign references <https://developer.algorand.org/docs/get-details/parameter_tables/#smart-contract-constraints>`_ an application call transaction may contain. At the time of writing, these limits are:
+
+* Accounts: 4
+* Assets: 8
+* Applications: 8
+* Sum of Accounts, Assets, and Applications: 8
+
+Because of this, methods that have a large amount of Reference Type arguments may be impossible to call as intended at runtime.
```
ahangsu, line 367: nit: just a stylistic suggestion, would you consider using `warning` or `note` in rst format to emphasize this paragraph?

docs/abi.rst
jasonpaulos: I've opted to leave the hard-coded values because I think they give a feel for the order of magnitude of the limits, plus they help illustrate the relationship between the per-field limits and the total reference limit. I hope that saying these values are only current at the time of writing and linking to a more authoritative source will be enough to future-proof this section.

michaeldiamant: @jasonpaulos Got it - thanks for the clarification.  Resolving - I'm open to proceeding as is.

docs/abi.rst
jasonpaulos: Good idea, I will wrap that in a warning

docs/abi.rst
```diff
@@ -0,0 +1,917 @@
...
+:any:`abi.AssetConfigTransaction`     :code:`acfg`           Static           An asset configuration transaction
+:any:`abi.AssetTransferTransaction`   :code:`axfer`          Static           An asset transfer transaction
+:any:`abi.AssetFreezeTransaction`     :code:`afrz`           Static           An asset freeze transaction
+:any:`abi.ApplicationCallTransaction` :code:`appl`           Static           An application call transaction
+===================================== ====================== ================ =======================================================================================================================================================
+
+Limitations
+""""""""""""""""""""""""""""""""""""""""""
+
+Due to the special meaning of Transaction Types, they cannot be used as the return value of a method. They can be used as method arguments, but only at the top-level. This means that it's not possible to embed a Transaction Type inside a tuple or array.
```
ahangsu, line 507: nit: emphasize `they **cannot** be used as the return ...`

ahangsu (approved):

PR #403, opened 2022-06-17
Extend abi-docs with experimental design language

michaeldiamant:
Extends #400 with warnings about level of support guarantees.  The PR attempts to mirror docstring language to be consistent with source code documentation.


jasonpaulos (approved):
Will merge as-is, but I may modify the wording in the parent PR.

PR #406, opened 2022-06-19
adding comment class as an experiment

barnjamin:
I'd like to annotate produced teal with some comments.

Adding a new class here to support wrapping some existing expression rather than forcing each expression implementation to accept a comment. We append the comment string to the last ops' args.

### DIFF
--- a/pyteal/ast/comment.py
+++ b/pyteal/ast/comment.py
@@ -1,0 +1,32 @@
+from typing import TYPE_CHECKING, Tuple
+
+from pyteal.types import TealType
+from pyteal.ir import TealBlock, TealSimpleBlock
+from pyteal.ast.expr import Expr
+
+if TYPE_CHECKING:
+    from pyteal.compiler import CompileOptions
+
+
+class Comment(Expr):
+    def __init__(self, expr: Expr, comment: str):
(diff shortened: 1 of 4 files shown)


pyteal/ast/comment.py
```diff
@@ -0,0 +1,32 @@
...
+        return tb, tsb
+
+    def __str__(self):
+        return "(Comment {} ({}))".format(self.comment, str(self.expr))
+
+    def type_of(self):
+        return TealType.none
+
+    def has_return(self):
+        return False
```
barnjamin, line 29: Should probably pass has return of expr

pyteal/ast/comment.py
```diff
@@ -0,0 +1,32 @@
...
+        tb, tsb = self.expr.__teal__(options)
+        # this seems not good, is there always going to oe an ops list? will it always have non None args?
+        tsb.ops[-1].args.append("// " + self.comment)
+        return tb, tsb
+
+    def __str__(self):
+        return "(Comment {} ({}))".format(self.comment, str(self.expr))
+
+    def type_of(self):
+        return TealType.none
```
barnjamin, line 26: Should pass type of expr

PR #408, opened 2022-06-21
docs: fix bytes/bit reference in Byteslice Arithmetic section

oliviera9:
Also, there is written that `Byteslice arithmetic operators allow up to 512-bit arithmetic.`: how is this possible?
All operators take 64 bit arguments...


zejoeh:
Hi @oliviera9, 1 byte is 8 bits, so 64 bytes = 512 bits. Byteslice arithmetic operates on arrays of bytes, and these arrays can each contain up to 64 individual bytes (hence 512 bits).

oliviera9:
I found it really strange the 64 bytes length of the args. Anyway, your explanation is perfectly clear, and this PR can be closed.

PR #409, opened 2022-06-21
adding any length tuple

barnjamin:
While trying to figure out the right way to define a tuple of length >6 I found that the typing_extensions package offers an `Unpack` for variadic length.  This PR appears to do what I want for something like:

```py
@router.method
def random_tuple(a: abi.TupleAny[abi.Uint16, abi.Uint64, abi.Uint64, abi.Uint64, abi.Uint64, abi.Uint64]):
    #...
```

We may consider removing the TupleN types in favor of this setup

It looks like mypy hates it, TypeVarTuples may not be bound yet either

https://peps.python.org/pep-0646/#variance-type-constraints-and-type-bounds-not-yet-supported

### DIFF
--- a/pyteal/ast/abi/__init__.py
+++ b/pyteal/ast/abi/__init__.py
@@ -30,6 +30,7 @@ from pyteal.ast.abi.tuple import (
     Tuple3,
     Tuple4,
     Tuple5,
+    TupleAny,
 )
 from pyteal.ast.abi.array_base import ArrayTypeSpec, Array, ArrayElement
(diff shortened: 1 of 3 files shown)


jasonpaulos:
The long-plan is to make the `Tuple` class use `TypeVarTuple`, so _if_ `typing_extensions` provides a `TypeVarTuple` that is adequate, I'd recommend modifying `Tuple` instead of introducing a new subclass.

But until mypy supports this feature (https://github.com/python/mypy/issues/12280), we may want to hold off, depending on how bad the mypy errors are.

PR #410, opened 2022-06-22
Comment as op

barnjamin:
Alternate impl to #407

for `Comment(Assert(Int(1)), "yep")`
```
#pragma version 6
int 1
assert
// yep
int 1
return
```


jasonpaulos (commented):
I also made a child PR which changes the `Comment` implementation: https://github.com/algorand/pyteal/pull/492

barnjamin:
With this pr, adding comments on asserted expressions, parsing the error out, and mapping it back to source teal I get the following output:

```
Txn WWVF5P2BXRNQDFFSGAGMCXJNDMZ224RJUGSMVPJVTBCVHEZMOMNA had error 'assert failed pc=883' at PC 883 and Source Line 579: 

        store 50
        store 49
        store 48
        store 47
        // correct asset a
        load 50
        txnas Assets
        bytec_0 // "a"
        app_global_get
        ==
        assert          <-- Error
        // correct asset b
        load 51
        txnas Assets
        bytec_1 // "b"
        app_global_get
        ==
        assert
        // correct pool token
        load 49
```

This is v close to what I'd want but the comment is a little far from the actual assert. I think the ideal output is that the comment is on the same line as (or just above) the assert so its easy to map it back. wdyt?

jasonpaulos (approved):
The specific case of `Assert` aside, I think this is a good addition by itself.

Note: probably we should get another approval before merging, as I also contributed to this PR.

ahangsu (approved):
generally lookin good, and I like the PR number, for that is my birthday.

ahangsu (commented):
Notice, should also add an entry in `CHANGELOG.md`~~, and consider a bit on #509~~

jasonpaulos (approved):

ahangsu (approved):
free approval bot says approve

PR #411, opened 2022-06-22
__str__ formatting

barnjamin:
When printing out the Expression tree, some of the Exprs had slightly different formatting. This just tries to convert them all to standard formatting of `(ExprNameAsTitleCase arg arg)`

### DIFF
--- a/pyteal/ast/abi/type.py
+++ b/pyteal/ast/abi/type.py
@@ -144,12 +144,9 @@ class BaseType(ABC):
 
     def __str__(self) -> str:
         at = self.type_spec().annotation_type()
-
-        # Check if the type is the same as the annotation type
-        # This suggests it is not generic and we can just use the class name
-        if type(self) is at:
+        # If the annotation type itself is a class, just return its name
+        if isinstance(at, type):
             return self.__class__.__name__
-
         return str(at)
 
 
--- a/pyteal/ast/abi/type.py
+++ b/pyteal/ast/abi/type.py
@@ -142,9 +142,8 @@ class BaseType(ABC):
             )
         return value.store_into(self)
 
-    def __str__(self)->str:
+    def __str__(self) -> str:
         return str(self.__class__.__name__)
-
 
 
 BaseType.__module__ = "pyteal"
--- a/pyteal/ast/abi/type.py
+++ b/pyteal/ast/abi/type.py
@@ -143,7 +143,7 @@ class BaseType(ABC):
         return value.store_into(self)
 
     def __str__(self) -> str:
-        return str(self.__class__.__name__)
+        return str(self.type_spec().annotation_type())
 
 
 BaseType.__module__ = "pyteal"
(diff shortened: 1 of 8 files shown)


pyteal/ast/abi/type.py
```diff
@@ -142,6 +142,9 @@ def _set_with_computed_type(self, value: "ComputedValue") -> Expr:
             )
         return value.store_into(self)
 
+    def __str__(self) -> str:
+        return str(self.__class__.__name__)
+
```
ahangsu, line 147: nit: seems a bit off when you are trying

```python
>>> from pyteal import *
>>> a = abi.type_spec_from_annotation(abi.DynamicArray[abi.Uint32])
>>> a = a.new_instance()
>>> a.__class__.__name__
'DynamicArray'
```

ahangsu (commented):
just a quick pass

ahangsu (commented):
I would approve if you could add a check under `abi/util_test.py/test_abi_type_translation`, to see `str` for `BaseType` is equal to the expected strings.

`BaseType` instances can be generated from `TypeSpec` specified in testcase.

barnjamin:
@ahangsu added to the tests, found that Bool was missing the `__module__` setting.

One weird bit is that with dynamic types the full `$module.$classname[$module.$classname]` is returned whereas with simple classes the module is omitted.

pyteal/ast/abi/type.py
```diff
@@ -142,6 +142,13 @@ def _set_with_computed_type(self, value: "ComputedValue") -> Expr:
             )
         return value.store_into(self)
 
+    def __str__(self) -> str:
+        at = self.type_spec().annotation_type()
```
ahangsu, line 146: this line is somewhat confusing, why not just returning `str(self.type_spec())`? Then we can guarantee that the `__str__` in type is exactly behaving the same as the one in type-spec.

ahangsu:
```
>>> from pyteal import *
>>> a = abi.AddressTypeSpec()
>>> str(a)
'address'
>>> a = abi.type_spec_from_annotation(abi.DynamicArray[abi.Uint32])
>>> str(a)
'uint32[]'
>>> from typing import *
>>> a = abi.type_spec_from_annotation(abi.StaticArray[abi.Uint32, Literal[10]])
>>> str(a)
'uint32[10]'
```

doubt it?

barnjamin: right it spits out the abi type, which may be fine?

barnjamin: yea, seems fine just tweaked it and updated tests

pyteal/ast/abi/util_test.py
```diff
@@ -489,7 +494,6 @@ def test_size_of():
     #         ]
     #     ),
     # ),
-    (algosdk.abi.TupleType([]), "()", abi.TupleTypeSpec(), abi.Tuple0),
```
ahangsu, line 492: nit: this line should stay?

barnjamin: I think its a dupe of line 387

ahangsu (approved):

tzaffi (approved):
LGTM

PR #412, opened 2022-06-22
adding program page related ops

barnjamin:
Adds ops and expressions related to https://github.com/algorand/go-algorand/pull/4148

### DIFF
--- a/pyteal/ast/txn.py
+++ b/pyteal/ast/txn.py
@@ -110,6 +110,22 @@ class TxnField(Enum):
     created_application_id = (61, "CreatedApplicationID", TealType.uint64, False, 5)
     last_log = (62, "LastLog", TealType.bytes, False, 6)
     state_proof_pk = (63, "StateProofPK", TealType.bytes, False, 6)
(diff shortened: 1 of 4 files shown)


ahangsu (approved):
lgtm, I think for the `txn_test.py` it is quite modular, such that you can just add new fields to start the test

pyteal/ast/txn.py
```diff
@@ -110,6 +110,22 @@ class TxnField(Enum):
     created_application_id = (61, "CreatedApplicationID", TealType.uint64, False, 5)
     last_log = (62, "LastLog", TealType.bytes, False, 6)
     state_proof_pk = (63, "StateProofPK", TealType.bytes, False, 6)
+    approval_program_pages = (64, "ApprovalProgramPages", TealType.bytes, True, 7)
```
michaeldiamant, line 113: @barnjamin Minor:  Can you extend https://pyteal.readthedocs.io/en/stable/accessing_transaction_field.html to cover the new fields?

Based on the existing grouping, I'm thinking https://pyteal.readthedocs.io/en/stable/accessing_transaction_field.html#application-call is the best fit.

PR #413, opened 2022-06-23
Add Replace

jdtzmn:
This PR adds the upcoming AVM 7 `replace2` and `replace3` opcodes.


pyteal/ir/ops.py
```diff
@@ -1,3 +1,4 @@
+from dataclasses import replace
```
barnjamin, line 1: intellisense got too excited?

pyteal/__init__.pyi
```diff
@@ -138,6 +138,7 @@ __all__ = [
     "Or",
     "Pop",
     "Reject",
+    "Replace",
```
michaeldiamant, line 141: @jdtzmn Unrelated to line:  While reviewing the PR, I added a couple of tests in https://github.com/jdtzmn/pyteal/pull/1.

Can you see if you feel it's worth pulling into the PR?  Otherwise, we can close the child PR.

PR #415, opened 2022-06-24
Add Block

jdtzmn:
This PR adds support for the `block` opcode in AVM 7.


pyteal/ast/block.py
```diff
@@ -0,0 +1,80 @@
...
+
+BlockField.__module__ = "pyteal"
+
+
+class Block(LeafExpr):
+    """An expression that accesses a block property."""
+
+    def __init__(self, field: BlockField, block: Expr) -> None:
+        super().__init__()
+        self.field = field
```
barnjamin, line 40: do we need to check that field is indeed a BlockField?

jdtzmn: My assumption was that `Block` wouldn't be used directly but rather through the `Block.timestamp()` methods. I'm not opposed, I was just following the convention I'd seen in other files. I previously had an explicit check in `JsonRef` because passing the property was the expected behavior in that case.

michaeldiamant (approved):
@jdtzmn Looks accurate to me - thanks for helping to bring PyTeal up to AVM 7 parity!

barnjamin (approved):

PR #417, opened 2022-06-24
Add JsonRef

jdtzmn:
This PR adds the upcoming AVM 7 `json_ref` opcode.


pyteal/ast/jsonref.py
```diff
@@ -0,0 +1,77 @@
...
+if TYPE_CHECKING:
+    from pyteal.compiler import CompileOptions
+
+
+class JsonRefType(Enum):
+    # fmt: off
+    #              id |    name   |      type      | min version
+    json_string = (0, "JSONString", TealType.bytes,  7)
+    json_uint64 = (1, "JSONUint64", TealType.uint64, 7)
+    json_object = (2, "JSONObject", TealType.bytes,  7)
```
barnjamin, line 19: Since these are already namespaced under JsonRefType does it make sense to omit `json_`?

pyteal/ast/jsonref.py
```diff
@@ -0,0 +1,77 @@
...
+
+        op = TealOp(self, Op.json_ref, self.type.arg_name)
+        return TealBlock.FromOp(options, op, self.json_obj, self.key)
+
+    def __str__(self):
+        return "(JsonRef {})".format(self.type.arg_name)
+
+    def type_of(self):
+        return self.type.type_of()
+
```
barnjamin, line 75: I like how you had the base64 `.std/.url`, does it make sense to do something similar here?

pyteal/ast/jsonref.py
barnjamin: if you implement the methods for these instead, the naming matters less I think

pyteal/ast/jsonref.py
jdtzmn: It very well might. I could easily be convinced. My thinking is that if I saw the following:
```py
JsonRef.string(Bytes('{"a":"b"}'), Bytes("a"))
```
I wouldn't be sure what the `.string` property meant at first glance. Without a thorough understanding of the `json_ref` opcode I wouldn't immediately understand whether string refers to the key type, or the input type, or the value's decoded type.

My hope was that explicitly having to set the format would make it clearer at first glance.

```py
JsonRef(JsonRef.json_string, Bytes('{"a":"b"}'), Bytes("a"))
```

However, as I'm writing this, I'm thinking that named functions `.as_string`, `.as_uint64`, and `.as_object` would solve both issues. I'll add those 👍

barnjamin (approved):

pyteal/ast/jsonref.py
jdtzmn: @barnjamin I'm having second thoughts about this. As I work more with the PyTeal library, it seems that the "PyTeal way" of doing things is to use dot notation for accessing _properties_ but passed parameters for encodings or specifying return value types. For example, you see `AssetParam.total()` to access a property of an asset param, but ` EcdsaVerify(EcdsaCurve.Secp256k1, *args)` for ecdsa verification.

I actually think using dot notation for everything could provide a better developer experience and would probably be in favor of adopting that method moving forward. However, if we were trying to stay consistent with existing behavior, we should get rid of `.std/.url/.as_string/etc` in favor of exported enums as function parameters.

pyteal/ast/jsonref.py
```diff
@@ -0,0 +1,107 @@
...
+        return self.ret_type
+
+
+JsonRefType.__module__ = "pyteal"
+
+
+class JsonRef(LeafExpr):
+    """An expression that accesses the value associated with a given key from a supported utf-8 encoded json object.
+
+    The json object must satisfy a [particular specification](https://github.com/algorand/go-algorand/blob/master/data/transactions/logic/jsonspec.md).
```
michaeldiamant, line 38: @jdtzmn Minor:  Based on https://devguide.python.org/documenting/#external-links, I think the external link ought to be formatted differently.

```suggestion
    The json object must satisfy a `particular specification <https://github.com/algorand/go-algorand/blob/master/data/transactions/logic/jsonspec.md>`_.
```

pyteal/ast/jsonref.py
```diff
@@ -0,0 +1,107 @@
...
+    """
+
+    def __init__(self, type: JsonRefType, json_obj: Expr, key: Expr) -> None:
```
michaeldiamant, line 41: @jdtzmn Minor and optional:  There's 2 places where a parameter is named `type`.  If you're on-board, I propose renaming because there's a builtin `type` available by default.  

The parameters shadow the builtin `type` definition.  In my experience, lexical shadowing is error prone.  

That said, I don't feel too strongly.  If you prefer as is, feel welcomed to resolve the thread.

pyteal/ast/jsonref.py
michaeldiamant: My lightly held 2c:  I'm on-board with the PR's approach.  If another reviewer has a strong opinion otherwise, I can be convinced to revert for consistency.

michaeldiamant (approved):
@jdtzmn Appreciate the effort here - Leaving approval under the assumption that threads I opened are resolved prior to merging.

pyteal/ast/jsonref.py
michaeldiamant: @jdtzmn Got it RE:  VSCode.  I'm in Pycharm, which _does_ render the link differently.

Additionally - Since API docstrings are added to https://pyteal.readthedocs.io/en/stable/api.html, you can observe the difference by generating the docs and viewing the HTML files via web browser.

pyteal/ast/jsonref.py
jdtzmn: If it were the only location in the codebase which used `type` as a parameter, I'd feel differently. But since it appears in many other places I'll resolve the conversation instead.

michaeldiamant (approved):

PR #421, opened 2022-06-27
MaybeValue error prior to scratch slot check

barnjamin:
During compilation we should know to error on an uninitialized MaybeValue prior to checking if the scratch slots are assigned. This attempts to add some flag to tell us if it's been done.

before:
```
pyteal.TealCompileError: Scratch slot load occurs before store
Traceback of origin expression (most recent call last):
  File "/home/ben/pyteal/examples/application/maybe_test.py", line 5, in <module>
    AssetHolding.balance(Global.zero_address(), Int(0)).value(),
  File "/home/ben/pyteal/pyteal/ast/maybe.py", line 46, in value
    return self.output_slots[0].load(self.types[0])
  File "/home/ben/pyteal/pyteal/ast/scratch.py", line 60, in load
  File "/home/ben/pyteal/pyteal/ast/scratch.py", line 117, in __init__
    super().__init__()


The above exception was the direct cause of the following exception:

  File "/home/ben/pyteal/examples/application/maybe_test.py", line 8, in <module>
    print(compileTeal(program, mode=Mode.Application, version=6))
  File "/home/ben/pyteal/pyteal/compiler/compiler.py", line 252, in compileTeal
    localSlotAssignments = assignScratchSlotsToSubroutines(subroutine_start_blocks)
  File "/home/ben/pyteal/pyteal/compiler/scratchslots.py", line 148, in assignScratchSlotsToSubroutines
    raise TealInternalError(msg) from errors[0]
pyteal.TealInternalError: Encountered 1 error when assigning slots to subroutine
```

after:
```
  File "/home/ben/pyteal/examples/application/maybe_test.py", line 5, in <module>
    AssetHolding.balance(Global.zero_address(), Int(0)).value(),
  File "/home/ben/pyteal/pyteal/ast/maybe.py", line 52, in value
    raise TealInputError(
pyteal.TealInputError: Expected MaybeValue to be initialized prior to calling `value`
```

### DIFF
--- a/pyteal/ast/maybe.py
+++ b/pyteal/ast/maybe.py
@@ -36,9 +36,13 @@ class MaybeValue(MultiValue):
 
         This will return 1 if the value exists, otherwise 0.
         """
-        if not self.initialized:
-            raise TealInputError("Expected MaybeValue to be initialized before calling `hasValue`")
-        return self.output_slots[1].load(self.types[1])
+        try:
+            sl = self.output_slots[1].load(self.types[1])
+        except TealInputError:
+            raise TealInputError(
+                "Expected MaybeValue to be initialized prior to calling `hasValue`"
+            )
+        return sl
 
     def value(self) -> ScratchLoad:
         """Get the value.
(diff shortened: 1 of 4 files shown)


barnjamin:
failing lots of tests because we do stuff like this:

```py
    def test_acct_param_balance_valid():
        arg = pt.Int(1)
        expr = pt.AccountParam.balance(arg)
        assert expr.type_of() == pt.TealType.none
       assert expr.value().type_of() == pt.TealType.uint64
```

ahangsu:
@barnjamin Ben, does that example still incur the error? My local test does not reproduce the error.

PR #422, opened 2022-06-27
Router return subroutine

barnjamin:
If I declare a method like 

```py
@router.method
def thing():
    return Int(1)
```

And I want to get the method selector or signature, I need to filter the `methods` array for those methods with the properties I'm interested in 

```py
thingmethod = list(filter(lambda m: m.name == "thing", router.methods)).pop()
signature = thingmethod.get_signature()
selector = Bytes(thingmethod.get_selector())
```

With this change I can call `thing.method_selector()` directly although the type hinting doesn't work very well.

### DIFF
--- a/pyteal/compiler/compiler_test.py
+++ b/pyteal/compiler/compiler_test.py
@@ -1,3 +1,4 @@
+from pyparsing import empty
 import pytest
 
 import pyteal as pt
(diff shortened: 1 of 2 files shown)


pyteal/compiler/compiler_test.py
```diff
@@ -1,3 +1,4 @@
+from pyparsing import empty
```
tzaffi, line 1:
```suggestion
```

ahangsu: yea, VSCode is getting too excited

PR #423, opened 2022-06-27
Support Secp256r1 curve

jdtzmn:
This PR adds the `Secp256r1` curve available in AVM 7.

* Converted `EcdsaDecompress` and `EcdsaRecover` to classes so that field version checking was possible.

### DIFF
--- a/pyteal/ast/ecdsa.py
+++ b/pyteal/ast/ecdsa.py
@@ -2,7 +2,7 @@ from enum import Enum
 from typing import Tuple, TYPE_CHECKING
 
 from pyteal.ast import Expr, MultiValue
-from pyteal.errors import TealTypeError, verifyTealVersion
+from pyteal.errors import TealTypeError, verifyFieldVersion, verifyTealVersion
 from pyteal.ir import Op, TealBlock, TealOp
 from pyteal.types import TealType, require_type
 
@@ -17,6 +17,7 @@ class EcdsaCurve(Enum):
     """Enum representing an elliptic curve specification used in ECDSA."""
 
     Secp256k1 = (0, "Secp256k1", 5)
+    Secp256r1 = (1, "Secp256r1", 7)
 
     def __init__(self, id: int, name: str, min_version: int) -> None:
         self.id = id
@@ -52,10 +53,12 @@ class EcdsaVerifyExpr(Expr):
 
     def __teal__(self, options: "CompileOptions"):
         verifyTealVersion(
-            max(self.op.min_version, self.curve.min_version),
+            self.op.min_version,
             options.version,
             "TEAL version too low to use op {}".format(self.op),
         )
+
+        verifyFieldVersion(self.curve.arg_name, self.curve.min_version, options.version)
 
         return TealBlock.FromOp(
             options, TealOp(self, self.op, self.curve.arg_name), *self.args
@@ -111,7 +114,7 @@ def EcdsaVerify(
     return EcdsaVerifyExpr(curve, data, sigA, sigB, pubkey[0], pubkey[1])
 
 
-def EcdsaDecompress(curve: EcdsaCurve, compressed_pk: Expr) -> MultiValue:
+class EcdsaDecompress(MultiValue):
     """Decompress an ECDSA public key.
 
     Args:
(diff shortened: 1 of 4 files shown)


pyteal/ast/ecdsa_test.py
```diff
@@ -26,22 +26,30 @@ def test_ecdsa_decompress():
         ]
     )
 
-    actual, _ = pubkey.__teal__(teal5Options)
+    actual, _ = pubkey.__teal__(teal7Options)
```
michaeldiamant, line 29: @jdtzmn Let me know if the thinking makes sense or if you'd like to discuss live?

* I like the idea to parameterize the test though I think the old assertion against `teal5Options` must be retained for _Secp256k1_.
* As shown, it's possible to bump `Secp256k1.min_version` to 7 without the test failing.
* If we agree a gap exists, I think there's at least several ways to address it.
  * A local solution is to either branch the test based on the opcode or build a dictionary from version to options.
  * Eventually, I think a more systemic solution will be needed to vet _all_ versions.  But, for now, I think it's _safe_ to solve the issue local to `ecdsa_test.py`.

pyteal/ast/ecdsa.py
```diff
@@ -111,33 +114,39 @@ def EcdsaVerify(
...
-    curve: EcdsaCurve, data: Expr, recovery_id: Expr, sigA: Expr, sigB: Expr
-) -> MultiValue:
+    def __teal__(self, options: "CompileOptions"):
+        verifyFieldVersion(self.curve.arg_name, self.curve.min_version, options.version)
+
+        return super().__teal__(options)
+
+
+class EcdsaRecover(MultiValue):
     """Reover an ECDSA public key from a signature.
```
michaeldiamant, line 150: @jdtzmn Minor:  Unrelated to the changeset, but easy to address the typo.

```suggestion
    """Recover an ECDSA public key from a signature.
```

pyteal/ast/ecdsa.py
```diff
@@ -111,33 +114,39 @@ def EcdsaVerify(
     return EcdsaVerifyExpr(curve, data, sigA, sigB, pubkey[0], pubkey[1])
 
 
-def EcdsaDecompress(curve: EcdsaCurve, compressed_pk: Expr) -> MultiValue:
+class EcdsaDecompress(MultiValue):
```
michaeldiamant, line 117: @algoidurovic Sanity checking:  In https://github.com/algorand/pyteal/pull/307, do you recall if there's a by-design why _only_ `EcdsaVerifyExpr` performs minimum version checks?

From my read of https://github.com/algorand/pyteal/pull/307, it seems we simply missed it.

pyteal/ast/ecdsa.py
```diff
@@ -111,33 +114,39 @@ def EcdsaVerify(
...
-def EcdsaRecover(
-    curve: EcdsaCurve, data: Expr, recovery_id: Expr, sigA: Expr, sigB: Expr
-) -> MultiValue:
+    def __teal__(self, options: "CompileOptions"):
+        verifyFieldVersion(self.curve.arg_name, self.curve.min_version, options.version)
+
+        return super().__teal__(options)
+
+
+class EcdsaRecover(MultiValue):
```
michaeldiamant, line 149: @jdtzmn and reviewers:  Can we treat the method conversion to class as a non-breaking change?

* It _feels_ transparent though I'm not convinced.  For example, a `type` check is liable to fail due to the change in return type (https://stackoverflow.com/questions/1549801/what-are-the-differences-between-type-and-isinstance).
* A few options exist:
  * Accept as is without further change.
  * Accept as is + prefix PR title with _Breaking change:_ (or similar).  Since we use https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes to generate release notes, the disclaimer will be explicit.  
  * Modify `MultiValue` to support field version checks.  There's likely several approaches here.  One avenue is an optional constructor parameter accepting a callback function (`Callable[CompileOptions, None]]`) to be invoked in `__teal__`.
* I'll follow the group's opinion.  My somewhat loosely held opinion is to either prefix PR title or enrich `MultiValue`.  It feels unlikely to me that we'll pattern match `MutliValue` subclasses, so I'm lightly biased towards enriching `MultiValue`.

pyteal/ast/ecdsa.py
algoidurovic: If I remember correctly, it was because the version checking needed to happen within the MultiValue class (ie all uses of MultiValue/MaybeValue had this issue). We decided it would be better to address that problem in a separate PR but it must've slipped my mind.

pyteal/ast/ecdsa.py
algoidurovic: While it will take more work, I prefer to "modify `MultiValue` to support field version checks". The typical usage pattern of MultiValue isn't to extend it unless new fields/methods are needed, like `MaybeValue`'s `hasValue` and `value` methods. This allows for more lightweight implementations of new expressions that use `MultiValue`.

pyteal/ast/ecdsa.py
michaeldiamant: @jdtzmn Thanks for fixing the typo - resolving.

pyteal/ast/ecdsa.py
jdtzmn: Done!

pyteal/ast/ecdsa.py
michaeldiamant: @algoidurovic Thanks for the double check - resolving.

pyteal/ast/ecdsa.py
michaeldiamant: @jdtzmn Thanks for exposing `compile_check` - Resolving.

michaeldiamant (approved):

barnjamin (approved):

algoidurovic (approved):

PR #427, opened 2022-06-27
Check transaction type arguments

jasonpaulos:
This PR adds logic to the ABI router to verify that the type of provided transactions in the group match the declared type of any transaction arguments.

Prior to this PR, a method could be declared as `example(pay)void`, but there would be no checks to ensure that the transaction argument was a payment transaction at runtime.

The only transaction type that is unchanged is the generic `txn` type, since it requires no verification.


ahangsu (approved):
lgtm, there are also manual testing in `router_test`, feels safe to me.

barnjamin (approved):

PR #431, opened 2022-07-01
fix misspelling of uint

barnjamin:


jasonpaulos (approved):
If the underlying field name were not also invalid, I'd recommend keeping the original function for backwards compatibility. But since that's not the case, no one can use `localNumUnit` correctly today, so a rename is acceptable.

Thanks for fixing.

PR #432, opened 2022-07-01
Check `Expr` length when converting to `abi.Address`

ahangsu:
Add a line of runtime length check for converting `Expr` (bytes) to `abi.Address`.


pyteal/ast/abi/address.py
```diff
@@ -95,7 +99,10 @@ def set(
                     f"Got bytes with length {len(value)}, expected {AddressLength.Bytes}"
                 )
             case Expr():
-                return self.stored_value.store(value)
+                return Seq(
+                    Assert(Len(value) == Int(AddressLength.Bytes.value)),
+                    self.stored_value.store(value),
+                )
             case CollectionSequence():
```
barnjamin, line 106: should we check the length of this as well?

ahangsu: hole up let me see... I think this check can be done on python level, right?

Namely, `len(value) == AddressLength.Bytes.value`?

ahangsu: nvm, adding a few lines of checks

barnjamin (approved):

pyteal/ast/abi/address.py
```diff
@@ -95,8 +99,15 @@ def set(
                     f"Got bytes with length {len(value)}, expected {AddressLength.Bytes}"
                 )
             case Expr():
-                return self.stored_value.store(value)
+                return Seq(
+                    Assert(Len(value) == Int(AddressLength.Bytes.value)),
+                    self.stored_value.store(value),
```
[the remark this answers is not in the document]
jasonpaulos: No need to change history, the follow up PR you made is a good approach

PR #433, opened 2022-07-01
Fix on previous address length check

ahangsu:
#432 on Address length `Expr` case generated wrong TEAL code. This PR fixes that issue.


jasonpaulos (approved):
Pending tests, LGTM. Thanks for fixing after the fact

PR #434, opened 2022-07-01
Typo fixing

ahangsu:
Typo fixing catch all from #431


tzaffi (approved):
LGTM - even though this is a breaking change for anyone who is using `globalNumUnit()` directly on `AppParam`

jasonpaulos (approved):

barnjamin:
@tzaffi I don't expect many (any?) people are using this since the resultant TEAL won't actually compile to binary

PR #435, opened 2022-07-06
Clarify minimum Python version management docs

michaeldiamant:
Updates Python version management docs to explicitly highlight the need to manage parallel Python versions rather than overriding the operating system-provided Python version.


algoanne (approved):
looks good

PR #440, opened 2022-07-13
Foreign prefix on App and Asset arrays

barnjamin:
It was pointed out that these should be renamed to the unprefixed version


michaeldiamant (approved):
@barnjamin Looks like you've caught all references.  `git grep -i foreign` does _not_ show more candidate references.

ahangsu (approved):

PR #442, opened 2022-07-15
Simplify multi-expression clauses

jdtzmn:
This PR simplifies multi-expression clauses by using `Seq` internally when more than one expression is given.

#### If/Then/Else

```python
If(x).Then(y1, y2, y3).Else(z1, z2, z3)
# is equivalent to
If(x).Then(Seq(y1, y2, y3)).Else(Seq(z1, z2, z3))
```

#### While/Do

```python
While(x).Do(x1, x2, x3)
# is equivalent to
While(x).Do(Seq(x1, x2, x3)
```

#### For/Do

```python
For(x, y, z).Do(w1, w2, w3)
# is equivalent to
For(x, y, z).Do(Seq(w1, w2, w3))
```

#### Cond

```python
Cond([c1, x1, x2, x3],
     [c2, y1, y2, y3])
# is equivalent to
Cond([c1, Seq(x1, x2, x3)],
     [c2, Seq(y1, y2, y3)])
```

#### Assert

```python
Assert(x, y, z)
# is equivalent to
Seq(Assert(x), Assert(y), Assert(z))
```

### DIFF
--- a/pyteal/ast/assert_.py
+++ b/pyteal/ast/assert_.py
@@ -12,18 +12,18 @@ if TYPE_CHECKING:
 class Assert(Expr):
     """A control flow expression to verify that a condition is true."""
 
-    def __init__(self, cond: Expr, *additionalConds: Expr) -> None:
+    def __init__(self, cond: Expr, *additional_conds: Expr) -> None:
         """Create an assert statement that raises an error if the condition is false.
 
         Args:
             cond: The condition to check. Must evaluate to a uint64.
-            *additionalConds: Additional conditions to check. Must evaluate to uint64.
+            *additional_conds: Additional conditions to check. Must evaluate to uint64.
         """
         super().__init__()
         require_type(cond, TealType.uint64)
-        for cond_single in additionalConds:
+        for cond_single in additional_conds:
             require_type(cond_single, TealType.uint64)
-        self.cond = [cond] + list(additionalConds)
+        self.cond = [cond] + list(additional_conds)
 
     def __teal__(self, options: "CompileOptions"):
         if len(self.cond) > 1:
--- a/pyteal/ast/assert_.py
+++ b/pyteal/ast/assert_.py
@@ -1,4 +1,4 @@
-from typing import TYPE_CHECKING, List
+from typing import TYPE_CHECKING
 
 from pyteal.types import TealType, require_type
 from pyteal.ir import TealOp, Op, TealBlock, TealSimpleBlock, TealConditionalBlock
@@ -12,7 +12,7 @@ if TYPE_CHECKING:
 class Assert(Expr):
     """A control flow expression to verify that a condition is true."""
 
-    def __init__(self, cond: Expr, *condMulti: List[Expr]) -> None:
+    def __init__(self, cond: Expr, *condMulti: list[Expr]) -> None:
         """Create an assert statement that raises an error if the condition is false.
 
         Args:
--- a/pyteal/ast/assert_.py
+++ b/pyteal/ast/assert_.py
@@ -12,7 +12,7 @@ if TYPE_CHECKING:
 class Assert(Expr):
     """A control flow expression to verify that a condition is true."""
 
-    def __init__(self, cond: Expr, *condMulti: list[Expr]) -> None:
+    def __init__(self, cond: Expr, *condMulti: Expr) -> None:
         """Create an assert statement that raises an error if the condition is false.
 
         Args:
--- a/pyteal/ast/assert_.py
+++ b/pyteal/ast/assert_.py
@@ -17,6 +17,7 @@ class Assert(Expr):
 
         Args:
             cond: The condition to check. Must evaluate to a uint64.
+            *additionalConds: Additional conditions to check. Must evaluate to uint64.
         """
         super().__init__()
         require_type(cond, TealType.uint64)
--- a/pyteal/ast/assert_.py
+++ b/pyteal/ast/assert_.py
@@ -12,7 +12,7 @@ if TYPE_CHECKING:
 class Assert(Expr):
     """A control flow expression to verify that a condition is true."""
 
-    def __init__(self, cond: Expr, *condMulti: Expr) -> None:
+    def __init__(self, cond: Expr, *additionalConds: Expr) -> None:
         """Create an assert statement that raises an error if the condition is false.
 
         Args:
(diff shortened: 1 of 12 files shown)


pyteal/ast/assert_.py
```diff
@@ -11,23 +12,29 @@
 class Assert(Expr):
     """A control flow expression to verify that a condition is true."""
 
-    def __init__(self, cond: Expr) -> None:
+    def __init__(self, cond: Expr, *condMulti: List[Expr]) -> None:
```
ahangsu, line 15: nit: I wonder what if we just keep only one argument: `__init__(self, *conds: Expr): -> None`? I think this is not breaking the old code, wile still introducing the new feature of multi-expression clause.

ahangsu: also, the type annotation seems off, maybe you just want `*conds: Expr`? by https://peps.python.org/pep-0484/#arbitrary-argument-lists-and-default-argument-values

pyteal/ast/cond.py
```diff
@@ -9,6 +10,24 @@
...
+
+    Example:
+        [ [a, b, c], [d, e, f] ] -> [ [a, Seq(b, c)], [d, Seq(e, f)] ]
+    """
+    reformatted = []
+    for arg in argv:
+        # Note: this is not a valid Cond arg, but will be caught later in Cond.__init__()
+        if len(arg) <= 1:
+            reformatted.append(arg)
+        else:
```
ahangsu, line 25: nit: is there any possible case that `len(arg) <= 1`?

ahangsu (commented):
took a quick pass

pyteal/ast/assert_.py
jdtzmn: Thanks for pointing out the type annotation error! I expected the behavior to be similar to JavaScript's. I thought adding an argument list instead of replacing the existing one gave better suggestions for recommended usage.

With `__init__(self, *conds: Expr): -> None`, the user sees no type error if they were to write `Assert()`. However, if the first positional argument is included, it suggests that the first argument is required and the remaining ones are optional – `Assert()` would show a warning because it's missing a required argument.

pyteal/ast/assert_.py
```diff
@@ -11,23 +12,29 @@
...
+            return Seq(*asserts).__teal__(options)
+
         if options.version >= Op.assert_.min_version:
             # use assert op if available
-            return TealBlock.FromOp(options, TealOp(self, Op.assert_), self.cond)
+            return TealBlock.FromOp(options, TealOp(self, Op.assert_), self.cond[0])
 
         # if assert op is not available, use branches and err
-        condStart, condEnd = self.cond.__teal__(options)
+        condStart, condEnd = self.cond[0].__teal__(options)
```
ahangsu, line 37: nit: the version check of `Op.assert_` is not imposed over case `len(self.cond) > 1`.

pyteal/ast/seq.py
```diff
@@ -83,3 +83,25 @@ def has_return(self):
...
+def use_seq_if_multiple(*exprs):
+    """If multiple expressions are provided, wrap them in a Seq expression."""
+
+    # Handle case where a list of expressions is provided
+    if len(exprs) == 1 and isinstance(exprs[0], list):
+        exprs = exprs[0]
+
+    if len(exprs) > 1:
+        return Seq(*exprs)
+    return exprs[0]
```
ahangsu, line 107: nit: shall we consider a case to cover `len(exprs) == 0`?

pyteal/ast/seq.py
```diff
@@ -83,3 +83,25 @@ def has_return(self):
 
 
 Seq.__module__ = "pyteal"
+
+
+@overload
+def use_seq_if_multiple(exprs: list[Expr]) -> Union[Seq, Expr]:
```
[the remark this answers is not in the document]
ahangsu: After a second thought, I don't think the union type makes sense here. `Seq` is inherited from `Expr`, then the type evaluates to only `Expr`.

ahangsu (commented):
another quik pass

pyteal/ast/seq.py
```diff
@@ -104,4 +104,6 @@ def use_seq_if_multiple(*exprs):
 
     if len(exprs) > 1:
         return Seq(*exprs)
+    if len(exprs) == 0:
+        return None
```
jdtzmn, line 108: Should this instead throw an error?

ahangsu: I guess it should?

jasonpaulos (commented):
Implementation looks good, but the docstrings should be modified to describe the new behavior

pyteal/ast/assert_.py
ahangsu: 👍

jasonpaulos (approved):

ahangsu (approved):

pyteal/ast/cond.py
```diff
@@ -9,16 +10,34 @@
     from pyteal.compiler import CompileOptions
 
 
+def reformat_multi_argv(argv: tuple[list[Expr], ...]) -> list[list[Expr]]:
```
michaeldiamant, line 13: @jdtzmn Minor:  Do you intend for `reformat_multi_argv` to be used by library users?

My sense is _no_.  If the method is intended for internal usage, can you rename to `_reformat_multi_argv`?

The leading `_` signifies internal usage as discussed in https://peps.python.org/pep-0008/#descriptive-naming-styles (see `_single_leading_underscore`).

I'm nitpicking because once exposed publicly, we cannot easily make changes.

jasonpaulos: To expand on this topic with my personal views:
* I view it as _essential_ to use a leading underscore for a non-public function name that is exported by the `pyteal` package. Commonly this would be static or instance methods on classes which are not designed for public use.
* I don't think it's as important for non-public functions which are not exported by the `pyteal` package (and therefore users can't access them) to use a leading underscore. I would never consider this wrong, as the underscore does help indicate the private nature of such a function to us developers, but I don't view it as necessarily required.

michaeldiamant: @jdtzmn Thanks for bringing in the changes - resolves the thread from my perspective.

@jasonpaulos I agree with the rationale.  Let's continue discussion out-of-band to the PR.

pyteal/ast/assert_.py
```diff
@@ -11,23 +12,30 @@
 class Assert(Expr):
     """A control flow expression to verify that a condition is true."""
 
-    def __init__(self, cond: Expr) -> None:
+    def __init__(self, cond: Expr, *additionalConds: Expr) -> None:
```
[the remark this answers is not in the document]
michaeldiamant: @jdtzmn Thanks for making the naming convention changes - resolving.

michaeldiamant (approved):
@jdtzmn Thanks for taking in my feedback late in the game.  

I don't want to hold up the PR, but I think the user guide benefits from an update to reflect the syntax improvements.

I'm happy to merge the PR as is.  And I've opened a separate PR to provide space to discuss user guide changes:  https://github.com/jdtzmn/pyteal/pull/3.

PR #444, opened 2022-07-16
Add Execute Method

barnjamin:
A very common use case is to fire off a single transaction, we can make this even easier by offering an `Execute` method that accepts the same arguments as `SetFields` and performing the Begin/Submit 

```py
Seq(
    InnerTxnBuilder.Begin(),
    InnerTxnBuilder.SetFields(
        {
            TxnField.type_enum: TxnType.ApplicationCall,
            TxnField.application_id: OpUp.opup_app_id,
        }
    ),
    InnerTxnBuilder.Submit(),
)
```
becomes
```py
 InnerTxnBuilder.Execute({
    TxnField.type_enum: TxnType.ApplicationCall,
    TxnField.application_id: OpUp.opup_app_id,
})
```

Not sold on name, maybe `Perform`?

### DIFF
--- a/pyteal/ast/itxn.py
+++ b/pyteal/ast/itxn.py
@@ -182,7 +182,23 @@ class InnerTxnBuilder:
 
     @classmethod
     def Execute(cls, fields: Dict[TxnField, Union[Expr, List[Expr]]]) -> Expr:
-        """Performs a single transaction given fields passed in"""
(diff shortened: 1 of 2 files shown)


barnjamin:
needs tests

jasonpaulos:
If you add just one simple test, this looks mergable to me

jasonpaulos (approved):

PR #447, opened 2022-07-18
Trace comments

barnjamin:
Running `examples/application/debug/contract.py` yields

```
#pragma version 6
int 1           // pyteal-src "/home/ben/pyteal/examples/application/debug/contract.py": line 6
int 2           // pyteal-src "/home/ben/pyteal/examples/application/debug/contract.py": line 6
callsub thing_0 // pyteal-src "/home/ben/pyteal/examples/application/debug/contract.py": line 6
pop
byte "ok"
len
int 0           // pyteal-src "/home/ben/pyteal/examples/application/debug/contract.py": line 8
>
assert          // pyteal-src "/home/ben/pyteal/examples/application/debug/contract.py": line 8
int 2           // pyteal-src "/home/ben/pyteal/examples/application/debug/contract.py": line 10
int 3           // pyteal-src "/home/ben/pyteal/examples/application/debug/contract.py": line 10
callsub thing_0 // pyteal-src "/home/ben/pyteal/examples/application/debug/contract.py": line 10
return

// thing
thing_0:        // pyteal-src "/home/ben/pyteal/examples/application/debug/util.py": line 5
+
retsub
```

Note: This would require every Expr to pass its `trace` in the `__teal__` method

If this is not atrocious I can start adding the trace to each Expr


pyteal/ast/int.py
```diff
@@ -30,7 +30,7 @@ def __init__(self, value: int) -> None:
 
     def __teal__(self, options: "CompileOptions"):
         op = TealOp(self, Op.int, self.value)
-        return TealBlock.FromOp(options, op)
+        return TealBlock.FromOp(options, op, traceback=self.trace)
```
barnjamin, line 33: Forced to pass traceback into blocks or ops, so this would need to happen for every Expr

pyteal/ast/subroutine.py
```diff
@@ -288,9 +290,11 @@ def invoke(
                         f"should have ABI typespec {arg_type} but got {arg.type_spec()}"
                     )
 
-        return SubroutineCall(
+        sc = SubroutineCall(
             self, args, output_kwarg=OutputKwArgInfo.from_dict(self.output_kwarg)
         )
+        sc.trace = trace
+        return sc
```
barnjamin, line 297: This is one I should have updated with the error traces

pyteal/ir/tealsimpleblock.py
```diff
@@ -7,10 +7,14 @@
 class TealSimpleBlock(TealBlock):
     """Represents a basic block of TealComponents in a graph that does not contain a branch condition."""
 
-    def __init__(self, ops: List[TealOp]) -> None:
+    def __init__(self, ops: List[TealOp], traceback: list[str] = []) -> None:
```
ahangsu, line 10: nit: do not use `[]` as default argument, they are **mutable default argument**. Please consider using `None`

pyteal/compiler/compiler.py
```diff
@@ -308,6 +308,16 @@ def compileTeal(
...
+            if idx <= 0:
+                return teal_line
+        except Exception:
+            return teal_line
+
+        return teal_line[:idx].strip(" \t") + "\t" * 2 + teal_line[idx:]
+
     lines = ["#pragma version {}".format(version)]
-    lines += [i.assemble() for i in teal]
+    lines += [fmt_cmt(i.assemble()) for i in teal]
```
ahangsu, line 322: nit: I am understanding correctly, I think this would enforce all the line to come with trace comment.

Do you think it is reasonable to add a keyword argument in `compileTeal` to activate/deactivate trace in generated Teal?

ahangsu (commented):
Just starting a quik pass

pyteal/compiler/compiler.py
barnjamin: yes! a debug flag would be great addition for this and other thigns as well i think

ahangsu: let's consider this: just like `optimizeOptions`, we should also consider something like `debugOptions`, does that sound plausible to you?

barnjamin: added `debug` as a simple boolean for now

pyteal/ir/tealcomponent.py
```diff
@@ -6,6 +6,24 @@
...
+        if "__init__" in tb[idx]:
+            continue
+        else:
+            break
+
+    file, line = tb[idx].split(",")[:2]  # Only take file: xxx, line xx, ...
+    file = file.replace("File ", "").replace(" ", "")  # Remove `File` and any spaces
+
+    return f"\t\t// pyteal-src {':'.join([file, line])}"
+
```
barnjamin, line 25: This one needs work to make sure we're actually catching the thing we want

pyteal/compiler/compiler.py
```diff
@@ -308,6 +311,20 @@ def compileTeal(
...
+            if idx <= 0:
+                return teal_line
+        except Exception:
+            return teal_line
+
+        line = teal_line[:idx].strip(" \t")
+        if debug:
+            return +"\t" * 2 + teal_line[idx:]
+
+        return line
```
barnjamin, line 326: yuck, ben

ahangsu:
oh no, feature/abi merged, you should probably remerge master branch.

PR #448, opened 2022-07-18
allow passing router description to Contract constructor

barnjamin:
Adding an optional arg to Router to allow a description to be passed through to Contract constructor


tzaffi (approved):
Seems like an obvious no-risk improvement that can go into the ABI-feature release.

Others ought chime in as well.

ahangsu (approved):
Looks good to me, feels like it is a good to have feature in router.

Should be careful about test and type-checking.

PR #453, opened 2022-07-20
`abi.Bool` improvements

jasonpaulos:
This PR makes two small improvements to `abi.Bool`:

1. When calling `abi.Bool.set` with an `Expr`, the program will no longer panic if the value is greater than 1. Instead, all values greater than 0 will be treated as true. This is more in line with other AVM opcodes which accept boolean values.
2. Rename `abi.Bool.decodeBit` to `abi.Bool.decode_bit`


michaeldiamant (approved):

ahangsu (approved):
lgtm

PR #461, opened 2022-07-21
Rename pyteal/ast/abi/tuple.py functions to follow PEP 8 conventions

michaeldiamant:
Renames `pyteal/ast/abi/bool.py` functions to follow PEP 8 conventions. Adds leading _ to demarcate internal functions.

If reviewers feel the internal usage does not apply, let me know and I'll adjust.


ahangsu (approved):
The leading underline seems good to me, dunno how @jasonpaulos think?

jasonpaulos (approved):
Changes look good to me, thanks

PR #464, opened 2022-07-21
Specify dependency versions

jasonpaulos:
Add version constraints for our runtime dependencies.

* For `py-algorand-sdk`, version [1.9.0](https://github.com/algorand/py-algorand-sdk/releases/tag/v1.9.0) is the first with ABI support
* For `semantic-version`, version [2.9.0](https://github.com/rbarrois/python-semanticversion/blob/master/ChangeLog#290-2022-02-06) is the first that supports Python 3.10


michaeldiamant (approved):
@jasonpaulos Seems reasonable - thanks for scrutinizing the dependencies!

PR #465, opened 2022-07-21
Add py.typed marker

jasonpaulos:
Currently we use mypy internally to type check our code, but unfortunately any packages that import PyTeal are not able to take advantage of our code's types when they use mypy.

The reason is the `pyteal` package must declare to mypy that we actually use types. More info [can be found here](https://mypy.readthedocs.io/en/stable/installed_packages.html#creating-pep-561-compatible-packages), but for our purposes the answer is to simply add a `py.typed` file to our package.

I've verified this actually works and allows dependent packages to do proper type checking with our code as well.


jasonpaulos:
I'm making this a draft until #264 is merged to avoid any further merge complications.

ahangsu (approved):

michaeldiamant (approved):
@jasonpaulos Matches how the docs recommend configuration.  Thanks for considering the use case - makes a lot of sense to add in!

PR #469, opened 2022-07-22
Merge Teal7 to AVM8, and consolidate Teal to AVM versioning

ahangsu:


ahangsu:
Seems somewhat complex, closing for now.

Issue #180, opened 2022-02-01
ABI:  Named tuple support


jasonpaulos:
## Summary

Python allows custom named tuples to be created with this syntax:

```python
from typing import NamedTuple

class ANamedTuple(NamedTuple):
    """a docstring"""
    foo: int
    bar: str
    baz: list
```

Since we have `abi.Tuple`, it is possible we can do something similar and support ABI named tuples in PyTeal.

## Scope
A named tuple is essentially a normal tuple, except it has a mapping from user-defined names to its indexes.

Users should be able to define as many named tuples as they want and use them just like they would use a normal `abi.Tuple`.

Probably we would want a base class called `abi.NamedTuple`, which user-defined named tuples would subclass, just like the Python example.

I have not yet thought about how `abi.NamedTuple` would discover the fields on it and map those to tuple indexes, but we can look at Python's `NamedTuple` implementation for inspiration.

barnjamin:
https://github.com/python/cpython/blob/main/Lib/typing.py#L2259-L2324 for reference, this is the implementation in python

barnjamin:
Initial attempt at using subclass of Tuple to provide arbitrary struct like interface https://github.com/algorand-devrel/abi-pyteal/blob/main/struct.py#L20-L41

jannotti:
I'd like to see us distinguish this from `abi`.  Since this is an "internal object", I'd rather it was, roughly, `pt.NameTuple` or even `pt.Struct`.  We should distinguish between a programming model that pyteal gives - "structs" - from how those things are serialized `abi.Tuple`, just as Go has structs, but there's no names at the ABI level (in function calling conventions).

PR #473, opened 2022-07-25
NamedTuple Implementation

ahangsu:
Take implementation and designs from `vex` and `beaker` for #180 implementation.

The implementation is still minimum, and the limit for length (5) in `Tuple` is still unrelieved, for we want to generate annotation at some point.

### DIFF
--- a/docs/abi.rst
+++ b/docs/abi.rst
@@ -210,7 +210,7 @@ PyTeal Type                                    ARC-4 Type             Dynamic / 
 :any:`abi.Address`                             :code:`address`        Static                              A 32-byte Algorand address. This is an alias for :code:`abi.StaticArray[abi.Byte, Literal[32]]`.
 :any:`abi.DynamicArray[T] <abi.DynamicArray>`  :code:`T[]`            Dynamic                             A variable-length array of :code:`T`
 :any:`abi.String`                              :code:`string`         Dynamic                             A variable-length byte array assumed to contain UTF-8 encoded content. This is an alias for :code:`abi.DynamicArray[abi.Byte]`.
-:any:`abi.Tuple`\*                             :code:`(...)`          Static when all elements are static A tuple of multiple types
+:any:`abi.Tuple`\*, :any:`abi.NamedTuple`\*    :code:`(...)`          Static when all elements are static A tuple of multiple types
 ============================================== ====================== =================================== =======================================================================================================================================================
 
 .. note::
@@ -222,6 +222,21 @@ PyTeal Type                                    ARC-4 Type             Dynamic / 
     * :any:`abi.Tuple3[T1,T2,T3] <abi.Tuple3>`: a tuple of three values, :code:`(T1,T2,T3)`
     * :any:`abi.Tuple4[T1,T2,T3,T4] <abi.Tuple4>`: a tuple of four values, :code:`(T1,T2,T3,T4)`
     * :any:`abi.Tuple5[T1,T2,T3,T4,T5] <abi.Tuple5>`: a tuple of five values, :code:`(T1,T2,T3,T4,T5)`
+
+    \*While we are still on PyTeal 3.10, we have a workaround for :any:`abi.Tuple` by :any:`abi.NamedTuple`, which allows one to define a tuple with more than 5 generic arguments, and access tuple elements by field name. For example:
+
+    .. code-block:: python
+
+        from pyteal import *
+        from typing import Literal as L
+
+        class InheritedFromNamedTuple(abi.NamedTuple):
+            acct_address: abi.Address
+            amount: abi.Uint64
+            retrivable: abi.Bool
+            desc: abi.String
+            list_of_addrs: abi.DynamicArray[abi.Address]
+            balance_list: abi.StaticArray[abi.Uint, L[10]]
 
 These ARC-4 types are not yet supported in PyTeal:
 
@@ -285,7 +300,7 @@ All basic types that represent a single value have a :code:`get()` method, which
 A brief example is below. Please consult the documentation linked above for each method to learn more about specific usage and behavior.
 
 .. code-block:: python
-    
+
     from pyteal import *
 
     @Subroutine(TealType.uint64)
@@ -306,7 +321,7 @@ The supported methods are:
 
 * :any:`abi.StaticArray.__getitem__(index: int | Expr) <abi.StaticArray.__getitem__>`, used for :any:`abi.StaticArray` and :any:`abi.Address`
 * :any:`abi.Array.__getitem__(index: int | Expr) <abi.Array.__getitem__>`, used for :any:`abi.DynamicArray` and :any:`abi.String`
-* :any:`abi.Tuple.__getitem__(index: int) <abi.Tuple.__getitem__>`
+* :any:`abi.Tuple.__getitem__(index: int) <abi.Tuple.__getitem__>`, used for :any:`abi.NamedTuple`
 
 .. note::
     Be aware that these methods return a :any:`ComputedValue`, similar to other PyTeal operations which return ABI types. More information about why that is necessary and how to use a :any:`ComputedValue` can be found in the :ref:`Computed Values` section.
@@ -759,7 +774,7 @@ Registering Methods
     **We strongly recommend** methods immediately access and validate compound type parameters *before* persisting arguments for later transactions. For validation, it is sufficient to attempt to extract each element your method will use. If there is an input error for an element, indexing into that element will fail.
 
     Notes:
-    
+
     * This recommendation applies to recursively contained compound types as well. Successfully extracting an element which is a compound type does not guarantee the extracted value is valid; you must also inspect its elements as well.
     * Because of this, :any:`abi.Address` is **not** guaranteed to have exactly 32 bytes. To defend against unintended behavior, manually verify the length is 32 bytes, i.e. :code:`Assert(Len(address.get()) == Int(32))`.
 
@@ -804,7 +819,7 @@ The first way to register a method is with the :any:`Router.add_method_handler` 
             # store the result in the sender's local state too
             App.localPut(Txn.sender(), Bytes("result", output.get())),
         )
-    
+
     # Register the `add` method with the router, using the default `MethodConfig`
     # (only no-op, non-creation calls allowed).
     router.add_method_handler(add)
--- a/docs/abi.rst
+++ b/docs/abi.rst
@@ -210,7 +210,7 @@ PyTeal Type                                    ARC-4 Type             Dynamic / 
 :any:`abi.Address`                             :code:`address`        Static                              A 32-byte Algorand address. This is an alias for :code:`abi.StaticArray[abi.Byte, Literal[32]]`.
 :any:`abi.DynamicArray[T] <abi.DynamicArray>`  :code:`T[]`            Dynamic                             A variable-length array of :code:`T`
 :any:`abi.String`                              :code:`string`         Dynamic                             A variable-length byte array assumed to contain UTF-8 encoded content. This is an alias for :code:`abi.DynamicArray[abi.Byte]`.
-:any:`abi.Tuple`\*, :any:`abi.NamedTuple`\*    :code:`(...)`          Static when all elements are static A tuple of multiple types
+:any:`abi.Tuple`\*, :any:`abi.NamedTuple`      :code:`(...)`          Static when all elements are static A tuple of multiple types
 ============================================== ====================== =================================== =======================================================================================================================================================
 
 .. note::
@@ -223,7 +223,7 @@ PyTeal Type                                    ARC-4 Type             Dynamic / 
     * :any:`abi.Tuple4[T1,T2,T3,T4] <abi.Tuple4>`: a tuple of four values, :code:`(T1,T2,T3,T4)`
     * :any:`abi.Tuple5[T1,T2,T3,T4,T5] <abi.Tuple5>`: a tuple of five values, :code:`(T1,T2,T3,T4,T5)`
 
-    \*While we are still on PyTeal 3.10, we have a workaround for :any:`abi.Tuple` by :any:`abi.NamedTuple`, which allows one to define a tuple with more than 5 generic arguments, and access tuple elements by field name. For example:
+    While we are still on PyTeal 3.10, we have a workaround for :any:`abi.Tuple` by :any:`abi.NamedTuple`, which allows one to define a tuple with more than 5 generic arguments, and access tuple elements by field name. For example:
 
     .. code-block:: python
 
@@ -236,7 +236,7 @@ PyTeal Type                                    ARC-4 Type             Dynamic / 
             retrivable: abi.Bool
             desc: abi.String
             list_of_addrs: abi.DynamicArray[abi.Address]
-            balance_list: abi.StaticArray[abi.Uint, L[10]]
+            balance_list: abi.StaticArray[abi.Uint64, L[10]]
 
 These ARC-4 types are not yet supported in PyTeal:
 
@@ -321,10 +321,17 @@ The supported methods are:
 
 * :any:`abi.StaticArray.__getitem__(index: int | Expr) <abi.StaticArray.__getitem__>`, used for :any:`abi.StaticArray` and :any:`abi.Address`
 * :any:`abi.Array.__getitem__(index: int | Expr) <abi.Array.__getitem__>`, used for :any:`abi.DynamicArray` and :any:`abi.String`
-* :any:`abi.Tuple.__getitem__(index: int) <abi.Tuple.__getitem__>`, used for :any:`abi.NamedTuple`
+* :any:`abi.Tuple.__getitem__(index: int) <abi.Tuple.__getitem__>`, used for :any:`abi.Tuple` and :any:`abi.NamedTuple`\*
 
 .. note::
-    Be aware that these methods return a :any:`ComputedValue`, similar to other PyTeal operations which return ABI types. More information about why that is necessary and how to use a :any:`ComputedValue` can be found in the :ref:`Computed Values` section.
+    \*For :any:`abi.NamedTuple`, one can access tuple elements through both methods
+
+    * :any:`abi.Tuple.__getitem__(index: int) <abi.Tuple.__getitem__>`
+    * :any:`abi.NamedTuple.__getattr__(name: str) <abi.NamedTuple.__getattr__>`
+
+    Be aware that, though the fields in a :any:`abi.NamedTuple` are annotated in ABI types, the return values from the annotated fields are :any:`ComputedValues <abi.ComputedValue>`.
+
+    Also be aware that these methods return a :any:`ComputedValue`, similar to other PyTeal operations which return ABI types. More information about why that is necessary and how to use a :any:`ComputedValue` can be found in the :ref:`Computed Values` section.
 
 A brief example is below. Please consult the documentation linked above for each method to learn more about specific usage and behavior.
 
--- a/pyteal/ast/abi/tuple.py
+++ b/pyteal/ast/abi/tuple.py
@@ -523,6 +523,12 @@ class NamedTuple(Tuple):
         if not anns:
             raise Exception("Expected fields to be declared but found none")
 
+        # NOTE: this `_ready` variable enables `__setattr__` during `__init__` execution,
+        # while after `__init__`, we cannot use `__setattr__` to set fields in `NamedTuple`.
+        # The reason for `_ready` is that, the field naming scheme in Python would change
+        # _double underscore led field name_ to something else,
+        # while _single underscore led variable names_ would stay intact.
+        self._ready = False
         self.__type_specs: OrderedDict[str, TypeSpec] = OrderedDict()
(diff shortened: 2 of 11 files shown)


tests/integration/abi_roundtrip_test.py
```diff
@@ -33,6 +34,15 @@
     "appl": GAI_ISSUE_2068,
 }
 
+
+class NamedTupleInherit(abi.NamedTuple):
+    a: abi.Bool
+    b: abi.Address
+    c: abi.Tuple2[abi.Uint64, abi.Bool]
+    d: abi.StaticArray[abi.Byte, Literal[10]]
+    e: abi.StaticArray[abi.Bool, Literal[4]]
```
barnjamin, line 43: do 6?

ahangsu: nah they won't let me, somewhere around it is restricted to generate `Tuple` type annotation, and we can go as far as 5.

I need to think a little more on how to get around.

ahangsu: seems to break beyond the boundary of 5 right now, wanna see how it goes?

tests/integration/abi_roundtrip_test.py
```diff
@@ -142,12 +154,21 @@ def roundtrip_setup(abi_type):
...
+            str(abi.NamedTuple)
+            if isclass(at) and issubclass(at, abi.NamedTuple)
+            else str(at)
         )
-    }
+        tli = tli.split("[")[0] if tli.startswith("pyteal") else tli.split("'")[1]
+
+        top_level_names.add(tli)
+
+    print(top_level_names)
```
tzaffi, line 171:
```suggestion
    print(f"{top_level_names=}")
```
That suggestion assumes you want to keep the printout. But you might also consider removing it altogether.

tzaffi (commented):
Looks like a very worthwhile addition. My main suggestion is to test that the expected functionality of `NamedTuple` - namely using the names instead of numeric indexes, will work as expected.

pyteal/ast/abi/tuple.py
```diff
@@ -484,3 +485,49 @@ def __init__(
 
 
 Tuple5.__module__ = "pyteal.abi"
+
+
+class NamedTupleTypeSpec(TupleTypeSpec):
+    """A NamedTupleType inherited from TupleTypeSpec, allowing for more than 5 elements."""
```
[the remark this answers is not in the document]
ahangsu: 👍 

uh, my english, smh

pyteal/ast/abi/tuple.py
```diff
@@ -484,3 +485,49 @@ def __init__(
...
+
+    def annotation_type(self) -> "type[NamedTuple]":
+        return self.instance_class
+
+
+NamedTupleTypeSpec.__module__ = "pyteal.abi"
+
+
+class NamedTuple(Tuple):
+    """A NamedTuple from Tuple with all it's elements named."""
```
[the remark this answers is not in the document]
ahangsu: 👍 

good catch

pyteal/ast/abi/tuple.py
```diff
@@ -484,3 +485,49 @@ def __init__(
...
+
+    def __init__(self):
+        if type(self) is NamedTuple:
+            raise TealInputError("NamedTuple must be subclassed.")
+
+        if not hasattr(self, "__annotations__"):
```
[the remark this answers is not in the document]
ahangsu: that is way better than calling `__annotations__` directly from attrs 👍

tests/integration/abi_roundtrip_test.py
ahangsu: oh boy, I forgot to delete this line in debugging. No worries on changing the test behavior.

tests/integration/abi_roundtrip_test.py
```diff
@@ -33,6 +34,16 @@
     "appl": GAI_ISSUE_2068,
 }
 
+
+class NamedTupleInherit(abi.NamedTuple):
```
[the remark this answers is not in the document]
tzaffi: Thanks for clarifying. Yeah, any test that proves its "named tupleness" in all its glory would be great.

tests/integration/graviton_test.py
```diff
@@ -908,6 +908,65 @@ def while_continue_accumulation(n):
...
+        f: abi.Uint64
+
+    sdk_abi_t = abi.algosdk_from_annotation(NamedTupleExample)
+
+    @Blackbox(input_types=[None])
+    @Subroutine(TealType.uint64)
+    def named_tuple_field_access(v_temp: NamedTupleExample):
+        return Seq(
+            (v_a0 := abi.Bool()).set(v_temp[0]),
+            (v_a1 := abi.Bool()).set(v_temp.a),
```
barnjamin, line 939: 👨‍🍳

tests/integration/graviton_test.py
```diff
@@ -908,6 +908,65 @@ def while_continue_accumulation(n):
...
+                And(
+                    v_a0.get() == v_a1.get(),
+                    v_b0.get() == v_b1.get(),
+                    v_c0.encode() == v_c1.encode(),
+                    v_d0.encode() == v_d1.encode(),
+                    v_e0.encode() == v_e1.encode(),
+                    v_f0.get() == v_f1.get(),
+                )
+            ),
+        )
```
barnjamin, line 960: Could you add one where we set the values on the NamedTuple?

ahangsu: yes, it now has both setting and getting (under name not index).

pyteal/ast/abi/tuple.py
```diff
@@ -484,3 +486,57 @@ def __init__(
...
+
+        for index, (name, annotation) in enumerate(anns.items()):
+            self.type_specs[name] = type_spec_from_annotation(annotation)
+
+        super().__init__(
+            NamedTupleTypeSpec(type(self), *list(self.type_specs.values()))
+        )
+
+        for index, name in enumerate(anns):
+            setattr(self, name, self[index])
```
jasonpaulos, line 536: One potential issue with this implementation is that it doesn't prevent users from assigning new values to these fields. For example:

```python
instance = NamedTupleInherit()
instance.c = True

# now instance.c returns True instead of indexing into the instance value
program = If(instance.c).Then(...)
```

If this is something we care to fix, one solution would be to override `__getattr__` and `__setattr__` for this class. I attempted to do this locally and I think you can have success with this approach:

```diff
diff --git a/pyteal/ast/abi/tuple.py b/pyteal/ast/abi/tuple.py
index 5da00d9..7f7f8df 100644
--- a/pyteal/ast/abi/tuple.py
+++ b/pyteal/ast/abi/tuple.py
@@ -523,17 +523,29 @@ class NamedTuple(Tuple):
         if not anns:
             raise Exception("Expected fields to be declared but found none")
 
+        self._ready = False
+
         self.type_specs: OrderedDict[str, TypeSpec] = OrderedDict()
+        self.indexes: dict[str, int] = dict()
 
         for index, (name, annotation) in enumerate(anns.items()):
             self.type_specs[name] = type_spec_from_annotation(annotation)
+            self.indexes[name] = index
 
         super().__init__(
             NamedTupleTypeSpec(type(self), *list(self.type_specs.values()))
         )
 
-        for index, name in enumerate(anns):
-            setattr(self, name, self[index])
+        self._ready = True
+    
+    def __getattr__(self, name: str) -> TupleElement[Any]:
+        return self.__getitem__(self.indexes[name])
+    
+    def __setattr__(self, name: str, value: Any) -> None:
+        if name == "_ready" or not self._ready:
+            super().__setattr__(name, value)
+            return
+        raise TealInputError("Cannot assign to NamedTuple attributes")
 
     def type_spec(self) -> TupleTypeSpec:
         return cast(NamedTupleTypeSpec, super().type_spec())

```

pyteal/ast/abi/tuple.py
```diff
@@ -484,3 +486,57 @@ def __init__(
...
+
+    def type_spec(self) -> TupleTypeSpec:
+        return cast(NamedTupleTypeSpec, super().type_spec())
+
+
+NamedTuple.__module__ = "pyteal.abi"
```
jasonpaulos, line 542: I'd like to see a unit test in `tuple_test.py` that confirms the return value of each user-defined field is as expected, something like the existing `test_Tuple_getitem` test

jasonpaulos (commented):
There is still one problem I can't imagine a solution for. If I have the class:

```python
class Point(abi.NamedTuple):
    x: abi.Uint64
    y: abi.Uint64
```

mypy will think that `Point().x` and `Point().y` are `abi.Uint64` instances, when in reality they are `abi.TupleElement[abi.Uint64]` instances. I am aware of no ways to remedy this.

pyteal/ast/abi/tuple.py
```diff
@@ -484,3 +486,57 @@ def __init__(
...
+
+    def __init__(self):
+        if type(self) is NamedTuple:
+            raise TealInputError("NamedTuple must be subclassed.")
+
+        anns = get_annotations(type(self))
+        if not anns:
+            raise Exception("Expected fields to be declared but found none")
+
+        self.type_specs: OrderedDict[str, TypeSpec] = OrderedDict()
```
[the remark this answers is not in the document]
ahangsu: good point.

pyteal/ast/abi/tuple.py
```diff
@@ -484,3 +486,57 @@ def __init__(
...
+            self.type_specs[name] = type_spec_from_annotation(annotation)
+
+        super().__init__(
+            NamedTupleTypeSpec(type(self), *list(self.type_specs.values()))
+        )
+
+        for index, name in enumerate(anns):
+            setattr(self, name, self[index])
+
+    def type_spec(self) -> TupleTypeSpec:
```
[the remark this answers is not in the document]
ahangsu: good point, that is indeed not necessary.

pyteal/ast/abi/tuple.py
ahangsu: good point, but I would like to know what is the `self._ready` for?

pyteal/ast/abi/tuple.py
ahangsu: good point, just add a similar test based on NamedTuple

pyteal/ast/abi/tuple.py
ahangsu: oh snap I kind of see it coming. I suppose it is for avoiding erroring in `NamedTuple` level assigning, but still error on attribute field level?

jasonpaulos: The `__setattr__` method is invoked when the attributes `type_specs`, `indexes`, and all attributes from `super().__init__()` are set. In order to allow these attributes to be set like normal, I introduced the `_ready` variable, which essentially disables the `__setattr__` method until after the `__init__` method has finished.

There may be others ways to achieve the same thing, but this is what I came up with.

ahangsu: really good point, I think it involves a few things here in Python class initialization, and class hidden member naming convention, for I tried `__ready` and it does not quite work.

Essentially we are disabling `__setattr__` after `__init__` is done for executing. I will add some notes to explain what is going on. Thanks for explaining.

pyteal/ast/abi/tuple_test.py
```diff
@@ -824,3 +824,67 @@ def test_TupleElement_store_into():
...
+    tuple_value = test_case()
+    tuple_len_static = tuple_value.type_spec().length_static()
+    for i in range(tuple_len_static):
+        elem_by_field: abi.TupleElement = getattr(tuple_value, f"f{i}")
+        assert (
+            type(elem_by_field) is abi.TupleElement
+        ), f"Test case {test_case} at field f{i} must be TupleElement"
+        elem_by_index: abi.TupleElement = tuple_value[i]
+        assert (
+            type(elem_by_field) is abi.TupleElement
```
jasonpaulos, line 874: Copy paste error?
```suggestion
            type(elem_by_index) is abi.TupleElement
```

pyteal/ast/abi/tuple_test.py
```diff
@@ -824,3 +824,67 @@ def test_TupleElement_store_into():
...
+        ), f"Test case {test_case} at index {i} must be TupleElement"
+        assert (
+            elem_by_field.index == i
+        ), f"Test case {test_case} at field f{i} should have index {i}."
+        assert (
+            elem_by_field.tuple == tuple_value
```
jasonpaulos, line 880: nit: it would be nice to use the `is` operator instead of `==` because we want the `tuple` property to be reference-equal to the original object

pyteal/ast/abi/tuple_test.py
```diff
@@ -824,3 +824,67 @@ def test_TupleElement_store_into():
...
+        elem_by_field: abi.TupleElement = getattr(tuple_value, f"f{i}")
+        assert (
+            type(elem_by_field) is abi.TupleElement
+        ), f"Test case {test_case} at field f{i} must be TupleElement"
+        elem_by_index: abi.TupleElement = tuple_value[i]
+        assert (
+            type(elem_by_field) is abi.TupleElement
+        ), f"Test case {test_case} at index {i} must be TupleElement"
+        assert (
+            elem_by_field.index == i
```
jasonpaulos, line 877: I'd like to see the `index` and `tuple` properties of `elem_by_index` tested as well, to ensure the `__getitem__` function behaves exactly the same on `NamedTuple`

pyteal/ast/abi/tuple.py
```diff
@@ -523,20 +523,37 @@ def __init__(self):
         if not anns:
             raise Exception("Expected fields to be declared but found none")
 
-        self.type_specs: OrderedDict[str, TypeSpec] = OrderedDict()
+        # NOTE: this `_ready` variable enables `__setattr__` during `__init__` execution,
+        # while after `__init__`, we cannot use `__setattr__` to set fields in `NamedTuple`.
+        # The reason for `_ready` is that, the field naming scheme in Python would change
+        # _double underscore led field name_ to something else,
+        # while _single underscore led variable names_ would stay intact.
+        self._ready = False
```
jasonpaulos, line 531: nit: `__ready` would be more consistent with the other field names

docs/abi.rst
```diff
@@ -306,7 +321,7 @@ The supported methods are:
 
 * :any:`abi.StaticArray.__getitem__(index: int | Expr) <abi.StaticArray.__getitem__>`, used for :any:`abi.StaticArray` and :any:`abi.Address`
 * :any:`abi.Array.__getitem__(index: int | Expr) <abi.Array.__getitem__>`, used for :any:`abi.DynamicArray` and :any:`abi.String`
-* :any:`abi.Tuple.__getitem__(index: int) <abi.Tuple.__getitem__>`
+* :any:`abi.Tuple.__getitem__(index: int) <abi.Tuple.__getitem__>`, used for :any:`abi.NamedTuple`
```
jasonpaulos, line 324: Minor change to clarify that this also describes `Tuple`
```suggestion
* :any:`abi.Tuple.__getitem__(index: int) <abi.Tuple.__getitem__>`, used for :any:`abi.Tuple` and :any:`abi.NamedTuple`
```

docs/abi.rst
```diff
@@ -210,7 +210,7 @@ PyTeal Type                                    ARC-4 Type             Dynamic /
 :any:`abi.Address`                             :code:`address`        Static                              A 32-byte Algorand address. This is an alias for :code:`abi.StaticArray[abi.Byte, Literal[32]]`.
 :any:`abi.DynamicArray[T] <abi.DynamicArray>`  :code:`T[]`            Dynamic                             A variable-length array of :code:`T`
 :any:`abi.String`                              :code:`string`         Dynamic                             A variable-length byte array assumed to contain UTF-8 encoded content. This is an alias for :code:`abi.DynamicArray[abi.Byte]`.
-:any:`abi.Tuple`\*                             :code:`(...)`          Static when all elements are static A tuple of multiple types
+:any:`abi.Tuple`\*, :any:`abi.NamedTuple`\*    :code:`(...)`          Static when all elements are static A tuple of multiple types
```
jasonpaulos, line 213: I'd like to suggest removing the asterisk on NamedTuple on this line and at the beginning of "While we are still on PyTeal 3.10..." below. Having two of the same asterisk symbols doesn't make sense to me.
```suggestion
:any:`abi.Tuple`\*, :any:`abi.NamedTuple`      :code:`(...)`          Static when all elements are static A tuple of multiple types
```

docs/abi.rst
```diff
@@ -306,7 +321,7 @@ The supported methods are:
 
 * :any:`abi.StaticArray.__getitem__(index: int | Expr) <abi.StaticArray.__getitem__>`, used for :any:`abi.StaticArray` and :any:`abi.Address`
 * :any:`abi.Array.__getitem__(index: int | Expr) <abi.Array.__getitem__>`, used for :any:`abi.DynamicArray` and :any:`abi.String`
-* :any:`abi.Tuple.__getitem__(index: int) <abi.Tuple.__getitem__>`
+* :any:`abi.Tuple.__getitem__(index: int) <abi.Tuple.__getitem__>`, used for :any:`abi.NamedTuple`
 
 .. note::
```
jasonpaulos, line 326: It would be nice to see a second note section here which explains that `NamedTuple` fields can be accessed by field name in addition to using `__getitem__`

docs/abi.rst
```diff
@@ -223,6 +223,21 @@ PyTeal Type                                    ARC-4 Type             Dynamic /
...
+        from pyteal import *
+        from typing import Literal as L
+
+        class InheritedFromNamedTuple(abi.NamedTuple):
+            acct_address: abi.Address
+            amount: abi.Uint64
+            retrivable: abi.Bool
+            desc: abi.String
+            list_of_addrs: abi.DynamicArray[abi.Address]
+            balance_list: abi.StaticArray[abi.Uint, L[10]]
```
jasonpaulos, line 239: This just says `Uint`, not a specific uint type

jasonpaulos (commented):
Thanks for the tests and user guide updates. I left a bunch of small comments

docs/abi.rst
ahangsu: I somewhat re-ordered the following note, let me know how this should be rephrased.

pyteal/ast/abi/tuple.py
ahangsu: tho `__ready` is consistent with other field's name, it unfortunately cannot fit here, for python would swap the naming of attributes, see: https://stackoverflow.com/questions/1301346/what-is-the-meaning-of-single-and-double-underscore-before-an-object-name, so we need to deliberately change the `__getattr__` implementation

jasonpaulos (commented):
In an attempt to solve the mypy problem [I pointed out earlier](https://github.com/algorand/pyteal/pull/473#pullrequestreview-1053310113), I've opened https://github.com/algorand/pyteal/pull/487 as a child PR.

Please let me know what you think in that PR.

pyteal/ast/abi/tuple.py
ahangsu: > the __getattr__ method would need to check if name == "_NamedTuple__ready"

yes exactly, then it is non trivial to see why `__ready` gets changed into `_NamedTuple__ready`, for the change is implicit.

docs/abi.rst
```diff
@@ -223,6 +223,21 @@ PyTeal Type                                    ARC-4 Type             Dynamic /
...
+        from pyteal import *
+        from typing import Literal as L
+
+        class InheritedFromNamedTuple(abi.NamedTuple):
+            acct_address: abi.Address
+            amount: abi.Uint64
+            retrivable: abi.Bool
+            desc: abi.String
+            list_of_addrs: abi.DynamicArray[abi.Address]
+            balance_list: abi.StaticArray[abi.Uint64, L[10]]
```
jasonpaulos, line 239: I forgot to update part
```suggestion
            acct_address: abi.Field[abi.Address]
            amount: abi.Field[abi.Uint64]
            retrivable: abi.Field[abi.Bool]
            desc: abi.Field[abi.String]
            list_of_addrs: abi.Field[abi.DynamicArray[abi.Address]]
            balance_list: abi.Field[abi.StaticArray[abi.Uint64, L[10]]]
```

ahangsu: yes here you go

pyteal/ast/abi/tuple.py
jasonpaulos: Ok great, this is the last thing blocking an approval for me

ahangsu: I added a few comments above declaration of `self._ready`, hopefully that makes sense to you.

jasonpaulos: Sorry if I was unclear, my preference is to rename `_ready` to `__ready`.

The reason being we should minimize the number of field names that cannot be used by `NamedTuple` subclasses.

ahangsu: ugh ok. I will swap the naming then.

barnjamin (approved):

pyteal/ast/abi/tuple.py
ahangsu: alright, final pass for a naming change...

PR #476, opened 2022-07-27
Can subroutine haz abi (in graviton tests)

tzaffi:
When reviewing #473 I relaized that the `PyTealDryRunExecutor` hadn't been updated to reflect the fact that `@Subroutines` may happily accept ABI-types. Consequently, there were all sorts of contortions necessary to get graviton tests to run on the `NamedTuple` example. 

This change should simplify usage of graviton tests in the case of regular `@Subroutines` with ABI-types.

### DIFF
--- a/tests/integration/graviton_test.py
+++ b/tests/integration/graviton_test.py
@@ -908,7 +908,7 @@ def blackbox_pyteal_while_continue_test():
         )
 
 
-def blackbox_pyteal_named_tupleness_test():
+def blackbox_pyteal_named_tupleness():
     from typing import Literal as L
     from tests.blackbox import Blackbox
     from pyteal import (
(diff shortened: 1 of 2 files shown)


tests/integration/graviton_test.py
```diff
@@ -961,22 +960,28 @@ def named_tuple_field_access(
...
+    #     ("bool[4]", [False] * 4),
+    #     ("uint64", 100),
+    # ]
+
+    # args: list[bytes] = [
+    #     ABIType.from_string(abi_t).encode(var) for abi_t, var in TYPES_N_VALS
+    # ]
+
+    # lsig_result = lsig_pytealer.dryrun(args)
+    # assert lsig_result.passed()
```
ahangsu, line 984: if graviton can do all the stuffs from python var to ABI encodings, that is great! I don't see any point we should keep my previous snippet of implementation here tho...

tzaffi: ok, I'll update without

tzaffi: updated

ahangsu: 👍

ahangsu (approved):

Issue #327, opened 2022-05-06
CI:  Fail readthedocs build on warning


michaeldiamant:
## Problem

Analogous to https://github.com/algorand/py-algorand-sdk/issues/335.

## Solution

Analogous to https://github.com/algorand/py-algorand-sdk/issues/335 with the following distinction:  py-algorand-sdk uses CircleCI rather than Github Workflows for its builds.

## Dependencies

Analogous to https://github.com/algorand/py-algorand-sdk/issues/335.

## Urgency

Analogous to https://github.com/algorand/py-algorand-sdk/issues/335.

PR #478, opened 2022-07-28
Sphinx Error on Warnings

ahangsu:
## Description

This PR closes https://github.com/algorand/pyteal/issues/327 with the suggestions from the following links:

- https://stackoverflow.com/questions/38048945/how-to-turn-warnings-into-errors-when-building-sphinx-documentation-with-setupto
- https://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-W


Makefile
```diff
@@ -24,10 +24,13 @@ bundle-docs-clean:
 
 bundle-docs: bundle-docs-clean
 	cd docs && \
-	make html && \
+	make html SPHINXOPTS="-W --keep-going" && \
```
michaeldiamant, line 27: @ahangsu Thanks for the earlier test to prove the highlighted line _can_ cause the build to fail (https://github.com/algorand/pyteal/runs/7568763886?check_suite_focus=true).

I'm wondering your thoughts on including `-n` to check for broken links?

* I see at least one commit where you tried it (https://github.com/algorand/pyteal/commit/35ab49b0e5cb53103bf9e9d22980ee4935cdf80f).  Though it's unclear why it was reverted.
* I've not tried using the option though it _seems_ like a prudent check.  If you haven't observed issues with `-n`, I'm up for trying it.

Makefile
```diff
@@ -63,7 +66,7 @@ sandbox-dev-up:
 
 sandbox-dev-stop:
```
michaeldiamant, line 67: @ahangsu Unrelated to the line:  Gauging PR scope - Do you plan to apply a similar configuration change to readthedocs in-scope to this PR or a different PR?

* Specifically, I'm referring to https://docs.readthedocs.io/en/stable/config-file/v2.html#sphinx-fail-on-warning in `.readthedocs.yaml`.
* Either works for me provided we make the change prior to closing the story.

Makefile
```diff
@@ -24,10 +24,13 @@ bundle-docs-clean:
...
 bundle-docs: bundle-docs-clean
 	cd docs && \
-	make html && \
+	make html SPHINXOPTS="-W --keep-going" && \
 	doc2dash --name pyteal --index-page index.html --online-redirect-url https://pyteal.readthedocs.io/en/ _build/html && \
 	tar -czvf pyteal.docset.tar.gz pyteal.docset
 
+# The `SPHINXOPTS="-W -keep-going"` comes from https://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-W
+# also check link https://stackoverflow.com/questions/38048945/how-to-turn-warnings-into-errors-when-building-sphinx-documentation-with-setupto
+
```
michaeldiamant, line 33: @ahangsu Optional - Feel welcomed to resolve as is:  Since we don't annotate other command invocations, I don't see a _need_ to do so here.  I think `SPHINXOPTS` provides a keyword for folks to google as needed.

I raise the comment because the `Makefile` is liable to grow quite large in size.  I do _not_ feel strongly and welcome you to resolve as is if you prefer.

Makefile
michaeldiamant: Alright - I found a related build with `-n` enabled:  https://github.com/algorand/pyteal/runs/7568763886?check_suite_focus=true.

Seems it generates a number of warnings.  I briefly reviewed and started googling though I don't know off-hand how to resolve.

Without further inspection, it's unclear to me how these types of warnings manifest themselves in the docs.  I'm trying to gauge if there's specific warnings here that make `-n` worth retaining.

It's also possible we agree that adding `-n` is an exercise for another day.  Since we're here, I'd appreciate a brief look on your side to better understand the situation.

Makefile
ahangsu: I can make the change for `.readthedocs.yml` in this PR, but this made me somewhat worry about double config, tell me if I am paranoid: we have a `SPHINX_OPTION` in `Makefile`, and we (will) also have a `fail_on_warning: true` in `.readthedocs.yaml`.

If that is double config, I will remove the change in `Makefile` and only change in `.readthedocs.yaml`.

ahangsu: Holdup, what I just said about double config might be wrong, here's my hunch:
- `bundle-docs` is just used for doc-test, right?
- then I felt necessary to also have `fail_on_warning: true` activated in `.readthedocs.yaml`, for that is on publishing, rather than testing purpose.

michaeldiamant: @ahangsu Right - readthedocs build (https://readthedocs.org/projects/pyteal/builds/) happens out-of-band to our CI.  

So - We either want to:
* Make doc-test mirror the functionality of readthedocs build.
* Make readthedocs build part of our CI process.

Ideally, I _think_ we'd include readthedocs build as a status check to avoid duplicating build steps.  However, I haven't investigated it and I suspect it's more involved.  

For now, if we can make doc-test mirror readthedocs build, I think it's a worthwhile improvement.

Makefile
ahangsu: I am collecting the `nitpicky-mode` activated by option `-n`, but not too much information is collected, and I am not too sure of if it is out of scope of the ticket, but I am happy to keep exploring.

AFAIK, `nitpicky-mode`: warn about all references where the target cannot be found, which seems to be a good practice. But this would enforce us to rewrite a bunch of docstrings, e.g.,:
- we sometimes use `(optional)` for args, while this might cause warning in Sphinx nitpicky mode.

I somewhat think adding `-n` is an exercise left to be done some other day, for it would expand the scope of work at this point...

Makefile
ahangsu: Thanks for explaining! I would more prefer including readthedocs build inside CI process. No wonder why readthedocs has only publish builds, and that confuses me yesterday.

For now, I am happy to let CI test mirror what is happening in documentation publishing, but we might need to open a future improvement ticket for readthedocs in CI test.

ahangsu: if we would like to include readthedocs build inside CI, I think we will need to change github workflow somehow, and mimic https://github.com/algorand/py-algorand-sdk/blob/develop/.circleci/config.yml, where py-algorand-sdk uploads built docs to readthedocs under latest tag.

There might be some other factors I overlooked, let me know if that is the path we want to go.

Makefile
ahangsu: I added a few fixes and configs for nitpicky mode, and resolved most of the warnings in this PR. Take a pass if you are in free loop.

Makefile
jasonpaulos: Just chiming in, the `docset` Circle CI step for the Python SDK does the same thing as PyTeal's [`build-docset`](https://github.com/algorand/pyteal/blob/d8b65b7231a45d33e951b5d867864c868e0109cb/.github/workflows/build.yml#L71) step. It's not responsible for publishing the docs to readthedocs.

michaeldiamant: Thread status:  We agree _not_ integrating readthedocs build is the path of less resistance.

I'm leaving the thread open until an attempt is made to reproduce issue _fixed by_ https://github.com/algorand/pyteal/pull/472.

.github/workflows/build.yml
```diff
@@ -82,7 +82,10 @@ jobs:
         with:
           python-version: "${{ matrix.python }}"
       - name: Install python dependencies
-        run: make setup-docs
+        run: |
+          pip install -r requirements.txt
```
jasonpaulos, line 86: Is this line needed?

I ask because I believe readthedocs will not install from our top-level `requirements.txt`, so unless this is needed for some other purpose, it would be more accurate to omit it.

Also, since we are all in on Makefiles now, it might be nice to keep this behavior under `make setup-docs`.

pyteal/ast/__init__.py
```diff
@@ -20,7 +20,7 @@
     Txn,
 )
 from pyteal.ast.gtxn import GtxnExpr, GtxnaExpr, TxnGroup, Gtxn
-from pyteal.ast.block import Block
+from pyteal.ast.block import Block, BlockField
```
jasonpaulos, line 23: In general I have concerns about exporting classes that were originally not exported.

What is the justification here?

docs/conf.py
```diff
@@ -42,6 +42,27 @@
...
+    ("py.class", r".*\.T_co$"),
+    ("py.obj", r".*\.T_co$"),
+    ("py.class", r".*\.T$"),
+    ("py.class", r"^T$"),
+    ("py.obj", r".*\.N$"),
+    ("py.class", r".*\.N$"),
+    ("py.class", r"^N$"),
+    ("py.class", r"^type\[.*"),
+    ("py.obj", r"typing\..*"),
+]
```
jasonpaulos, line 65: Could you explain how this was generated and what it does?

.github/workflows/build.yml
ahangsu: yea, good catch, this line of pip install is not necessary, so we can take that away.

I thought something funky disallowed me to write setup-docs in makefile, but it turned out that some tab-space issue in makefile. I will get it fixed.

pyteal/ast/__init__.py
ahangsu: I was unaware of that `BlockField` is not suggested to be exported, but I am not sure how to suppress warning in sphinx on cannot find reference for `BlockField`. Probably need to say something about excluding `BlockField` in searching?

docs/conf.py
ahangsu: These are sphinx trying to index from `abi` type arguments, and the erroring comes from `-n` nitpicky-mode.

This config asks to ignore the error coming from `py:class` or `py:obj` cannot find reference for something matching the regex.

Makefile
michaeldiamant: @ahangsu Thanks for the experimentation - The build error in https://github.com/algorand/pyteal/pull/485 proves we'll catch a similar class of error in the future.  Resolving.

docs/conf.py
```diff
@@ -42,6 +42,27 @@
...
+    ("py.obj", r".*\.T[0-9]+$"),
+    ("py.class", r".*\.T_co$"),
+    ("py.obj", r".*\.T_co$"),
+    ("py.class", r".*\.T$"),
+    ("py.class", r"^T$"),
+    ("py.obj", r".*\.N$"),
+    ("py.class", r".*\.N$"),
+    ("py.class", r"^N$"),
+    ("py.class", r"^type\[.*"),
+    ("py.obj", r"typing\..*"),
```
michaeldiamant, line 64: @ahangsu I opened an optional child PR that removes 2 ignores by upgrading Sphinx:  https://github.com/algorand/pyteal/pull/486.

Can you have a look and see if you feel it's worth pulling in?

I do not feel strongly and will close it if you prefer as is.

ahangsu: just approved

michaeldiamant: @ahangsu Thanks for taking in the change - merged.

pyteal/ast/__init__.py
jasonpaulos: I'm a loss for how to solve this without nitpick ignore. If we could somehow disable the nitpick check for our API page, that would be a solution, but I couldn't find support for that type of customization.

If increasing the nitpick ignore list is undesirable, perhaps we should not enable the nitpick option at this time (and defer to a later issue/PR)?

ahangsu: same here, I don't see a straightforward solution for suppressing nitpick warning on specific classes, by adding some tag in docstring.

I am somewhat concerned that the nitpick mode is bringing in more complexity than desired correctness, we should probably open a ticket as a reminder about potential space for using nitpick mode in sphinx...

docs/conf.py
ahangsu: after the discussion we decided to close this thread, for we are removing the nitignore configs in `conf.py`.

pyteal/ast/__init__.py
ahangsu: after the discussion we decided to close this thread, for we are keeping `__init__.py` unchanged.

Makefile
ahangsu: after the discussion we decided to close this thread, for we will revisit `-n` nitpick mode in the future. At this moment, we still cannot specify a class un-exported to be not documented or referenced in Sphinx.

jasonpaulos (approved):

michaeldiamant (approved):
@ahangsu Thanks for working through a few iterations here - the changes immediately add correctness confidence!  

Friendly reminder to add to _Unreleased_ CHANGELOG section prior to merge.

PR #487, opened 2022-08-02
Add `NamedTuple` field annotation

jasonpaulos:
In this PR I've attempted to solve the problem of mypy thinking `NamedTuple` field values are not instances of `ComputedValue`.

I did this by making a new type `Field` which is actually just an alias of `TupleElement`. Additionally, I made `NamedTuple` require that every annotation use `Field`.

This is my attempt at a compromise between brevity and readability vs type correctness.


ahangsu (approved):
lgtm

jasonpaulos:
So that there is less work for other reviewers, I'll merge this into the parent PR, but I'm still open to taking feedback about it there

PR #492, opened 2022-08-03
Rework internals of `Comment`

jasonpaulos:
Child PR to #410 which reworks `Comment` internals.

Now, `Comment` is a function which returns a `Seq`. That `Seq` will contain one `CommentExpr` class instance for each line of the comment.

@barnjamin let me know what you think.


barnjamin:
nice! thank you!

Issue #498, opened 2022-08-09
Adding abi.Bytes alias


fabrice102:
## Problem

Currently, for dynamic byte arrays `byte[]` we need to write `abi.DynamicArray(abi.Byte)`, which is quite long.
Same for static array.

## Solution

Add some type aliases.
Propositions:
* `abi.Bytes` and `abi.Bytes[32]` for dynamic and 32-byte static array.
* `abi.DynamicByteArray` and `abi.StaticByteArray[32]`

## Dependencies

n/a

## Urgency

nice to have

ahangsu:
I believe the `abi.DynamicByteArray` can be solved by using `abi.String`.

For fixed length byte array, if you want 32 byte array, we can try `abi.Address`.

If we want other fixed length byte array, we can try to create a `abi.ByteArray` for fixed length byte array.

jasonpaulos:
> I believe the `abi.DynamicByteArray` can be solved by using `abi.String`.
> 
> For fixed length byte array, if you want 32 byte array, we can try `abi.Address`.

This would technically work, but the problem is that `abi.String` and `abi.Address` have type-specific meanings. An `abi.Address` would not be appropriate for storing a non-address 32-byte array, such as a hash output or random seed.

Same thing for `abi.String` -- if you use this type, observers of the blockchain will assume that its value is valid UTF-8 and can be shown as a user-friendly string. `goal app method` will attempt to show a string if your method returns this type. So I would not recommend storing arbitrary bytes in it.

I think PR #500 takes a good approach to addressing this issue.

PR #500, opened 2022-08-10
Static/Dynamic-Bytes Convenience

### LINKED ISSUE #495: Better `set` for `abi.DynamicArray[abi.Byte]`
## Problem

I'd like to put a byte string as the plain TealType.byte in a DynamicArray with `set` (or some other new method?)

Currently the easiest way I've found is either adding the length manually and calling decode or first creating an abi.String and calling `set` then `byte_array.decode(s.encode())`

ahangsu:
Per #498 issue, we created a pair of helper classes `StaticBytes` and `DynamicBytes` in `abi`, and hopefully the implementation resolves #495 too.

~~Some TODOs down the road: make `set` method easier to use by accepting `bytes`/`bytearray` and `pyteal.bytes`.~~

### DIFF
--- a/pyteal/ast/abi/array_dynamic.py
+++ b/pyteal/ast/abi/array_dynamic.py
@@ -1,16 +1,11 @@
-from typing import (
-    Union,
-    Sequence,
-    TypeVar,
-    cast,
-)
-
+from typing import Union, Sequence, TypeVar, cast
 
 from pyteal.errors import TealInputError
 from pyteal.ast.expr import Expr
 from pyteal.ast.seq import Seq
 
 from pyteal.ast.abi.type import ComputedValue, BaseType
+from pyteal.ast.bytes import Bytes
 from pyteal.ast.abi.uint import Uint16, Byte, ByteTypeSpec
 from pyteal.ast.abi.array_base import ArrayTypeSpec, Array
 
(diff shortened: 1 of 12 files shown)


pyteal/ast/abi/array_dynamic.py
```diff
@@ -102,3 +97,35 @@ def length(self) -> Expr:
...
+    def set(
+        self,
+        values: Union[
+            bytes,
+            bytearray,
+            Expr,
+            Sequence[Byte],
+            DynamicArray[Byte],
+            ComputedValue[DynamicArray[Byte]],
+        ],
```
ahangsu, line 117: I deliberately do not allow input to be `str`, for there might be encoding issue ('utf-8'?) here, and the interface of `set` would be not uniform. If one would like to pass in knowledge contained in `str`, one has to convert to `bytes`.

jasonpaulos (commented):
The implementations look good, but could you add unit tests for the new cases in `DynamicBytes.set` and `StaticBytes.set`?

ahangsu:
yes, I need to add `set` testcases for both classes, and I need to update `abi.rst` documentation somewhat.

pyteal/ast/abi/array_dynamic.py
```diff
@@ -102,3 +97,35 @@ def length(self) -> Expr:
 
 
 DynamicArray.__module__ = "pyteal.abi"
+
+
+class DynamicBytes(DynamicArray[Byte]):
+    """The convenience class that represents ABI static byte array."""
```
barnjamin, line 103:
```suggestion
    """The convenience class that represents ABI dynamic byte array."""
```

barnjamin:
Looks awesome! Love the helpers for assigning setting!

jasonpaulos (commented):
While reviewing I noticed that the existing `_encode_string` function has an issue, so I made a bugfix PR here: https://github.com/algorand/pyteal/pull/508

Ideally that PR would be merged before this one.

pyteal/ast/abi/array_dynamic_test.py
```diff
@@ -203,6 +203,78 @@ def test_DynamicArray_set_computed():
...
+    assert expr.type_of() == pt.TealType.none
+    assert not expr.has_return()
+
+    actual, _ = expr.__teal__(options)
+    actual.addIncoming()
+    actual = actual.NormalizeBlocks(actual)
+
+    expected = pt.TealSimpleBlock(
+        [
+            pt.TealOp(None, pt.Op.byte, "0x" + pt.Bytes(test_case).byte_str),
```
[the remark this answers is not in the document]
ahangsu: that would work, I will change accordingly

jasonpaulos (commented):
Implementation and tests look good! My only comment is, can you add docstrings to the `set` methods, like this one from `String`? https://github.com/algorand/pyteal/blob/cabe6b87be35f68128e39795464f1446294992b5/pyteal/ast/abi/string.py#L86-L104

Without these, the `DynamicBytes` and `StaticBytes` `set` methods will show the same docstring as their parent classes, which is no longer complete.

PR #501, opened 2022-08-11
add ExecuteMethod helper

barnjamin:
Similar to InnerTxnBuilder.Execute but for MethodCall

### DIFF
--- a/pyteal/ast/itxn.py
+++ b/pyteal/ast/itxn.py
@@ -301,7 +301,7 @@ class InnerTxnBuilder:
                 app_id=app_id,
                 method_signature=method_signature,
                 args=args,
-                extra_fields=cast(dict, extra_fields),
+                extra_fields=extra_fields,
             ),
             cls.Submit(),
         )
@@ -313,7 +313,7 @@ class InnerTxnBuilder:
         app_id: Expr,
         method_signature: str,
         args: list[abi.BaseType | Expr | dict[TxnField, Expr | list[Expr]]],
-        extra_fields: dict[TxnField, Expr | list[Expr]] = {},
+        extra_fields: dict[TxnField, Expr | list[Expr]] = None,
     ) -> Expr:
         """Adds an ABI method call transaction to the current inner transaction group.
 
@@ -482,7 +482,7 @@ class InnerTxnBuilder:
             # Set the fields for the app call in app args and foreign arrays
             *fields_to_set,
             # Add any remaining fields specified by the user
-            InnerTxnBuilder.SetFields(extra_fields),
+            InnerTxnBuilder.SetFields(cast(dict, extra_fields)),
         )
 
 
--- a/pyteal/ast/itxn.py
+++ b/pyteal/ast/itxn.py
@@ -482,7 +482,7 @@ class InnerTxnBuilder:
             # Set the fields for the app call in app args and foreign arrays
             *fields_to_set,
             # Add any remaining fields specified by the user
-            InnerTxnBuilder.SetFields(cast(dict, extra_fields)),
+            InnerTxnBuilder.SetFields({} if extra_fields is None else extra_fields),
         )
 
 
--- a/pyteal/ast/itxn.py
+++ b/pyteal/ast/itxn.py
@@ -217,7 +217,7 @@ class InnerTxnBuilder:
 
             InnerTxnBuilder.Begin()
             InnerTxnBuilder.SetFields(fields)
-            InnerTxnBuilder.End()
+            InnerTxnBuilder.Submit()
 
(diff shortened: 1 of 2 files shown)


jasonpaulos (commented):
Minor comments, otherwise looks good

pyteal/ast/itxn_test.py
```diff
@@ -459,240 +459,50 @@ def test_InnerTxnBuilder_method_call(
...
-        "add(uint64,uint64)void",
-        [pt.abi.Uint64(), pt.abi.Uint64(), pt.abi.Uint64()],
-        None,
-        None,
-        pt.TealInputError,
-    ),
+@pytest.mark.parametrize(
+    "app_id, sig, args, extra_fields, expected_expr, expected_error", ITXN_METHOD_CASES
 )
+def test_InnerTxnBuilder_execute_method(
```
ahangsu, line 465: I think this test is largely redundant with `test_InnerTxnBuilder_method_call`, are they mergable?

pyteal/ast/itxn.py
```diff
@@ -249,6 +249,63 @@ def SetFields(cls, fields: dict[TxnField, Expr | list[Expr]]) -> Expr:
...
+        .. code-block:: python
+
+            InnerTxnBuilder.Begin()
+            InnerTxnBuilder.MethodCall(
+                app_id=app_id,
+                method_signature=method_signature,
+                args=args,
+                extra_fields=extra_fields,
+            ),
+            InnerTxnBuilder.End()
```
ahangsu, line 274: end or submit?

pyteal/ast/itxn_test.py
barnjamin: I combined them 👍

pyteal/ast/itxn.py
```diff
@@ -249,6 +249,63 @@ def SetFields(cls, fields: dict[TxnField, Expr | list[Expr]]) -> Expr:
...
         return Seq(fieldsToSet)
 
+    @classmethod
+    def ExecuteMethodCall(
+        cls,
+        *,
+        app_id: Expr,
+        method_signature: str,
+        args: list[abi.BaseType | Expr | dict[TxnField, Expr | list[Expr]]],
+        extra_fields: dict[TxnField, Expr | list[Expr]] = {},
```
ahangsu, line 259: wait... is `{}` mutable? consider using `None` as default and internally cast to `dict`, and it would apply to `MethodCall`...?

pyteal/ast/itxn_test.py
```diff
@@ -458,241 +466,26 @@ def test_InnerTxnBuilder_method_call(
...
+        pt.InnerTxnBuilder.Begin(), expected_expr, pt.InnerTxnBuilder.Submit()
+    ).__teal__(avm6Options)
+    expected.addIncoming()
+    expected = pt.TealBlock.NormalizeBlocks(expected)
 
+    actual, _ = expr.__teal__(avm6Options)
+    actual.addIncoming()
+    actual = pt.TealBlock.NormalizeBlocks(actual)
 
-# txn_test.py performs additional testing
```
jasonpaulos, line 698: Retain comment?

pyteal/ast/itxn.py
jasonpaulos: `MethodCall` also needs to be updated with this. I'm not sure how the existing code works without this 🤔

ahangsu: ^ `MethodCall` argument `extra_fields` is also default to be `{}`... my bad didn't catch it before.

barnjamin: ah, I'll need one of you to explain to me why this is bad plz

I think I've changed it to what it should be now

pyteal/ast/itxn.py
```diff
@@ -425,7 +482,7 @@ def MethodCall(
             # Set the fields for the app call in app args and foreign arrays
             *fields_to_set,
             # Add any remaining fields specified by the user
-            InnerTxnBuilder.SetFields(extra_fields),
+            InnerTxnBuilder.SetFields(cast(dict, extra_fields)),
```
ahangsu, line 485: I don't think this is the right way to do:
```suggestion
            InnerTxnBuilder.SetFields({} if extra_fields is None else extra_fields),
```

The mutable default argument in python is somewhat notorious, see https://docs.python-guide.org/writing/gotchas/ as a small example. In our case, we do not internally change stuffs to such argument, but it is a code pattern to avoid in python.

ahangsu (approved):
my comments are all resolved, waiting on other folks' pass.

jasonpaulos (approved):
Looks good, thanks for incorporating feedback.

PR #503, opened 2022-08-11
Update `Block` docs to match spec change

jasonpaulos:
Update `Block` docs to match the recent spec change of the `block` opcode: https://github.com/algorandfoundation/specs/pull/71


michaeldiamant (approved):
@jasonpaulos Thanks for recalling to propagate the spec update!

PR #506, opened 2022-08-15
trying assert in commented expression

barnjamin:
Trying to address the comment here: https://github.com/algorand/pyteal/pull/410#issuecomment-1205148027

Should supersede #410  

Also adding the traces when in debug mode

```py
from pyteal import (
    Seq,
    Subroutine,
    TealType,
    Assert,
    Int,
    compileTeal,
    Mode,
    Comment,
    Pop,
)


@Subroutine(TealType.uint64)
def tst_subr():
    """Subroutine commented here"""
    return Seq(Assert((Int(0), "newp")), Int(1))


program = Seq(
    Comment(
        """
    
    Dope program section here

    """
    ),
    Comment("popper", Pop(Int(1))),
    # Make this be
    Assert((Int(1), "ok")),
    # On different
    Assert((Int(1), "no way"), (Int(1), "wow")),
    # lines
    tst_subr(),
)


print(compileTeal(program, mode=Mode.Application, version=7, debug=True))
```
returns
```
#pragma version 7
// "/home/ben/pyteal/pyteal/ast/comment.py":l66
 // 
// "/home/ben/pyteal/pyteal/ast/comment.py":l66
 //     
// "/home/ben/pyteal/pyteal/ast/comment.py":l66
 //     Dope program section here
// "/home/ben/pyteal/pyteal/ast/comment.py":l66
 // 
// "/home/ben/pyteal/pyteal/ast/comment.py":l66
 //     
// "/home/ben/pyteal/examples/application/assert_comment.py":l28
 int 1
// "/home/ben/pyteal/pyteal/ast/unaryexpr.py":l121
 pop // popper
// "/home/ben/pyteal/examples/application/assert_comment.py":l30
 int 1
// "/home/ben/pyteal/examples/application/assert_comment.py":l30
 assert // ok
// "/home/ben/pyteal/examples/application/assert_comment.py":l32
 int 1
// "/home/ben/pyteal/examples/application/assert_comment.py":l32
 assert // no way
// "/home/ben/pyteal/examples/application/assert_comment.py":l32
 int 1
// "/home/ben/pyteal/examples/application/assert_comment.py":l32
 assert // wow
// "/home/ben/pyteal/pyteal/ast/subroutine.py":l290
 callsub tstsubr_0
// "/home/ben/pyteal/examples/application/assert_comment.py":l20
 return

// tst_subr
tstsubr_0:
// "/home/ben/pyteal/examples/application/assert_comment.py":l17
 int 0
// "/home/ben/pyteal/examples/application/assert_comment.py":l17
 assert // newp
// "/home/ben/pyteal/examples/application/assert_comment.py":l17
 int 1
// "/home/ben/pyteal/examples/application/assert_comment.py":l17
 retsub
```

With `debug=False`

```
#pragma version 7
// 
//     
//     Dope program section here
// 
//     
int 1
pop // popper
int 1
assert // ok
int 1
assert // no way
int 1
assert // wow
callsub tstsubr_0
return

// tst_subr
tstsubr_0:
int 0
assert // newp
int 1
retsub
```

### DIFF
--- a/pyteal/ast/assert_.py
+++ b/pyteal/ast/assert_.py
@@ -3,6 +3,7 @@ from typing import TYPE_CHECKING, Optional, cast
 from pyteal.types import TealType, require_type
 from pyteal.ir import TealOp, Op, TealBlock, TealSimpleBlock, TealConditionalBlock
 from pyteal.ast.expr import Expr
+from pyteal.ast.comment import Comment
 from pyteal.ast.seq import Seq
 
 if TYPE_CHECKING:
@@ -56,8 +57,10 @@ class Assert(Expr):
         if len(self.cond) > 1:
             asserts = []
             for idx, cond in enumerate(self.cond):
-                a = Assert((cond, self.comments[idx]))
+                a = Assert(cond)
                 a.trace = cond.trace
+                if self.comments[idx] is not None:
+                    a = cast(Assert, Comment(cast(str, self.comments[idx]), a))
                 asserts.append(a)
 
             return Seq(*asserts).__teal__(options)
--- a/pyteal/ast/assert_.py
+++ b/pyteal/ast/assert_.py
@@ -12,7 +12,9 @@ if TYPE_CHECKING:
 class Assert(Expr):
     """A control flow expression to verify that a condition is true."""
 
-    def __init__(self, cond: Expr, *additional_conds: Expr) -> None:
+    def __init__(
+        self, cond: Expr | tuple[Expr, str], *additional_conds: Expr | tuple[Expr, str]
+    ) -> None:
         """Create an assert statement that raises an error if the condition is false.
 
         Args:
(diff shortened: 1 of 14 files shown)


pyteal/ast/assert_.py
```diff
@@ -12,22 +12,53 @@
...
+            extra_conds.append(cond_single)
+
+        self.cond: list[Expr] = [cond] + list(extra_conds)
+        self.comments: list[str] = comments
 
     def __teal__(self, options: "CompileOptions"):
         if len(self.cond) > 1:
-            asserts = [Assert(cond) for cond in self.cond]
+            asserts = []
+            for idx, cond in enumerate(self.cond):
```
barnjamin, line 57: Changed from list comprehension so we can set the trace to where it was initially defined

pyteal/ast/assert_.py
```diff
@@ -12,22 +13,56 @@
 class Assert(Expr):
     """A control flow expression to verify that a condition is true."""
 
-    def __init__(self, cond: Expr, *additional_conds: Expr) -> None:
+    def __init__(
+        self,
+        cond: Expr | tuple[Expr, Optional[str]],
+        *additional_conds: Expr | tuple[Expr, Optional[str]]
```
jasonpaulos, line 19: Alternative interface idea: what about adding a keyword arg like this?

```python
def __init__(self, cond: Expr, *additional_conds: Expr, *, comment: str) -> None:
```

A benefit of this is that I think it'd be easier for the developer to add/remove these as needed when debugging, and it's very explicit.

A downside is that multi-clause Asserts aren't as well supported. However, I argue this isn't as important, since the caller can always refactor `Assert(x, y, z)` into `Seq(Assert(x, comment="..."), Assert(y, z))` if they want to really target their comments.

pyteal/ast/assert_.py
```diff
@@ -12,22 +13,56 @@
...
 
     def __teal__(self, options: "CompileOptions"):
         if len(self.cond) > 1:
-            asserts = [Assert(cond) for cond in self.cond]
+            asserts = []
+            for idx, cond in enumerate(self.cond):
+                a = Assert(cond)
+                a.trace = cond.trace
+                if self.comments[idx] is not None:
+                    a = cast(Assert, Comment(cast(str, self.comments[idx]), a))
```
jasonpaulos, line 63: I don't like this -- how about declare `asserts: list[Expr] = []` instead of a bad cast?

pyteal/compiler/compiler.py
```diff
@@ -237,6 +239,7 @@ def compileTeal(
     version: int = DEFAULT_PROGRAM_VERSION,
     assembleConstants: bool = False,
     optimize: OptimizeOptions = None,
+    debug: bool = False,
```
jasonpaulos, line 242: Debug comments look just like a source map to me. I think it would be better to actually return a distinct source map object at this point.

PR #508, opened 2022-08-15
Fix bug where `String.set` duplicates the argument `Expr` in the AST

jasonpaulos:
When called with an `Expr`, the existing `_encoded_string` helper function includes the argument twice in the returned AST. For idempotent expressions, this is fine, but this is not acceptable for expressions whose execution has side effects.

For example, this code would produce an unintuitive result:

```python
value = abi.String()

program = value.set(
    Seq(
        Log(Bytes("hello")),
        Bytes("value")
    )
)
```

Because the current implementation duplicates the argument `Expr` in the AST, the example code above would actually log "hello" twice.

---

As a side effect of the fix in this PR, the implementation of `String.set` when called with a Python string or byte array is now more efficient, since the encoding is computed at compile time.

### DIFF
--- a/pyteal/ast/abi/string.py
+++ b/pyteal/ast/abi/string.py
@@ -1,6 +1,8 @@
 from typing import Union, Sequence, cast
 from collections.abc import Sequence as CollectionSequence
 
+from algosdk.abi import ABIType
+
 from pyteal.ast.abi.uint import Byte
 from pyteal.ast.abi.type import ComputedValue, BaseType
 from pyteal.ast.abi.array_dynamic import DynamicArray, DynamicArrayTypeSpec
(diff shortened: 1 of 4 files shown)


pyteal/ast/abi/string.py
```diff
@@ -9,15 +11,27 @@
...
 from pyteal.ast.substring import Suffix
 from pyteal.ast.naryexpr import Concat
 
 from pyteal.errors import TealInputError
 
 
-def _encoded_string(s: Expr):
-    return Concat(Suffix(Itob(Len(s)), Int(6)), s)
+def _encoded_byte_string(s: bytes | bytearray) -> Expr:
+    prefix = ABIType.from_string("uint16").encode(len(s))
```
ahangsu, line 24: wow... py-sdk ABI in pyteal

ahangsu (approved):
lookin good, the "bug" of double side effect is subtle

barnjamin (approved):

PR #509, opened 2022-08-16
Comment vanish on `Seq()` or empty expr

ahangsu:
A helper PR to #410, let `Comment` return `Seq()` on input expression being empty `Seq()`.

Also handle cascaded case for `Comment(Comment(Seq(), "level0"), "level1")`.


ahangsu:
I got it wrong. I screwed up. Nuke it.

PR #510, opened 2022-08-16
adding comment kwarg to assert expression

barnjamin:
Taking Jasons suggestion from https://github.com/algorand/pyteal/pull/506#discussion_r946167309 and adding a keyword argument to the Assert expression.

This will add the _same_ comment just before the `assert` in the resulting TEAL. 

If the caller wishes to have different messages for each argument, it can be refactored to call each explicitly.

This seems like the best way to have the comment be in a predictable location (just above assert) for giving a reason the assert occurred.

### DIFF
--- a/pyteal/ast/assert_.py
+++ b/pyteal/ast/assert_.py
@@ -21,6 +21,7 @@ class Assert(Expr):
         Args:
             cond: The condition to check. Must evaluate to a uint64.
             *additional_conds: Additional conditions to check. Must evaluate to uint64.
+            comment: String comment to place on the line immediately prior to the assert op
         """
         super().__init__()
         require_type(cond, TealType.uint64)
--- a/pyteal/ast/assert_.py
+++ b/pyteal/ast/assert_.py
@@ -3,6 +3,7 @@ from typing import TYPE_CHECKING
 from pyteal.types import TealType, require_type
 from pyteal.ir import TealOp, Op, TealBlock, TealSimpleBlock, TealConditionalBlock
 from pyteal.ast.expr import Expr
+from pyteal.ast.comment import Comment
 from pyteal.ast.seq import Seq
 
 if TYPE_CHECKING:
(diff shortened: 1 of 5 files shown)


pyteal/ast/assert_.py
```diff
@@ -12,7 +13,9 @@
 class Assert(Expr):
     """A control flow expression to verify that a condition is true."""
 
-    def __init__(self, cond: Expr, *additional_conds: Expr) -> None:
+    def __init__(
+        self, cond: Expr, *additional_conds: Expr, comment: str = None
+    ) -> None:
```
jasonpaulos, line 18: Add new arg to docstring?

pyteal/ast/assert_.py
```diff
@@ -23,16 +26,25 @@ def __init__(self, cond: Expr, *additional_conds: Expr) -> None:
...
+
+        self.comment = comment
         self.cond = [cond] + list(additional_conds)
 
     def __teal__(self, options: "CompileOptions"):
         if len(self.cond) > 1:
-            asserts = [Assert(cond) for cond in self.cond]
+            asserts: list[Expr] = []
+            for cond in self.cond:
+                asrt = Assert(cond, comment=self.comment)
```
jasonpaulos, line 37: I don't love that this duplicate the comment for every child condition, but I'm not sure what would be better

barnjamin: Agree, its weird. This seemed like the best compromise and the author can always refactor to individual Asserts.

jasonpaulos: Ok, I'm happy enough with this at the moment to approve. We can revisit the behavior in the future if users complain.

jasonpaulos (approved):
I suggest also adding a changelog line, but otherwise, this looks good

ahangsu (approved):
generally lookin good, just one question not blocking my approval

pyteal/ast/assert_.py
```diff
@@ -12,27 +13,39 @@
...
+                asrt.trace = cond.trace
```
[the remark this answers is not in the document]
barnjamin: trace is useful for the error messages, when we create a new Expr within pyteal the trace gets set to the internal pyteal file rather than the authors source file (which is probably what they actually want)

By setting the trace here to the conds trace, we're passing the info about where it was initially declared to be at least much closer to the location it was declared in the authors source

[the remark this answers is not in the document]
ahangsu: I see, that makes sense.

PR #511, opened 2022-08-18
removing error thrown if seq contains something other than an expression

barnjamin:
Why not?

```py
program = Seq(
    x := ScratchVar(), 
    x.store(Int(10)), 
    x.load()
)

print(compileTeal(program, mode=Mode.Application, version=7))
```

### DIFF
--- a/pyteal/ast/seq.py
+++ b/pyteal/ast/seq.py
@@ -47,13 +47,16 @@ class Seq(Expr):
         if len(exprs) == 1 and isinstance(exprs[0], list):
             exprs = exprs[0]
 
+        filtered_exprs: list[Expr] = []
         for i, expr in enumerate(exprs):
             if not isinstance(expr, Expr):
-                raise TealInputError("{} is not a pyteal expression.".format(expr))
+                continue
             if i + 1 < len(exprs):
                 require_type(expr, TealType.none)
 
-        self.args = exprs
+            filtered_exprs.append(expr)
+
+        self.args = filtered_exprs
 
     def __teal__(self, options: "CompileOptions"):
         start = TealSimpleBlock([])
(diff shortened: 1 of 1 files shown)


pyteal/ast/seq.py
```diff
@@ -47,13 +47,16 @@ def __init__(self, *exprs):
         if len(exprs) == 1 and isinstance(exprs[0], list):
             exprs = exprs[0]
 
+        filtered_exprs: list[Expr] = []
         for i, expr in enumerate(exprs):
             if not isinstance(expr, Expr):
-                raise TealInputError("{} is not a pyteal expression.".format(expr))
+                continue
```
michaeldiamant, line 53: @barnjamin Optional:  If not opposed, can you add a brief inline comment explaining why `Seq` construction validation is lax?

The PR description captures the spirit of what I have in mind.  If you prefer as is, feel welcomed to resolve.

michaeldiamant (approved):
@barnjamin I think it's worthwhile to get 1 more review before merging.  As discussed in Slack, I didn't come up with a _good_ reason to hold here.

Aside:  Can you update the CHANGELOG prior to merge?

jasonpaulos (commented):
I think you're approaching this from the wrong side. If only an exception is made for `Seq`, what about `Assert(...)`, `While(x).Do(...)`, etc.?

I think it would be easier/more cohesive to make `ScratchVar` inherit from `Expr` and return an empty TealBlock from the `__teal__` method. That would achieve the goal you're looking for.

michaeldiamant:
>  think it would be easier/more cohesive to make ScratchVar inherit from Expr and return an empty TealBlock from the __teal__ method. 

* I agree it's a more cohesive change, but I'm initially skeptical it's a worthwhile tradeoff.  I think a verbal might help me better understand + convince me.
* If it feels like a clear win to folks, I'm not looking to stand in the way.  Since I already reviewed the PR, I felt compelled to chime in.

barnjamin:
Closing this for now since there is probably a better general approach by making `ScratchVar` (or some new thing that acts like `ScratchVar`) be an `Expr`

PR #514, opened 2022-08-19
Modify ABI DynamicBytes and StaticBytes to support router/subroutine

fabrice102:
Router/subroutine requires that an associated TypeSpec
to act as a factory for those types.

Indeed, if this is not provided, the router will execute the
methods using the base types (`StaticArray/DynamicArray`)
 rather than the derived types (`StaticBytes/DynamicBytes`)
which prevents use of the new functions `get/set`.

This commit is meant to add this associated TypeSpec.

Notes:
1. I am not very familiar with the code and may
have made mistakes. However, unit tests work and the code
works as expected on a project I'm working on.
2. It would be great to add router tests and documentation
about using `DynamicBytes/StaticBytes` as this is expected
to be quite common. It's important people don't revert
to `string` for that purpose as `string` does not carry the
right semantic (it implies it is UTF-8).

### DIFF
--- a/pyteal/ast/abi/array_dynamic.py
+++ b/pyteal/ast/abi/array_dynamic.py
@@ -18,7 +18,7 @@ class DynamicArrayTypeSpec(ArrayTypeSpec[T]):
     def new_instance(self) -> "DynamicArray[T]":
         return DynamicArray(self)
 
-    def annotation_type(self) -> "type[DynamicArray[T]]":
+    def annotation_type(self) -> type["DynamicArray[T]"]:
         return DynamicArray[self.value_type_spec().annotation_type()]  # type: ignore[misc]
 
     def is_length_dynamic(self) -> bool:
@@ -107,7 +107,7 @@ class DynamicBytesTypeSpec(DynamicArrayTypeSpec):
     def new_instance(self) -> "DynamicBytes":
         return DynamicBytes()
 
-    def annotation_type(self) -> "type[DynamicBytes]":
+    def annotation_type(self) -> type["DynamicBytes"]:
         return DynamicBytes
 
     def __str__(self) -> str:
--- a/pyteal/ast/abi/array_dynamic.py
+++ b/pyteal/ast/abi/array_dynamic.py
@@ -100,7 +100,7 @@ class DynamicArray(Array[T]):
 DynamicArray.__module__ = "pyteal.abi"
 
 
-class DynamicBytesTypeSpec(DynamicArrayTypeSpec):
+class DynamicBytesTypeSpec(DynamicArrayTypeSpec[Byte]):
     def __init__(self) -> None:
         super().__init__(ByteTypeSpec())
 
--- a/pyteal/ast/abi/array_dynamic.py
+++ b/pyteal/ast/abi/array_dynamic.py
@@ -100,11 +100,34 @@ class DynamicArray(Array[T]):
 DynamicArray.__module__ = "pyteal.abi"
 
 
+class DynamicBytesTypeSpec(DynamicArrayTypeSpec):
+    def __init__(self) -> None:
+        super().__init__(ByteTypeSpec())
+
+    def new_instance(self) -> "DynamicBytes":
+        return DynamicBytes()
+
+    def annotation_type(self) -> "type[DynamicBytes]":
+        return DynamicBytes
+
+    def __str__(self) -> str:
+        return "byte[]"
+
+    def __eq__(self, other: object) -> bool:
+        return isinstance(other, DynamicBytesTypeSpec)
+
+
+DynamicBytesTypeSpec.__module__ = "pyteal.abi"
+
+
 class DynamicBytes(DynamicArray[Byte]):
     """The convenience class that represents ABI dynamic byte array."""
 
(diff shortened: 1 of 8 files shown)


pyteal/ast/abi/array_dynamic.py
```diff
@@ -100,11 +100,34 @@ def length(self) -> Expr:
...
 
 
+class DynamicBytesTypeSpec(DynamicArrayTypeSpec):
+    def __init__(self) -> None:
+        super().__init__(ByteTypeSpec())
+
+    def new_instance(self) -> "DynamicBytes":
+        return DynamicBytes()
+
+    def annotation_type(self) -> "type[DynamicBytes]":
```
barnjamin, line 110:
```suggestion
    def annotation_type(self) -> type["DynamicBytes"]:
```

barnjamin: I note the other annotation_type above has the same quote around "type[DynamicArray[T]]", if the above change is correct the other one should be tweaked as well

pyteal/ast/abi/array_static.py
```diff
@@ -149,11 +149,30 @@ def __getitem__(self, index: Union[int, Expr]) -> "ArrayElement[T]":
...
 
 
+class StaticBytesTypeSpec(StaticArrayTypeSpec[Byte, N], Generic[N]):
+    def __init__(self, array_length: int) -> None:
+        super().__init__(ByteTypeSpec(), array_length)
+
+    def new_instance(self) -> "StaticBytes[N]":
+        return StaticBytes(self)
+
+    def annotation_type(self) -> "type[StaticBytes[N]]":
```
barnjamin, line 159: same, type should already be defined so only StaticBytes need be quoted

pyteal/ast/abi/array_static_test.py
```diff
@@ -255,7 +257,7 @@ def test_StaticArray_set_computed():
 
 @pytest.mark.parametrize("test_case", BYTES_SET_TESTCASES)
 def test_StaticBytes_set_py_bytes(test_case: bytes | bytearray):
-    value = abi.StaticBytes(len(test_case))
+    value: abi.StaticBytes[Literal[16]] = abi.StaticBytes(abi.StaticBytesTypeSpec(16))
```
barnjamin, line 260: @fabrice102 any reason for this type def with explicit literal? Omitting `Literal[16]` should prevent mypy from complaining and we can get the expected length to pass with just `len(test_case)` so we don't need to hardcode 16

barnjamin, line 260:
```suggestion
    value: abi.StaticBytes = abi.StaticBytes(abi.StaticBytesTypeSpec(16))
```

pyteal/ast/abi/array_static_test.py
```diff
@@ -281,7 +283,9 @@ def test_StaticBytes_set_py_bytes(test_case: bytes | bytearray):
 
 @pytest.mark.parametrize("test_case", BYTES_SET_TESTCASES)
 def test_StaticBytes_expr(test_case: bytes | bytearray):
-    value = abi.StaticBytes(len(test_case) * 2)
+    value: abi.StaticBytes[Literal[32]] = abi.StaticBytes(
+        abi.StaticBytesTypeSpec(16 * 2)
+    )
```
barnjamin, line 288: same as above

ahangsu (commented):
Generally looking good to me, Ben covered most of the points. Could you also add a line to `CHANGELOG.md`, under section `Unreleased/Fixed`:
```markdown
- Modify ABI `DynamicBytes` and `StaticBytes` to support `router`/`subroutine` [#514](https://github.com/algorand/pyteal/pull/514)
```

pyteal/ast/abi/array_static_test.py
fabrice102: My `mypy` was complaining without the typedef for some reasons.
I don't understand why though...

barnjamin: Even with the `abi.StaticBytes` (instead of `abi.StaticBytes[Literal[16]]`)?

I just tried it locally and it complained with no type but was fine with omitted literal

ahangsu (commented):
Looking good to me, would you also merge a commit from this branch: https://github.com/algorand/pyteal/tree/temp-bytes-typespec? It is one commit ahead your local repo, containing related testing for new typespecs.

Once new set of tests are added, I am approving.

barnjamin (approved):

ahangsu (approved):

pyteal/ast/abi/array_dynamic.py
```diff
@@ -100,11 +100,34 @@ def length(self) -> Expr:
 DynamicArray.__module__ = "pyteal.abi"
 
 
+class DynamicBytesTypeSpec(DynamicArrayTypeSpec):
```
ahangsu, line 103: nit: 
```suggestion
class DynamicBytesTypeSpec(DynamicArrayTypeSpec[ByteType]):
```

jasonpaulos: +1 to this, except `ByteType` should just be `Byte`

PR #518, opened 2022-08-29
Improvement: Docstring Parsing for Method Description

### LINKED ISSUE #516: Improve method JSON descriptions
Currently, the JSON description field for an ABI method is simply the method's Python docstring, with all newlines turned into spaces: https://github.com/algorand/pyteal/blob/a66907ef4341db8e595f97f4a53413d6338f6c74/pyteal/ast/subroutine.py#L630-L637

This can be improved in a few ways:
1. Keep filtering out single newlines (which are used to wrap a single line in the Python docstring), but treat double newlines as a newline in the JSON doc string. For example, given the following Python docstring:
```python
def example1():
    """This is an example docstring. This first line wraps since it's too
    long to fit in a single line of source code.
    
    This is a second line.
    """
    pass
```
We should produce this JSON string:
```json
"This is an example docstring. This first line wraps since it's too long to fit in a single line of source code.\nThis is a second line."
```
2. Right now, there is no way to populate the description fields of arguments or return values. I propose that we adopt some way to parse a Python docstring and extract information about these fields. For example, given the following Python docstring:
```python
def example2(a: abi.Uint8, b: abi.Uint32, *, output: abi.Uint64):
    """This is a second example docstring.
    
    Args:
        a: A description for the first argument.
        b: A description for the second argument.
    
    Returns:
        A description for the return value.
    """
    pass
```
We should produce the following method JSON:
```json
{
  "name": "example2",
  "args": [
    {
      "type": "uint8",
      "name": "a",
      "desc": "A description for the first argument."
    },
    {
      "type": "uint32",
      "name": "b",
      "desc": "A description for the second argument."
    }
  ],
  "returns": {
    "type": "uint64",
    "desc": "A description for the return value."
  },
  "desc": "This is a second example docstring."
}
```

barnjamin:
Closes #516


docs/requirements.txt
```diff
@@ -3,3 +3,5 @@ sphinx-rtd-theme==1.0.0
 # dependencies from setup.py
 py-algorand-sdk>=1.9.0,<2.0.0
 semantic-version>=2.9.0,<3.0.0
+# dependencies from ./requirement.txt
+docstring-parser==0.14.1
```
ahangsu, line 7: ref: https://stackoverflow.com/questions/15889621/sphinx-how-to-exclude-imports-in-automodule

This line is simply put to make all package imported for pyteal source code, so build doc could test pass

pyteal/ast/subroutine.py
```diff
@@ -612,10 +613,33 @@ def method_signature(self, overriding_name: str = None) -> str:
     def method_spec(self) -> sdk_abi.Method:
         skip_names = ["return", "output"]
 
+        desc: str = ""
+        arg_descs: dict[str, str] = {}
+        output_desc: str = ""
+
+        if self.subroutine.implementation.__doc__ is not None:
+            docstring = parse_docstring(self.subroutine.implementation.__doc__)
```
ahangsu, line 621: another problem: you do not know what style of docstring gets passed in, and `parse` can't decide on its own. Seems we don't have ability to tell which style of docstring just from `docstring_parser` lib, and wonder if we can correctly support 4 styles of docstrings.

barnjamin: The `docstring` var should contain an element that specifies what type of docstring it thinks it is, or at least it did in the one example i tried marking it as `Google`

requirements.txt
```diff
@@ -9,3 +9,4 @@ pytest-timeout==2.1.0
 pytest-xdist==2.5.0
 semantic-version==2.10.0
 types-setuptools==57.4.18
+docstring-parser==0.14.1
```
jasonpaulos, line 12: This is a runtime requirement, it should be in `setup.py`

ahangsu: good point, resolving

pyteal/ast/subroutine.py
```diff
@@ -648,8 +648,10 @@ def get_ret_desc(_ret_desc: str, _output_desc: str) -> str:
                     res = res + " " + _output_desc
                 return res
 
+        print(self.subroutine.implementation.__doc__)
         if self.subroutine.implementation.__doc__ is not None:
             docstring = parse_docstring(self.subroutine.implementation.__doc__)
+            print(docstring.params)
```
ahangsu, line 654: nit: print log take out

pyteal/ast/subroutine.py
```diff
@@ -648,8 +648,10 @@ def get_ret_desc(_ret_desc: str, _output_desc: str) -> str:
                     res = res + " " + _output_desc
                 return res
 
+        print(self.subroutine.implementation.__doc__)
```
ahangsu, line 651: nit: print log

pyteal/ast/subroutine.py
```diff
@@ -677,6 +679,8 @@ def get_ret_desc(_ret_desc: str, _output_desc: str) -> str:
 
             return_desc = get_ret_desc(return_desc, output_desc)
 
+        print(arg_descs)
+
```
ahangsu, line 683: nit: print log

pyteal/ast/subroutine.py
```diff
@@ -610,12 +611,77 @@ def method_signature(self, overriding_name: str = None) -> str:
...
+            """
+            if not _ret_desc:
+                return _output_desc
+            else:
+                res = _ret_desc
+                if _output_desc:
+                    res = res + " " + _output_desc
+                return res
+
+        if self.subroutine.implementation.__doc__ is not None:
```
ahangsu, line 651: nit: I think the check on this line should simply 
```suggestion
        if self.subroutine.implementation.__doc__:
```
since we can't parse anything from an empty docstring either

pyteal/ast/subroutine.py
```diff
@@ -610,12 +611,77 @@ def method_signature(self, overriding_name: str = None) -> str:
...
+            - if neither exists, then outputs empty string
+            - if both exists, concat with break line
+            """
+            if not short_desc:
+                return long_desc if long_desc else ""
+            else:
+                method_desc = short_desc
+                if long_desc:
+                    method_desc = method_desc + "\n" + long_desc
+                return method_desc
```
ahangsu, line 634:
```suggestion
                _method_desc = short_desc
                if long_desc:
                    _method_desc = _method_desc + "\n" + long_desc
                return _method_desc
```
tell me if I am paranoid, I don't want this `method_desc` be overlapping with `method_desc` in following section

pyteal/ast/subroutine_test.py
```diff
@@ -1302,3 +1302,112 @@ def mySubroutine(a1, a2, a3, a4, a5, a6, a7, a8, a9, a10):
...
+    assert (
+        mspec_dict["returns"]["desc"]
+        == "A PyTeal expression that sets output Uint64 value as argument a. "
+        + "an abi Uint64 value get set from argument a"
+    )
+
+    @pt.ABIReturnSubroutine
+    def epy_style(a: pt.abi.Uint64, *, output: pt.abi.Uint64):
+        """
+        Example of a ABIReturnSubroutine with Google format docstring.
```
ahangsu, line 1338: nit: change docstring to 
```suggestion
        Example of a ABIReturnSubroutine with epytext format docstring.
```

pyteal/ast/subroutine_test.py
```diff
@@ -1302,3 +1302,112 @@ def mySubroutine(a1, a2, a3, a4, a5, a6, a7, a8, a9, a10):
...
+        @param output: an abi Uint64 value get set from argument a
+        @return: A PyTeal expression that sets output Uint64 value as argument a.
+        """
+        return output.set(a)
+
+    mspec_dict = epy_style.method_spec().dictify()
+
+    assert (
+        mspec_dict["desc"]
+        == "Example of a ABIReturnSubroutine with Google format docstring."
```
ahangsu, line 1350:
```suggestion
        == "Example of a ABIReturnSubroutine with epytext format docstring."
```

pyteal/ast/subroutine_test.py
```diff
@@ -1302,3 +1302,112 @@ def mySubroutine(a1, a2, a3, a4, a5, a6, a7, a8, a9, a10):
...
+    assert (
+        mspec_dict["returns"]["desc"]
+        == "A PyTeal expression that sets output Uint64 value as argument a. "
+        + "an abi Uint64 value get set from argument a"
+    )
+
+    @pt.ABIReturnSubroutine
+    def numpy_style(a: pt.abi.Uint64, *, output: pt.abi.Uint64):
+        """
+        Example of a ABIReturnSubroutine with Google format docstring.
```
ahangsu, line 1362:
```suggestion
        Example of a ABIReturnSubroutine with numpy format docstring.
```

pyteal/ast/subroutine_test.py
```diff
@@ -1302,3 +1302,112 @@ def mySubroutine(a1, a2, a3, a4, a5, a6, a7, a8, a9, a10):
...
+        uint64
+            A PyTeal expression that sets output Uint64 value as argument a.
+
+        """
+        return output.set(a)
+
+    mspec_dict = numpy_style.method_spec().dictify()
+    assert (
+        mspec_dict["desc"]
+        == "Example of a ABIReturnSubroutine with Google format docstring."
```
ahangsu, line 1382:
```suggestion
        == "Example of a ABIReturnSubroutine with numpy format docstring."
```

pyteal/ast/subroutine_test.py
```diff
@@ -1302,3 +1302,112 @@ def mySubroutine(a1, a2, a3, a4, a5, a6, a7, a8, a9, a10):
...
+    assert (
+        mspec_dict["returns"]["desc"]
+        == "A PyTeal expression that sets output Uint64 value as argument a. "
+        + "an abi Uint64 value get set from argument a"
+    )
+
+    @pt.ABIReturnSubroutine
+    def rest_style(a: pt.abi.Uint64, *, output: pt.abi.Uint64):
+        """
+        Example of a ABIReturnSubroutine with Google format docstring.
```
ahangsu, line 1394:
```suggestion
        Example of a ABIReturnSubroutine with REST format docstring.
```

pyteal/ast/subroutine_test.py
```diff
@@ -1302,3 +1302,112 @@ def mySubroutine(a1, a2, a3, a4, a5, a6, a7, a8, a9, a10):
...
+        :param output: an abi Uint64 value get set from argument a
+        :returns: A PyTeal expression that sets output Uint64 value as argument a.
+        """
+        return output.set(a)
+
+    mspec_dict = rest_style.method_spec().dictify()
+
+    assert (
+        mspec_dict["desc"]
+        == "Example of a ABIReturnSubroutine with Google format docstring."
```
ahangsu, line 1406:
```suggestion
        == "Example of a ABIReturnSubroutine with REST format docstring."
```

ahangsu (commented):
nits on testcases

ahangsu (approved):
looks good to me, though I had some commits to this PR, so my approval is simply indicative rather than "approval-approval". need another pair of eyes to read closely.

docs/requirements.txt
```diff
@@ -3,3 +3,4 @@ sphinx-rtd-theme==1.0.0
 # dependencies from setup.py
 py-algorand-sdk>=1.9.0,<2.0.0
 semantic-version>=2.9.0,<3.0.0
+docstring-parser>=0.14.1
```
jasonpaulos, line 6: A maximum version should be enforced, otherwise we may accept a version with breaking changes

pyteal/ast/subroutine.py
```diff
@@ -610,12 +611,77 @@ def method_signature(self, overriding_name: str = None) -> str:
...
+                docstring.short_description, docstring.long_description
+            )
+
+            desc = " ".join(
+                [
+                    i.strip()
+                    for i in method_desc.split("\n")
+                    if not (i.isspace() or len(i) == 0)
+                ]
+            )
```
jasonpaulos, line 663: This does not look like it will satisfy the first item from #516.

To be sure, the tests should include multiple "paragraphs" in their docstrings. Right now they only populate the short description.

pyteal/ast/subroutine.py
```diff
@@ -610,12 +611,77 @@ def method_signature(self, overriding_name: str = None) -> str:
...
+                if long_desc:
+                    _method_desc = _method_desc + "\n" + long_desc
+                return _method_desc
+
+        def get_ret_desc(_ret_desc: str, _output_desc: str) -> str:
+            """
+            return description is based on return desc and output desc:
+            - if only return (or output) desc, then output only return (or output) desc
+            - if neither exists, then outputs empty string
+            - if both exists, concat with space
```
jasonpaulos, line 641: I find this behavior about the return annotation unintuitive. My recommendation: only use the function return annotation and ignore any annotations on the `output` argument.

pyteal/ast/subroutine.py
```diff
@@ -624,18 +690,10 @@ def method_spec(self) -> sdk_abi.Method:
         spec = {
             "name": self.name(),
             "args": args,
-            "returns": {"type": str(self.type_of())},
+            "desc": desc,
```
jasonpaulos, line 693: If no description is provided, I'd prefer this field is omitted from the JSON object. I haven't tested, but the code here suggests it will be populated as an empty string.

pyteal/ast/subroutine.py
```diff
@@ -610,32 +611,71 @@ def method_signature(self, overriding_name: str = None) -> str:
...
+
+                if name in arg_descs:
+                    arg_obj["desc"] = arg_descs[name]
+
+                args.append(arg_obj)
+
+        # Create the return obj for the method, adding description if set
+        return_obj = {"type": str(self.type_of())}
+        if return_desc:
+            return_obj["desc"] = return_desc
```
jasonpaulos, line 672: Suggestion: don't add this description if the method returns void.

jasonpaulos (commented):
I made a child PR, but it showcases some new problems, namely with newlines that are present in the argument descriptions: https://github.com/algorand/pyteal/pull/520

pyteal/ast/subroutine.py
barnjamin: I did this, but I'm not sure if we should skip it. 

If the author takes the time to write a return description, why not add it?

jasonpaulos: The arguments as I see it are:

### In favor of always annotating the return description:
* If the author wrote a return annotation, might as well carry it forward to the contract JSON. They can always proofread the generated JSON and remove the annotation if they see fit.

### Against always annotating the return description:
* For an ABI method that returns void, there is no reason to write a return description; anything you have to say about what the function does should go into the general function description. Yes, PyTeal method implementations *do* return an `Expr`, but this is only a language-level convention and irrelevant to an external caller.
* I suspect (but don't know how common this really is) that developers tend to write return annotations in the form of "Returns a PyTeal expression that..." (that's at least what we do internally). This type of explanation is meaningless outside of the realm of PyTeal and would only serve to add noise to a contract JSON.

I'm still slightly in favor of the current behavior (dropping return description if return is void), but I don't feel so strongly about it.

examples/application/abi/algobank.json
```diff
@@ -36,17 +40,19 @@
...
                     "type": "uint64",
-                    "name": "amount"
+                    "name": "amount",
+                    "desc": "The amount of Algos requested to be withdraw, in microAlgos. This method will fail\nif this amount exceeds the amount of Algos held by this app for the method call sender."
                 },
                 {
                     "type": "account",
-                    "name": "recipient"
+                    "name": "recipient",
+                    "desc": "An account who will receive the withdrawn Algos. This may or may not be the same\nas the method call sender."
```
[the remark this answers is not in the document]
barnjamin: Ah I didn't read it carefully, I thought you'd found something but fixed it in that PR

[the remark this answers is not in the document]
barnjamin: aight, this should be fixed with the latest

pyteal/ast/subroutine.py
```diff
@@ -610,32 +611,73 @@ def method_signature(self, overriding_name: str = None) -> str:
...
+            method_desc = "\n\n".join(method_descriptions)
+
+            # Turn double new line into single, replacing single newline with space
+            desc = "\n".join(
                 [
-                    i.strip()
-                    for i in self.subroutine.implementation.__doc__.split("\n")
+                    i.replace("\n", " ").strip()
+                    for i in method_desc.split("\n\n")
                     if not (i.isspace() or len(i) == 0)
```
ahangsu, line 636: minor trick, I think you might like 
```suggestion
                    if i.strip()
```
for it covers both check non space and len > 0

pyteal/ast/subroutine_test.py
```diff
@@ -1302,3 +1302,159 @@ def mySubroutine(a1, a2, a3, a4, a5, a6, a7, a8, a9, a10):
...
+    assert mspec_dict["desc"] == short_desc
+    assert mspec_dict["args"][0]["desc"] == a_doc
+    assert mspec_dict["returns"]["desc"] == return_doc
+
+    # Short and long descriptions
+    documented_method.__doc__ = f"""{long_desc}
+
+    :param a: {a_doc} 
+    :returns: {return_doc}
+    """
```
ahangsu, line 1389: is this really short and long description?

barnjamin: doesnt look like it 🤔

ahangsu: nvm, I think `long_desc` has multi-lines, and the first line can serve as short description, no worries.

barnjamin: oh, i see its "both" by having no newline before it. See the comment below this one

pyteal/ast/subroutine.py
```diff
@@ -610,32 +611,73 @@ def method_signature(self, overriding_name: str = None) -> str:
...
+            # Get the special return description
+            return_desc = (
+                ""
+                if not docstring.returns or not docstring.returns.description
+                else docstring.returns.description.replace("\n", " ").strip()
+            )
+
+            # Generate the ABI method object given the subroutine args
+            # Add in description if one is set
+            for name, val in self.subroutine.annotations.items():
```
ahangsu, line 656: just a comment when I am concerned about `args` ordering.

we are lucky this time: since py 3.7, `dict` retains insertion order. So when we generate `args`, the ordering is guaranteed so long as the `_validate` method in `subroutine`, which relies on `get_annotation` returns right order.

ahangsu (approved):

jasonpaulos (approved):
Looks good, thanks for taking suggestions

PR #519, opened 2022-08-29
Test round-trip `TypeSpec` values

jasonpaulos:
In an effort to avoid new bugs like the one that https://github.com/algorand/pyteal/pull/514 pointed out, I've added new assertions to `test_type_spec_from_annotation`.

The original issue was that `DynamicBytes().type_spec().new_instance()` _did not_ produce another `DynamicBytes` instance (`StaticBytes` was similarly affected).

The new assertions ensures that all tested `TypeSpec` implementations have a `new_instance()` method that produces a value that is actually an instance of the annotation type, and whose `type_spec()` is equal to the original `TypeSpec` that created the value.

Additional changes:
* Migrated tests in `pyteal/ast/abi/util_test.py` to use `pytest.mark.parametrize` to accept inputs.
* Added previously untested cases to `test_type_spec_from_annotation`.
* I discovered we had another instance of the original issue, with the `TupleX` classes, meaning if you defined a subroutine that takes a `TupleX` argument, you would actually receive a `Tuple` instance. I think this is a bug in name only, as no incorrect behavior could result from this substitution. Still, to be consistent, this PR also changes the `TupleTypeSpec.new_instance()` method to return an instance of a `TupleX` class when possible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+## Fixed
+* Subroutines annotated with a `TupleX` class are now invoked with an instance of that exact class, instead of the more general `Tuple` class ([#519](https://github.com/algorand/pyteal/pull/519))
+
 # 0.17.0
 
 ## Added
--- a/pyteal/ast/abi/util_test.py
+++ b/pyteal/ast/abi/util_test.py
@@ -117,7 +117,7 @@ def test_substringForDecoding(
 
 class IntAnnotationTest(NamedTuple):
     annotation: Any
-    expected: int | Exception
+    expected: int | type[Exception]
 
 
 IN_ANNOTATION_TEST_CASES: list[IntAnnotationTest] = [
@@ -134,8 +134,8 @@ IN_ANNOTATION_TEST_CASES: list[IntAnnotationTest] = [
 
 
 @pytest.mark.parametrize("annotation, expected", IN_ANNOTATION_TEST_CASES)
-def test_int_literal_from_annotation(annotation: Any, expected: int | Exception):
-    if type(expected) is not int:
+def test_int_literal_from_annotation(annotation: Any, expected: int | type[Exception]):
+    if not isinstance(expected, int):
         with pytest.raises(expected):
             int_literal_from_annotation(annotation)
         return
@@ -146,7 +146,7 @@ def test_int_literal_from_annotation(annotation: Any, expected: int | Exception)
 
 class TypeAnnotationTest(NamedTuple):
     annotation: Any
-    expected: abi.TypeSpec | Exception
+    expected: abi.TypeSpec | type[Exception]
 
 
 class ExampleNamedTuple(abi.NamedTuple):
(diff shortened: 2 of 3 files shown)


pyteal/ast/abi/util_test.py
```diff
@@ -10,289 +10,350 @@
...
+        annotation=abi.AssetTransferTransaction,
+        expected=abi.AssetTransferTransactionTypeSpec(),
+    ),
+    TypeAnnotationTest(
+        annotation=abi.ApplicationCallTransaction,
+        expected=abi.ApplicationCallTransactionTypeSpec(),
+    ),
+    TypeAnnotationTest(annotation=abi.Account, expected=abi.AccountTypeSpec()),
+    TypeAnnotationTest(annotation=abi.Asset, expected=abi.AssetTypeSpec()),
+    TypeAnnotationTest(annotation=abi.Application, expected=abi.ApplicationTypeSpec()),
```
jasonpaulos, line 327: The fact that I also changed the format of test cases makes this this hard to see (sorry about that), but these are the only new cases.

The `DynamicBytes` and `StaticBytes` cases were reordered to be closer to the dynamic and static array tests, but are otherwise unchanged.

ahangsu (commented):
starting my pass, a minor note on `CHANGELOG.md` field needed

pyteal/ast/abi/util_test.py
```diff
@@ -10,289 +10,350 @@
...
+        # get_origin will return None for annotations without generic args, e.g. `Byte`
+        annotation_origin = annotation
+
+    assert isinstance(
+        new_instance, annotation_origin
+    ), "TypeSpec.new_instance() returns a value that does not match the annotation type"
+
+    assert (
+        actual == new_instance.type_spec()
+    ), "TypeSpec.new_instance().type_spec() does not match original TypeSpec"
```
ahangsu, line 356: for other pair of eyes: this part is the roundtrip testing for `TypeSpec`, generated new instance should
- hold the same type-spec as the type-spec generates it, 
- and such new instance is an instance of origin class of annotation

ahangsu (approved):
the round trip testing in `test_type_spec_from_annotation` looks good to me

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Unreleased
 
+## Fixed
+* Subroutines annotated with a `TupleX` class are now invoked with an instance of that exact class, instead of the more general `Tuple` class ([#519](https://github.com/algorand/pyteal/pull/519))
```
jasonpaulos, line 4: I've opted to only mention the user-facing effects of this PR

pyteal/ast/abi/util_test.py
jasonpaulos: After making the changes in cb1d8e3f57ca74b64a1d2e5dec9e277a185760c8, I discovered test cases for `String` and `Address` were also missing, so I've added them as well.

michaeldiamant (approved):
@jasonpaulos Great safety improvement - thanks for the effort here!

PR #520, opened 2022-09-01
Use new docstring features in AlgoBank

jasonpaulos:
This is a child PR to #518

It modifies the AlgoBank docstrings to use the new features, but there are still some issues. As you can see from the argument descriptions, multi-line descriptions still have newlines in them. We should probably strip these out as well. Multi-line return descriptions likely suffer from the same problem, but I don't have an example of that here.

Also, I was suspicious of this way of combining the short and long descriptions, so I modified the tests to exploit it and then improved the implementation: https://github.com/algorand/pyteal/blob/553ab34642640280d16aee3da30275f7f0fb68e5/pyteal/ast/subroutine.py#L622-L627

### DIFF
--- a/examples/application/abi/algobank.json
+++ b/examples/application/abi/algobank.json
@@ -6,11 +6,13 @@
             "args": [
                 {
                     "type": "pay",
-                    "name": "payment"
+                    "name": "payment",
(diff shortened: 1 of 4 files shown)


barnjamin (approved):
Thanks!

PR #521, opened 2022-09-01
Minor improvement on `Makefile` to allow local end user always have latest dependencies

ahangsu:
2 things:
- adding `--upgrade` option in `pip`, so local end devs can always use `make setup-development` to upgrade their dependencies
- seems `semantic-version` is a runtime dependency, but it appear again in `requirements.txt`, can be taken out?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 
 setup-development:
 	pip install -e .
-	pip install -r requirements.txt
+	pip install -r requirements.txt --upgrade
 
 setup-docs:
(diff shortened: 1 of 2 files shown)


tzaffi (approved):
LGTM

jasonpaulos (approved):
Changes look good, but I belive this will only update `requirements.txt` dependencies, not those defined in `setup.py`. It would be nice if there were a way to upgrade those as well, but if not, this is still good to have.

PR #523, opened 2022-09-06
v0.18.0

jasonpaulos:
## Added

* ABI Methods will now parse the docstring for the method and set the description for any parameters that are described. ([#518](https://github.com/algorand/pyteal/pull/518))
  * Note: the docstring must adhere to one of google, rst, numpy , or epy formatting styles.

## Fixed
* Subroutines annotated with a `TupleX` class are now invoked with an instance of that exact class, instead of the more general `Tuple` class ([#519](https://github.com/algorand/pyteal/pull/519))

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+# 0.18.0
+
 ## Added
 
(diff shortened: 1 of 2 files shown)


ahangsu (approved):
auto approval (not anonymous)

PR #524, opened 2022-09-07
fix subroutine args with no doc

barnjamin:


jasonpaulos (approved):
Thanks for catching this.

As a nit, could you add a changelog entry?

pyteal/ast/subroutine_test.py
```diff
@@ -1419,6 +1419,12 @@ def documented_method(a: pt.abi.Uint64, *, output: pt.abi.Uint64):
     assert mspec_dict["args"][0]["desc"] == a_doc
     assert mspec_dict["returns"]["desc"] == return_doc
 
+    # No doc
+    documented_method.__doc__ = None
+    mspec_dict = ABIReturnSubroutine(documented_method).method_spec().dictify()
+    assert "descr" not in mspec_dict
```
jasonpaulos, line 1425:
```suggestion
    assert "desc" not in mspec_dict
```

ahangsu (approved):

jasonpaulos (approved):

PR #533, opened 2022-09-13
An attempt to share scratch slot assignments among subroutines

jasonpaulos:
This PR sets up a pipeline for sharing scratch slot assignments among subroutines when possible, and provides a placeholder greedy algorithm to determine which subroutines should share scratch slots.

Very little testing and analysis has been done, so use this at your own risk.

Notably, this code no longer honors user-defined scratch slot locations. That can probably be fixed, it was just too much to think about what I was prototyping this.

### DIFF
--- a/pyteal/compiler/subroutines_test.py
+++ b/pyteal/compiler/subroutines_test.py
@@ -5,21 +5,138 @@ import pytest
 import pyteal as pt
 
 from pyteal.compiler.subroutines import (
-    findRecursionPoints,
+    find_callstack_exclusive_subroutines,
+    find_recursion_points,
     spillLocalSlotsDuringRecursion,
     resolveSubroutines,
 )
 
 
-def test_findRecursionPoints_empty():
+def test_find_callstack_exclusive_subroutines_empty():
     subroutines = dict()
 
     expected = dict()
-    actual = findRecursionPoints(subroutines)
+    actual = find_callstack_exclusive_subroutines(subroutines)
     assert actual == expected
 
 
-def test_findRecursionPoints_none():
+def test_find_callstack_exclusive_subroutines_none():
+    def sub1Impl():
+        return None
+
+    def sub2Impl(a1):
+        return None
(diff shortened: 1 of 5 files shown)


pyteal/compiler/subroutines_test.py
```diff
@@ -5,21 +5,138 @@
...
+        subroutine5: set(),
+    }
+
+    expected = {
+        subroutine1: set(),
+        subroutine2: {subroutine4},
+        subroutine3: {subroutine4, subroutine5},
+        subroutine4: {subroutine2, subroutine3, subroutine5},
+        subroutine5: {subroutine3, subroutine4},
+    }
```
tzaffi, line 89: This approach looks promising though I haven't dug into the details.

Summarizing this test example as a graph (represented by a file-system-like tree hierarchy):

```
s1 -> s2 -> s3
               -> s5
     -> s4
```
and suppose that each subroutine requires exacty 10 slots. Then if my understanding is correct, you end up with the final assignments:
```
s1: 20-29
s2: 10-19
s3: 0-9
s4: 0-9
s5: 0-9
```
This is the same slot assignment as I get from the theoretical pseudocode that I've written elsewhere.
Is my understanding correct?
And if so, what assignment would the following graph produce?

```
s1 -> s2 -> s3 -> s4
               -> s5 -> s6
```
The SCC algo ought to produce:
```
s1: 30-39
s2: 20-29
s3: 10-19
s4: 0-9
s5: 10-19
s6: 0-9
```

jasonpaulos: > And if so, what assignment would the following graph produce?
> 
> ```
> s1 -> s2 -> s3 -> s4
>          -> s5 -> s6
> ```

I expect the code here would produce the same result that you indicated above, or this:

```
s1: 30-39
s2: 20-29
s3: 10-19
s4: 0-9
s5: 0-9
s6: 10-19
```

PR #536, opened 2022-09-15
Unit test speed up

### LINKED ISSUE #526: Shorten unit test duration
Currently, we have one unit test which takes significantly more time to run than all others:

```
118.13s call     pyteal/ast/router_test.py::test_method_config
```

@ahangsu believes it can be refactored to take less time.

ahangsu:
Relying on randomness, randomly take some testcases from full permutation to test method config.

Test duration from 100+s to about 3s.

Closes #526 

No CHANGELOG for changes are not reflected to lib usage.

### DIFF
--- a/pyteal/ast/router_test.py
+++ b/pyteal/ast/router_test.py
@@ -1,4 +1,5 @@
 import pyteal as pt
+import secrets
 from pyteal.ast.router import ASTBuilder
 import pytest
 import typing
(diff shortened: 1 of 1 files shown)


barnjamin:
@ahangsu is this good to go? I'd love to get it merged in if so :patpray:

barnjamin (approved):

pyteal/ast/router_test.py
```diff
@@ -320,12 +355,10 @@ def test_method_config():
         for oc in ON_COMPLETE_CASES
         if str(oc) != str(pt.OnComplete.ClearState)
     ]
+    focg = FullOrderCombinationGen(list(pt.CallConfig), len(ON_COMPLETE_CASES))
```
michaeldiamant, line 358: @ahangsu I opened an _optional_ child PR (https://github.com/algorand/pyteal/pull/569) intending to make debugging simpler.  Can you have a look?

Feel welcomed to keep as is.

michaeldiamant (approved):
@ahangsu Thanks for taking the time to vet the issue.  Looks reasonable to me - moving one step closer to property-based testing!

pyteal/ast/router_test.py
ahangsu: merged

PR #539, opened 2022-09-20
Erroring on constructing an odd length hex byte string

### LINKED ISSUE #538: Odd length hex string should not compile
`byte 0x0` is invalid TEAL `1309: encoding/hex: odd length hex string`

therefore `Bytes("base16", "0x0")` should not compile into invalid TEAL

ahangsu:
closes #538

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Unreleased
 
 ## Fixed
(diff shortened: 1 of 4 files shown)


jasonpaulos (commented):
Thanks for working on this, but I think a better solution would be to error if an odd-length hex string is passed in instead of correcting it.

Issue #537, opened 2022-09-19
ExecuteMethodCall complains about expecting `byte[]` and getting `byte[]`


barnjamin:
### Subject of the issue

Without digging more I expect this is a diff between `DynamicBytes` and `DynamicArray[Byte]`

Posting this so we don't forget to fix

PR #531, opened 2022-09-12
Relax the type checking on arguments to subroutine

barnjamin:
Addresses #530

Issue #530, opened 2022-09-12
Relax ABI Subroutine argument type checking


barnjamin:
If a subroutine is declared as 

```py
@Subroutine(TealType.none)
def txn_checks(txn: abi.Transaction):
    # ...
```

and called with 

```py
@router.method
def payme(ptxn: abi.PaymentTransaction):
    return Seq(
         txn_checks(ptxn)
         # ...
```

Pyteal will complain with:

`pyteal.TealInputError: supplied argument pay at index 0 should have ABI typespec txn but got pay`

We should add a check for whether or not the arg_type is a Transaction type and allow more specific Transaction types to be passed here: https://github.com/algorand/pyteal/blob/master/pyteal/ast/subroutine.py#L285

PR #540, opened 2022-09-22
Refined ABI type-check on assignment

ahangsu:
After a few short discussions with @jasonpaulos on 2022/09/21-2022/09/22, we decided to introduce another typecheck mechanism on variable assignment for ABI values.

Such mechanism would benefit following issues and PRs: #530, #531, #537

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Unreleased
 
 ## Added
-* Refined ABI type-check on assignment. ([#540](https://github.com/algorand/pyteal/pull/540))
+* Added `abi.type_spec_is_assignable_to` to check for compatible ABI type assignments. ([#540](https://github.com/algorand/pyteal/pull/540))
 
 ## Fixed
 * Erroring on constructing an odd length hex string. ([#539](https://github.com/algorand/pyteal/pull/539))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 ## Fixed
 * Erroring on constructing an odd length hex string. ([#539](https://github.com/algorand/pyteal/pull/539))
 * Incorrect behavior when overriding a method name ([#550](https://github.com/algorand/pyteal/pull/550))
+* Add missing `abi.NamedTupleTypeSpec` equality override, such that equality holds only when `instance_class` and `value_type_specs` match. ([#540](https://github.com/algorand/pyteal/pull/540))
+* Prohibited instantiating `abi.NamedTuple` from inheriting subclasses of `abi.NamedTuple`, for fields in subclasses are not inherited. ([#540](https://github.com/algorand/pyteal/pull/540))
 
 # 0.18.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+## Added
+* Refined ABI type-check on assignment. ([#540](https://github.com/algorand/pyteal/pull/540))
+
 ## Fixed
 * Erroring on constructing an odd length hex string. ([#539](https://github.com/algorand/pyteal/pull/539))
 
--- a/pyteal/ast/abi/tuple.py
+++ b/pyteal/ast/abi/tuple.py
@@ -559,6 +559,11 @@ class NamedTuple(Tuple):
         if type(self) is NamedTuple:
             raise TealInputError("NamedTuple must be subclassed")
 
+        if NamedTuple not in type(self).__bases__:
+            raise TealInputError(
+                "NamedTuple instances must be constructed by direct inheritance of NamedTuple"
+            )
+
         anns = get_annotations(type(self))
         if not anns:
             raise TealInputError("Expected fields to be declared but found none")
--- a/pyteal/ast/abi/tuple.py
+++ b/pyteal/ast/abi/tuple.py
@@ -15,7 +15,7 @@ from typing import (
 from collections import OrderedDict
 
 from pyteal.types import TealType
-from pyteal.errors import TealInputError
+from pyteal.errors import TealInputError, TealInternalError
 from pyteal.ast.expr import Expr
 from pyteal.ast.seq import Seq
 from pyteal.ast.int import Int
@@ -555,14 +555,17 @@ class NamedTuple(Tuple):
     .. automethod:: __getattr__
     """
 
+    def __init_subclass__(cls) -> None:
+        super().__init_subclass__()
+        if NamedTuple not in cls.__bases__:
+            raise TealInternalError(
+                f"Cannot construct {cls} by inheriting {cls.__bases__}. "
+                f"Must be constructed by direct inheritance from NamedTuple"
+            )
+
     def __init__(self):
         if type(self) is NamedTuple:
             raise TealInputError("NamedTuple must be subclassed")
-
-        if NamedTuple not in type(self).__bases__:
-            raise TealInputError(
-                "NamedTuple instances must be constructed by direct inheritance of NamedTuple"
-            )
 
         anns = get_annotations(type(self))
         if not anns:
--- a/pyteal/ast/abi/tuple.py
+++ b/pyteal/ast/abi/tuple.py
@@ -556,26 +556,12 @@ class NamedTuple(Tuple):
     """
 
     def __init_subclass__(cls) -> None:
-        def __bfs_bases_tree_contain_named_tuple(tt: type):
-            type_queue = [tt]
-            visited_type = set()
-            while type_queue:
-                current_type = type_queue.pop()
-                if current_type == NamedTuple:
-                    return True
-                visited_type.add(current_type)
-                for _t in current_type.__bases__:
-                    if _t in visited_type:
-                        continue
-                    type_queue.append(_t)
-            return False
-
         is_named_tuple_in_bases = False
 
         for base_t in cls.__bases__:
             if base_t == NamedTuple:
                 is_named_tuple_in_bases = True
-            elif __bfs_bases_tree_contain_named_tuple(base_t):
+            elif issubclass(base_t, NamedTuple):
                 raise TealInternalError(
                     f"Cannot construct {cls} by inheriting {cls.__bases__}. "
                     f"Must be constructed by direct inheritance from NamedTuple"
--- a/pyteal/ast/abi/tuple.py
+++ b/pyteal/ast/abi/tuple.py
@@ -552,10 +552,31 @@ class NamedTuple(Tuple):
 
             my_user = User()
 
+    .. automethod:: __init_subclass__
     .. automethod:: __getattr__
     """
 
     def __init_subclass__(cls) -> None:
+        """This method ensures one to construct only directly from `NamedTuple`,
+        rather than inheriting from `NamedTuple`'s inheritance.
+
+        We want to prohibit the following examples:
+
+            .. code-block:: python
+
+                from pyteal import *
+
+                class LegalInheritance(abi.NamedTuple):
+                    a: abi.Field[abi.Uint64]
+
+                # following are bad cases we guard against
+
+                class IllegalInheritance0(LegalInheritance):
+                    a: abi.Field[abi.Uint64]
+
+                class IllegalInheritance1(LegalInheritance, abi.NamedTuple):
+                    a: abi.Field[abi.Uint64]
+        """
         is_named_tuple_in_bases = False
 
         for base_t in cls.__bases__:
--- a/pyteal/ast/abi/tuple.py
+++ b/pyteal/ast/abi/tuple.py
@@ -564,7 +564,7 @@ class NamedTuple(Tuple):
     """
 
     def __init_subclass__(cls) -> None:
-        """This method ensures one to construct only directly from `NamedTuple`,
+        """This method ensures one only constructs directly from `NamedTuple`,
         rather than inheriting from `NamedTuple`'s inheritance.
 
         We want to prohibit the following examples:
--- a/pyteal/ast/abi/tuple.py
+++ b/pyteal/ast/abi/tuple.py
@@ -522,6 +522,13 @@ class NamedTupleTypeSpec(TupleTypeSpec):
         self.instance_class: type["NamedTuple"] = instance_class
         super().__init__(*value_type_specs)
 
+    def __eq__(self, other: object) -> bool:
+        return (
+            isinstance(other, NamedTupleTypeSpec)
+            and self.instance_class == other.instance_class
+            and self.value_type_specs() == other.value_type_specs()
+        )
+
     def annotation_type(self) -> "type[NamedTuple]":
         return self.instance_class
 
--- a/pyteal/ast/abi/tuple.py
+++ b/pyteal/ast/abi/tuple.py
@@ -556,12 +556,38 @@ class NamedTuple(Tuple):
     """
 
     def __init_subclass__(cls) -> None:
-        super().__init_subclass__()
-        if NamedTuple not in cls.__bases__:
+        def __bfs_bases_tree_contain_named_tuple(tt: type):
+            type_queue = [tt]
+            visited_type = set()
+            while type_queue:
+                current_type = type_queue.pop()
+                if current_type == NamedTuple:
+                    return True
+                visited_type.add(current_type)
+                for _t in current_type.__bases__:
+                    if _t in visited_type:
+                        continue
+                    type_queue.append(_t)
+            return False
+
+        is_named_tuple_in_bases = False
+
+        for base_t in cls.__bases__:
+            if base_t == NamedTuple:
+                is_named_tuple_in_bases = True
+            elif __bfs_bases_tree_contain_named_tuple(base_t):
+                raise TealInternalError(
+                    f"Cannot construct {cls} by inheriting {cls.__bases__}. "
+                    f"Must be constructed by direct inheritance from NamedTuple"
+                )
+
+        if not is_named_tuple_in_bases:
             raise TealInternalError(
-                f"Cannot construct {cls} by inheriting {cls.__bases__}. "
-                f"Must be constructed by direct inheritance from NamedTuple"
+                "Unexpected: did not find NamedTuple in __bases__,"
+                "did not find NamedTuple in __bases__ member's __bases__"
             )
+
+        super().__init_subclass__()
 
     def __init__(self):
         if type(self) is NamedTuple:
--- a/pyteal/ast/abi/util.py
+++ b/pyteal/ast/abi/util.py
@@ -551,10 +551,9 @@ def type_spec_is_assignable_to(a: TypeSpec, b: TypeSpec) -> bool:
     )
 
     match a, b:
+        case NamedTupleTypeSpec(), NamedTupleTypeSpec():
+            return a == b
         case TupleTypeSpec(), TupleTypeSpec():
-            match a, b:
-                case NamedTupleTypeSpec(), NamedTupleTypeSpec():
-                    return a == b
             a, b = cast(TupleTypeSpec, a), cast(TupleTypeSpec, b)
             if a.length_static() != b.length_static():
                 return False
--- a/pyteal/ast/abi/util.py
+++ b/pyteal/ast/abi/util.py
@@ -509,6 +509,7 @@ def type_spec_is_assignable_to(a: TypeSpec, b: TypeSpec) -> bool:
         DynamicArrayTypeSpec,
         StringTypeSpec,
         AddressTypeSpec,
+        UintTypeSpec,
     )
 
     match a, b:
@@ -542,6 +543,9 @@ def type_spec_is_assignable_to(a: TypeSpec, b: TypeSpec) -> bool:
                 case DynamicArrayTypeSpec(), DynamicArrayTypeSpec():
                     return True
             return False
+        case UintTypeSpec(), UintTypeSpec():
+            a, b = cast(UintTypeSpec, a), cast(UintTypeSpec, b)
+            return a.size == b.size
 
     if isinstance(a, type(b)):
         return True
--- a/pyteal/ast/abi/util.py
+++ b/pyteal/ast/abi/util.py
@@ -509,6 +509,9 @@ def type_spec_is_assignable_to(a: TypeSpec, b: TypeSpec) -> bool:
         * value of type :code:`a` has identical encoding as value of type :code:`b`
         * type :code:`b` is as general as, or more general than type :code:`a`
 
+    For `abi.NamedTuple`, we allow mutual assigning between `abi.Tuple` and `abi.NamedTuple`.
+    But between `abi.NamedTuple`, we only return true when the type specs are identical, or we cannot compare against generality.
+
     Some examples are illustrated as following:
 
     =========================== =========================== ============= ====================================================================
@@ -538,6 +541,7 @@ def type_spec_is_assignable_to(a: TypeSpec, b: TypeSpec) -> bool:
 
     from pyteal.ast.abi import (
         TupleTypeSpec,
+        NamedTupleTypeSpec,
         ArrayTypeSpec,
         StaticArrayTypeSpec,
         DynamicArrayTypeSpec,
@@ -548,6 +552,9 @@ def type_spec_is_assignable_to(a: TypeSpec, b: TypeSpec) -> bool:
 
     match a, b:
         case TupleTypeSpec(), TupleTypeSpec():
+            match a, b:
+                case NamedTupleTypeSpec(), NamedTupleTypeSpec():
+                    return a == b
             a, b = cast(TupleTypeSpec, a), cast(TupleTypeSpec, b)
             if a.length_static() != b.length_static():
                 return False
--- a/pyteal/ast/abi/util.py
+++ b/pyteal/ast/abi/util.py
@@ -507,6 +507,8 @@ def type_spec_is_assignable(a: TypeSpec, b: TypeSpec) -> bool:
         ArrayTypeSpec,
         StaticArrayTypeSpec,
         DynamicArrayTypeSpec,
+        StringTypeSpec,
+        AddressTypeSpec,
     )
 
     match a, b:
@@ -532,9 +534,20 @@ def type_spec_is_assignable(a: TypeSpec, b: TypeSpec) -> bool:
                     return a.length_static() == b.length_static()
                 case DynamicArrayTypeSpec(), DynamicArrayTypeSpec():
                     return True
-                case _:
-                    return False
-        case ArrayTypeSpec(), _:
+            return False
+        case (ArrayTypeSpec(), _) | (_, ArrayTypeSpec()):
+            if isinstance(b, ArrayTypeSpec):
+                a, b = b, a
+            match a, b:
+                case DynamicArrayTypeSpec(), StringTypeSpec():
+                    a, b = cast(DynamicArrayTypeSpec, a), cast(StringTypeSpec, b)
+                    return a.value_type_spec() == b.value_type_spec()
+                case StaticArrayTypeSpec(), AddressTypeSpec():
+                    a, b = cast(StaticArrayTypeSpec, a), cast(AddressTypeSpec, b)
+                    return (
+                        a.value_type_spec() == b.value_type_spec()
+                        and a.length_static() == b.length_static()
+                    )
             return False
 
     if isinstance(a, type(b)):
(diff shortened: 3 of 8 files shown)


pyteal/ast/abi/util_test.py
```diff
@@ -771,3 +771,79 @@ def test_sdk_type_specs_from_signature(sig_str, sig_args, sig_rets):
...
+
+class NamedTDecl(abi.NamedTuple):
+    a: abi.Field[abi.Uint64]
+    b: abi.Field[
+        abi.Tuple3[abi.PaymentTransaction, abi.Address, abi.StaticBytes[Literal[16]]]
+    ]
+    c: abi.Field[abi.Transaction]
+
+
+TYPE_SPEC_ASSIGNABLE_CASES = [
```
jasonpaulos, line 784: I'd like to see a few more test cases of the existing behavior:
* `String` is not assignable to `DynamicArray[Byte]` or `DynamicBytes`, and the reverse
* `Address` is not assignable to `StaticArray[Byte, 32]` or `StaticBytes[32]`, and the reverse
* `StaticBytes[x]` is not assignable to `StaticArray[Byte, y]`, where `x != y`

pyteal/ast/abi/util.py
```diff
@@ -499,3 +499,47 @@ def type_specs_from_signature(sig: str) -> tuple[list[TypeSpec], Optional[TypeSp
...
+def type_spec_is_assignable(a: TypeSpec, b: TypeSpec) -> bool:
+    from pyteal.ast.abi import (
+        TupleTypeSpec,
+        ArrayTypeSpec,
+        StaticArrayTypeSpec,
+        DynamicArrayTypeSpec,
+    )
+
+    match a, b:
+        case TupleTypeSpec(), TupleTypeSpec():
```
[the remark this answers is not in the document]
ahangsu: For both case 1 and 2, I think we should all them, and I extended the method such that both cases can pass, in both ways of assignments (a, b or b, a).

For case 3, I add a new testcase for assigning between different `namedTuples`. Let me know if I should extend them.

pyteal/ast/abi/util.py
```diff
@@ -499,3 +499,47 @@ def type_specs_from_signature(sig: str) -> tuple[list[TypeSpec], Optional[TypeSp
...
+                    a, b = cast(StaticArrayTypeSpec, a), cast(StaticArrayTypeSpec, b)
+                    return a.length_static() == b.length_static()
+                case DynamicArrayTypeSpec(), DynamicArrayTypeSpec():
+                    return True
+                case _:
+                    return False
+        case ArrayTypeSpec(), _:
+            return False
+
+    if isinstance(a, type(b)):
```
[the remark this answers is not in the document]
ahangsu: This case is taken care of in latest commit.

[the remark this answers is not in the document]
ahangsu: It would be captured in earlier `match`, and should not get into this position of code.

pyteal/ast/abi/util_test.py
ahangsu: These test cases should be added in recent commits.

pyteal/ast/abi/util_test.py
```diff
@@ -771,3 +771,164 @@ def test_sdk_type_specs_from_signature(sig_str, sig_args, sig_rets):
...
+    ),
+    (
+        abi.type_spec_from_annotation(abi.DynamicArray[abi.Uint64]),
+        abi.type_spec_from_annotation(abi.String),
+        False,
+    ),
+]
+
+
+@pytest.mark.parametrize("a, b, expected", TYPE_SPEC_ASSIGNABLE_CASES)
```
michaeldiamant, line 932: Leaving a note to document a verbal with @ahangsu and @jasonpaulos:  We discussed + agreed it'd make sense to enforce a more exhaustive exercising of `TypeSpec` assignability.  

* Rationale:  Without having to rely on a maintainer's memory alone to extend test cases, we want to defend against regressions as new `TypeSepc`s are introduced / modified.
* Caveats:  
  * We did _not_ agree upon an implementation approach.  It's understood that @ahangsu will make a pass and we can revisit.
  * There's some open questions around ease to exhaustively check _constructive_ types (e.g. `TupleTypeSpec`, `ArrayTypeSpec`).  It's understood that the solution may _not_ fully automate assignability checks.
* Related: In case helpful for brainstorming, we recently implemented a safe guard to warn us that new test cases are needed in https://github.com/algorand/pyteal/pull/519/files#diff-a1205be6b51287e45f2be7311682488d51bddf8900c13deb97df282aa090349bR396.  It's _not_ a direct fit to the problem at-hand, but it may provide inspiration.

pyteal/ast/abi/util.py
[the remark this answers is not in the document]
jasonpaulos: Thanks for clarifying. I just want to point out that this is a potential failure point if we were to introduce new container types.

pyteal/ast/abi/util.py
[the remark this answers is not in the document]
jasonpaulos: After more thought, I propose this rule: give types `a` and `b`, a value of type `a` is safe to view as type `b` (i.e. `type_spec_is_assignable(a,b)`) if and only if the following properties are true:
* Types `a` and `b` are encoded identically
* And type `b` is as general or more general than type `a`

So that would imply the following:

| Type `a` | Type `b` | `type_spec_is_assignable(a,b)` | reason |
| -------- | -------- | ------------------------------ | ------ |
| `DynamicArray[Byte]` | `DynamicBytes` | `true` | `DynamicBytes` is as general as `DynamicArray[Byte]` |
| `DynamicBytes` | `DynamicArray[Byte]` | `true` | `DynamicArray[Byte]` is as general as `DynamicBytes` |
| `StaticArray[Byte,N]` | `StaticBytes[N]` | `true` | `StaticBytes[N]` is as general as `StaticArray[Byte,N]` |
| `StaticBytes[N]` | `StaticArray[Byte,N]` | `true` | `StaticArray[Byte,N]` is as general as `StaticBytes[N]` |
| `String` | `DynamicBytes` | `true` | `DynamicBytes` is more general than `String` |
| `DynamicBytes` | `String` | `false` | `String` is more specific than `DynamicBytes` |
| `Address` | `StaticBytes[32]` | `true` | `StaticBytes[32]` is more general than `Address` |
| `StaticBytes[32]` | `Address` | `false` | `Address` is more specific than `StaticBytes[32]` |

Does this make sense?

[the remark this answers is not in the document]
ahangsu: Good good, that makes sense to me. I just made the change to implementation and testcase.

Next up is to have a full coverage over abi object reflected by the discussion on last Friday (2022/09/30).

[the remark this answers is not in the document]
ahangsu: Yes, I will take a pass and try to get it merged into my implementation.

pyteal/ast/abi/util.py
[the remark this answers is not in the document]
michaeldiamant: Open to discussion, but want to leave the note because some time passed + we've had at least one more verbal.

Assuming there's no obvious + easy type check change, I still think it's relatively moderate cost to implement a check like https://github.com/algorand/pyteal/blob/fdf10f7c7dad4741ceec84b07413211a84086210/pyteal/ast/abi/util_test.py#L395-L404

* Since returning `true` when _not_ assignable is likely the more dangerous situation, I can imagine strengthening the analogous check by requiring the expected result to be `false` (unassignable). 
  * In the parlance of https://github.com/algorand/pyteal/pull/548/, it means looking for at least one `UnsafeAssignment` or `UnsafeBidirectional`.
* The proposal feels easier to maintainer from my perspective than a more automated check that attempts to traverse the type system hierarchy.  Obviously, the proposal leaves room for error, but I've not been able to convince myself that a better tradeoff exists.

pyteal/ast/abi/util_test.py
```diff
@@ -772,3 +772,250 @@ def test_sdk_type_specs_from_signature(sig_str, sig_args, sig_rets):
...
+]
+
+
+@pytest.mark.parametrize("tc", UNSAFE_BIDIRECTIONAL_TEST_CASES)
+def test_type_spec_is_assignable_unsafe_bidirectional(tc: UnsafeBidirectional):
+    for ia, a in enumerate(tc.xs):
+        for ib, b in enumerate(tc.xs):
+            if ia == ib:
+                continue
+            assert not abi.type_spec_is_assignable_to(a, b)
```
michaeldiamant, line 1007: @ahangsu Optional:  As an additional sanity check, rather than omitting the identity check, I see no harm in including it.  What do you think?

If you feel the change reduces clarity and/or you prefer as is, feel welcomed to resolve the thread without change.

```suggestion
            if ia == ib:
                assert abi.type_spec_is_assignable_to(a, b)
            else:
                assert not abi.type_spec_is_assignable_to(a, b)
```

pyteal/ast/abi/util_test.py
```diff
@@ -772,3 +772,250 @@ def test_sdk_type_specs_from_signature(sig_str, sig_args, sig_rets):
...
+                abi.type_spec_from_annotation(NamedTDecl),
+            ]
+        ),
+    ]
+    + [
+        SafeBidirectional([spec, spec])
+        for spec in abi.ReferenceTypeSpecs + abi.TransactionTypeSpecs
+    ]
+    + [
+        SafeBidirectional([spec_t(), spec_t()])
```
michaeldiamant, line 844: @ahangsu Request for clarification:  There's a few `SafeBidirectional` test cases that duplicate the value under test.  Examples include the highlighted line + these examples (from up above):
```
        SafeBidirectional([abi.BoolTypeSpec(), abi.BoolTypeSpec()]),
        SafeBidirectional([abi.StringTypeSpec(), abi.StringTypeSpec()]),
        SafeBidirectional([abi.AddressTypeSpec(), abi.AddressTypeSpec()]),
```

Can you help me understand how duplicating the value rather than showing a single value enriches the test?

If there's no reason to support value duplication, then I'd optionally consider changing from list to set.

pyteal/ast/abi/util_test.py
```diff
@@ -772,3 +772,250 @@ def test_sdk_type_specs_from_signature(sig_str, sig_args, sig_rets):
...
+class NamedTComp2(abi.NamedTuple):
+    b1: abi.Field[abi.Address]
+    b0: abi.Field[abi.DynamicBytes]
+
+
+class SafeBidirectional(NamedTuple):
+    xs: list[abi.TypeSpec]
+
+
+SAFE_BIDIRECTIONAL_TEST_CASES: list[SafeBidirectional] = (
```
michaeldiamant, line 804: @ahangsu Testing my understanding:  Should the following test case pass?  

It fails when I tested locally.  If you think it should fail, can you help me understand the intuition?  I think the encoding is equivalent and it feels like an alias to me.

`SafeBidirectional([abi.ByteTypeSpec(), abi.Uint8TypeSpec()]),`

pyteal/ast/abi/util_test.py
```diff
@@ -772,3 +772,250 @@ def test_sdk_type_specs_from_signature(sig_str, sig_args, sig_rets):
...
+        return
+    assert exists_in_safe_assignment(ts)
+
+
+class UnsafeBidirectional(NamedTuple):
+    xs: list[abi.TypeSpec]
+
+
+UNSAFE_BIDIRECTIONAL_TEST_CASES: list[UnsafeBidirectional] = [
+    UnsafeBidirectional(
```
michaeldiamant, line 957: @ahangsu Minor:  I think these uint cases can be added for more coverage.  Are you comfortable bringing them in?

```suggestion
    UnsafeBidirectional(
        [
            abi.Uint8TypeSpec(),
            abi.Uint16TypeSpec(),
            abi.Uint32TypeSpec(),
            abi.Uint64TypeSpec(),
        ]
    ),
    UnsafeBidirectional(
        [
            abi.ByteTypeSpec(),
            abi.Uint16TypeSpec(),
            abi.Uint32TypeSpec(),
            abi.Uint64TypeSpec(),
        ]
    ),
    UnsafeBidirectional(
```

pyteal/ast/abi/util_test.py
```diff
@@ -772,3 +772,250 @@ def test_sdk_type_specs_from_signature(sig_str, sig_args, sig_rets):
...
+            if type(safe_assignment.a) == _ts:
+                return True
+            for t in safe_assignment.bs:
+                if type(t) == _ts:
+                    return True
+        return False
+
+    if isabstract(ts) or all(isabstract(parent) for parent in ts.__bases__):
+        return
+    assert exists_in_safe_assignment(ts)
```
michaeldiamant, line 949: @ahangsu Optional:  What do you think of the following change?

* When reviewing the highlighted logic, I became a little worried that we may later add a new type and it'll _unintentionally_ go through the `return` branch.
* The proposal is to enforce that _absence_ is a deliberate decision by manually maintaining a list of types allowed to be absent.
* If you feel confident the as is approach provides sufficient coverage, then feel welcomed to resolve as is.

```suggestion
    # Abstract types and types without safe assignments should _not_ appear in test cases.
    if isabstract(ts) or ts in [
        abi.AccountTypeSpec,
        abi.ApplicationTypeSpec,
        abi.AssetTypeSpec,
        abi.BoolTypeSpec,
        abi.ByteTypeSpec,
        abi.Uint8TypeSpec,
        abi.Uint16TypeSpec,
        abi.Uint32TypeSpec,
        abi.Uint64TypeSpec,
    ]:
        assert not exists_in_safe_assignment(ts)
    else:
        assert exists_in_safe_assignment(ts)
```

pyteal/ast/abi/util_test.py
```diff
@@ -772,3 +772,250 @@ def test_sdk_type_specs_from_signature(sig_str, sig_args, sig_rets):
...
+    def exists_in_unsafe_bidirectional(_ts: type):
+        for unsafe_bidirectional in UNSAFE_BIDIRECTIONAL_TEST_CASES:
+            for t in unsafe_bidirectional.xs:
+                if type(t) == _ts:
+                    return True
+        return False
+
+    if isabstract(ts):
+        return
+    assert exists_in_unsafe_bidirectional(ts)
```
michaeldiamant, line 1021: @ahangsu Optional:  In a similar vein to prior remarks, the _suggestion_ proposes ensuring test cases are deliberately omitted by inverting the assertion.   Feel welcomed to resolve if you prefer as is.

```suggestion
    if isabstract(ts):
        assert not exists_in_unsafe_bidirectional(ts)
    else:
        assert exists_in_unsafe_bidirectional(ts)
```

michaeldiamant: Additional note - If you take in the suggestion here, an analogous change can be made in `test_type_spec_is_assignable_safe_bidirectional_full_coverage`.

pyteal/ast/abi/util_test.py
ahangsu: This feels like something worth discussing. When we started working on ABI, even before `TypeSpec` was introduced, we did want to make `byte` inherit from `byte`, see #194.

Later that approach did not work out, and in PR #222 we introduced `TypeSpec`, but the inheritance of `byte` from `uint8` was removed, in this commit: https://github.com/algorand/pyteal/pull/222/commits/4fe7dbf1bf24c1e1b30a96291e9f0e308f70f8d0.

My thinking is, we probably should not change the implementation on `uint` side, but add a rule over `uint`, such that 2 typespecs are mutually assignable if they have same byte size.

I would also like to consult @jasonpaulos for his opinion, in case he sees some drawbacks that we failed to see.

pyteal/ast/abi/util_test.py
ahangsu: I don't see the harm of including it, but so long as typespecs are identical, they should be bidirectional assignable. I would step even forward for change
```suggestion
            if ia == ib:
                assert abi.type_spec_is_assignable_to(a, b)
                assert abi.type_spec_is_assignable_to(b, a)
                continue
            assert not abi.type_spec_is_assignable_to(a, b)
```
What do you think?

pyteal/ast/abi/util_test.py
ahangsu: The boundary check makes sense to me, though the abstract class cannot be instantiated, such that they naturally do not exist in test cases, but it may still worth it to add the check. I will add corresponding change in `SAFE_BIDIRECTIONAL` side.

pyteal/ast/abi/util_test.py
```diff
@@ -772,3 +772,250 @@ def test_sdk_type_specs_from_signature(sig_str, sig_args, sig_rets):
...
+def test_type_spec_is_assignable_safe_bidirectional_full_coverage(ts: type):
+    def exists_in_safe_bidirectional(_ts: type):
+        for safe_bidirectional in SAFE_BIDIRECTIONAL_TEST_CASES:
+            for t in safe_bidirectional.xs:
+                if type(t) == _ts:
+                    return True
+        return False
+
+    if isabstract(ts):
+        return
```
ahangsu, line 869:
```suggestion
    if isabstract(ts):
        assert not exists_in_safe_bidirectional(ts)
```

pyteal/ast/abi/util_test.py
```diff
@@ -772,3 +772,250 @@ def test_sdk_type_specs_from_signature(sig_str, sig_args, sig_rets):
...
+    assert exists_in_safe_bidirectional(ts)
```
ahangsu, line 870:
```suggestion
    if isabstract(ts):
        assert not exists_in_safe_bidirectional(ts)
    else:
        assert exists_in_safe_bidirectional(ts)
```

pyteal/ast/abi/util_test.py
ahangsu: sure, that makes sense to me

pyteal/ast/abi/util_test.py
ahangsu: moving to set is making sense, it was mistaken to have duplicate cases.

There are some difficulty in moving to set, for some type spec instances are not hashable. Need to figure out how to get it work.

pyteal/ast/abi/util_test.py
ahangsu: I can take this change, and I will try to explain more on comment on the choice of the set of type specs to skip.

pyteal/ast/abi/util_test.py
ahangsu: I pushed the change to reflect that the `uint8` and `byte` are mutual assignable, let me know if we should roll back.

pyteal/ast/abi/util_test.py
michaeldiamant: Makes sense to me - thanks for bringing in the change.

pyteal/ast/abi/util_test.py
michaeldiamant: @ahangsu While it'd be nice to understand the behavior, I think https://github.com/algorand/pyteal/pull/540/commits/4315dd0d90da2c8a1e428b551ce7c0e160faf8a2 is sufficient for now.  Thanks for addressing the concern - Resolving.

pyteal/ast/abi/util_test.py
michaeldiamant: @ahangsu Thanks for the change in https://github.com/algorand/pyteal/pull/540/commits/3256254417b66dcc8c2a3a353e196207cc8fea71.  It matches my expectation.  

@jasonpaulos Leaving the thread open for you to sanity check - Can you confirm you're on-board?

pyteal/ast/abi/util_test.py
```diff
@@ -772,3 +772,284 @@ def test_sdk_type_specs_from_signature(sig_str, sig_args, sig_rets):
...
+        [abi.type_spec_from_annotation(NamedTDecl)],
+    ),
+    SafeAssignment(
+        abi.AddressTypeSpec(),
+        [abi.StaticArrayTypeSpec(abi.ByteTypeSpec(), 32), abi.StaticBytesTypeSpec(32)],
+    ),
+    SafeAssignment(
+        abi.type_spec_from_annotation(NamedTComp0),
+        [abi.type_spec_from_annotation(NamedTComp1)],
+    ),
```
jasonpaulos, line 924: I'm still uncertain about allowing different `NamedTuple` subclasses to be assignable to each other. We cannot know that one is more general than the other.

Consider this example:

```python
class Point(abi.NamedTuple):
    x: abi.Field[abi.Uint64]
    y: abi.Field[abi.Uint64]

class AccountRecord(abi.NamedTuple):
    algoBalance: abi.Field[abi.Uint64]
    assetBalance: abi.Field[abi.Uint64]
```

While these two `NamedTuples` are assignable under the current implementation, they represent very different pieces of information and it would likely not be safe/intended to allow one to be viewed as the other.

The conservative option at this point would be to disallow different `NamedTuple` subclasses from being assignable to each other.

In terms of the example, I propose that `Point` and `AccountRecord` should be assignable to `abi.Tuple[abi.Uint64, abi.Uint64]`, but not to each other.

Thoughts on this?

jasonpaulos (commented):
Mostly looks good, but I have concerns about `NamedTuples`

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Unreleased
 
+## Added
+* Refined ABI type-check on assignment. ([#540](https://github.com/algorand/pyteal/pull/540))
```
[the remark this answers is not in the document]
ahangsu: Sure, that makes sense. Thanks!

pyteal/ast/abi/util.py
```diff
@@ -499,3 +499,57 @@ def type_specs_from_signature(sig: str) -> tuple[list[TypeSpec], Optional[TypeSp
         return_type = type_spec_from_algosdk(sdk_method.returns.type)
 
     return [type_spec_from_algosdk(arg.type) for arg in sdk_method.args], return_type
+
+
+def type_spec_is_assignable_to(a: TypeSpec, b: TypeSpec) -> bool:
```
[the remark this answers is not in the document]
ahangsu: Pushed something for docstring in https://github.com/algorand/pyteal/pull/540/commits/312e96f74a6308062798d745823bad39e13b3549, does that look good?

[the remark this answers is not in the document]
jasonpaulos: Looks great, thanks!

pyteal/ast/abi/util_test.py
ahangsu: I pushed something in [d12d158](https://github.com/algorand/pyteal/pull/540/commits/d12d158dc04b27884496eb2a8d245d8a552be2e9) as the first step.

My thinking for the thread above:
> Do we want to have assignability between `NamedTuple`?

Only when the `NamedTupleTypeSpec` are *strictly identical*, or we cannot compare the generality.

> Do we still want assignability between `Tuple` and `NamedTuple`?

My preference is still keep the mutual assignability. I see scenarios that one may want to assign some `Tuple` like stuffs into more structured `NamedTuple`.

> Do we want to consider the child classes of `Point` or `AccountRecord`, i.e., child classes of child class of `NamedTuple`?

I think we want to prohibit the patterns. `NamedTuple` instances must be inherited from `NamedTuple`, and only inherited from `NamedTuple`.

Thoughts on this?

jasonpaulos: > I think we want to prohibit the patterns. `NamedTuple` instances must be inherited from `NamedTuple`, and only inherited from `NamedTuple`.

I definitely agree on this point. I don't think we originally considered this case when we implemented `NamedTuple`, but if you were to subclass a subclass of `NamedTuple` today, you would see incorrect behavior.

Consider this example:

```python
class Point2D(abi.NamedTuple):
    """A point in 2D space."""
    x: abi.Field[abi.Uint64]
    y: abi.Field[abi.Uint64]

class Point3D(Point2D):
    """A point in 3D space."""
    z: abi.Field[abi.Uint64]
```

You would expect `Point3D` to inherit the `x` and `y` properties of `Point2D`, plus its own `z` property, but that's not actually what happens. Instead, `Point3D` would only have the `z` property.

This is because the [`get_annotations`](https://github.com/algorand/pyteal/blob/33a4e5e6691ac64f963bdcde036c2a557cf7945d/pyteal/ast/abi/tuple.py#L562) function we use to find all fields on `NamedTuple` classes _only_ looks at the fields declared in that class and ignores all inherited fields.

Due to this incorrect behavior, I'd argue that the fact you can declare and instantiate `Point3D` at all is a bug and should be prevented.

I think the preferred alternative to this example would be embedding instead of inheriting, e.g.:

```python
class Point3D(Point2D):
    """A point in 3D space."""
    point2d: abi.Field[Point2D]
    z: abi.Field[abi.Uint64]
```

Syntactically this is a little less convenient (you have to go through `point2d` to access at `x` and `y`), but it does convey the same idea.

pyteal/ast/abi/tuple.py
```diff
@@ -555,6 +555,14 @@ class User(abi.NamedTuple):
     .. automethod:: __getattr__
     """
 
+    def __init_subclass__(cls) -> None:
+        super().__init_subclass__()
+        if NamedTuple not in cls.__bases__:
```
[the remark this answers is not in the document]
ahangsu: good find, I think I prefer the second route, though `__init_subclass__` give me some hope that the misbehavior can be stopped in some form.

pyteal/ast/abi/tuple.py
```diff
@@ -556,13 +556,39 @@ class User(abi.NamedTuple):
...
+                        continue
+                    type_queue.append(_t)
+            return False
+
+        is_named_tuple_in_bases = False
+
+        for base_t in cls.__bases__:
+            if base_t == NamedTuple:
+                is_named_tuple_in_bases = True
+            elif __bfs_bases_tree_contain_named_tuple(base_t):
```
jasonpaulos, line 578: Can `issubclass` replace `__bfs_bases_tree_contain_named_tuple`?

ahangsu: good find, take bfs down.

jasonpaulos:
I have one final reservation. Right now `NamedTupleTypeSpec` does not override the `__eq__` method, meaning this code currently passes:

```python
class Point(abi.NamedTuple):
    x: abi.Field[abi.Uint64]
    y: abi.Field[abi.Uint64]

class AccountRecord(abi.NamedTuple):
    algoBalance: abi.Field[abi.Uint64]
    assetBalance: abi.Field[abi.Uint64]

p = Point()
ar = AccountRecord()

assert p.type_spec() == ar.type_spec()
assert abi.type_spec_is_assignable_to(p.type_spec(), ar.type_spec())
```

I'd argue having the equals method return true between different `NamedTuple` subclasses and between a `Tuple` and a `NamedTuple` is _not_ what we want.

ahangsu:
oh boy 🤦 just pushed to save the `NamedTupleTypeSpec` equality check

pyteal/ast/abi/tuple.py
```diff
@@ -552,9 +559,50 @@ class User(abi.NamedTuple):
 
             my_user = User()
 
+    .. automethod:: __init_subclass__
     .. automethod:: __getattr__
     """
 
+    def __init_subclass__(cls) -> None:
+        """This method ensures one to construct only directly from `NamedTuple`,
```
michaeldiamant, line 567: @ahangsu Minor and optional - It seems the wording can be clarified.  I don't feel strongly.

```suggestion
        """This method ensures one only constructs directly from `NamedTuple`,
```

ahangsu: 👍

michaeldiamant (approved):
@ahangsu Looks correct - thanks for addressing the added feedback.

Consider these approval notes:
* I optionally suggest holding for a 2nd approval.
* Once this PR is merged, I expect https://github.com/algorand/pyteal/pull/531 will be updated to leverage `type_spec_is_assignable_to` _prior to_ issuing a PyTeal release.

jasonpaulos (commented):
This looks good! I only left a small comment about calling out some of the recent changes in the changelog

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Unreleased
 
+## Added
+* Added `abi.type_spec_is_assignable_to` to check for compatible ABI type assignments. ([#540](https://github.com/algorand/pyteal/pull/540))
+
 ## Fixed
```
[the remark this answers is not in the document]
ahangsu: pushed the change

ahangsu:
I just dump everything to master, and get #531 merged ASAP.

jasonpaulos (approved):

PR #548, opened 2022-10-04
Add illustrative testing DSL for type_spec_is_assignable_to

michaeldiamant:
Extends #540 with an optional idea that _may_ simplify test case definition for `type_spec_is_assignable_to`.

The intent is to provide a more expressive way to define test cases, which _hopefully_ simplifies test maintainer comprehension and improves test coverage.

The PR is presented assuming there'll be discussion + modification prior to being accepted (if at all).

### DIFF
--- a/pyteal/ast/abi/util_test.py
+++ b/pyteal/ast/abi/util_test.py
@@ -1,6 +1,5 @@
 from typing import cast, Callable, NamedTuple, Literal, Optional, Any, get_origin
 from inspect import isabstract
-from dataclasses import dataclass
 import pytest
 
 import algosdk.abi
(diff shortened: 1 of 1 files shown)


ahangsu (commented):
consider using `NamedTuple` from `typing`, to have less imports

pyteal/ast/abi/util_test.py
```diff
@@ -932,3 +983,30 @@ class NamedTComp2(abi.NamedTuple):
...
+        case SafeAssignment():
+            z = cast(SafeAssignment, tc)
+            for b in z.bs:
+                assert abi.type_spec_is_assignable_to(z.a, b)
+                assert abi.type_spec_is_assignable_to(b, z.a) is False
+
+        case UnsafeAssignment():
+            z = cast(UnsafeAssignment, tc)
+            for b in z.bs:
+                assert abi.type_spec_is_assignable_to(z.a, b) is False
```
ahangsu, line 1012:
```suggestion
                assert abi.type_spec_is_assignable_to(z.a, b) is False

            case _:
                assert False, f"Unknown testcase group {type(tc)}."
```

ahangsu: boundary guard for unknown testcase, in case we add new testcase group later.

ahangsu (approved):
I see this to be a (set of) substitution for `type_spec_assignable` test, and I am thinking of rewriting my previous bundle of test case into your way of categorizing them.

As of the testcases now, I think it looks good enough for me to merge in.

Issue #545, opened 2022-09-29
Allow `OpUp` custom fees


cusma:
## Problem

Currently the `OpUp` utility creates InnerTxns with minimum fee as default, consuming contract funds in case of insufficient pooled fees.

It would be useful to have an explicit option to make sure that InnerTxns fees for the budget top-up MUST be paid by the upper call.

## Solution

I don't know if would be better to set it on `OpUp` instantiation like: 
```python
op_up = OpUp(mode=OpUpMode.OnCall, inner_fee=0)
```

rather than on `OpUp` usage like:
```python
op_up.maximize_budget(fee=budget, inner_fee=0)
```

## Dependencies

None.

## Urgency

Not urgent.

cusma:
@michaeldiamant just opened #553 with a proposal.

PR #553, opened 2022-10-06
`OpUp` inner fee setting and tests.

cusma:
This is a little PR with a proposal to address https://github.com/algorand/pyteal/issues/545

The intuition is that _unless explicitly specified_ `OpUp` assumes that Inner Transactions fee for the budget top-up operations are backed by the upper call.

### DIFF
--- a/pyteal/ast/opup.py
+++ b/pyteal/ast/opup.py
@@ -58,9 +58,7 @@ class OpUp:
             )
     """
 
-    def __init__(
-        self, mode: OpUpMode, target_app_id: Expr = None, inner_fee: Expr = Int(0)
-    ):
+    def __init__(self, mode: OpUpMode, target_app_id: Expr = None):
         """Create a new OpUp object.
 
         Args:
@@ -69,11 +67,6 @@ class OpUp:
             target_app_id (optional): In Explicit mode, the OpUp utility
                 requires the app_id to target for inner app calls. Defaults
                 to None.
-            inner_fee (optional): OpUp utility requires inner
-                transactions fee to be paid by the upper call as default. To
-                let the application pay inner transactions fee with app
-                account's funds, set this parameter to `Global.min_txn_fee`
-                or a given fixed `Int`.
         """
 
         # With only OnCall and Explicit modes supported, the mode argument
@@ -94,16 +87,15 @@ class OpUp:
 
         self.mode = mode
 
+    def _construct_itxn(self, inner_fee: Expr) -> Expr:
         require_type(inner_fee, TealType.uint64)
-        self.inner_fee = inner_fee
 
-    def _construct_itxn(self) -> Expr:
         if self.mode == OpUpMode.Explicit:
             return InnerTxnBuilder.Execute(
                 {
                     TxnField.type_enum: TxnType.ApplicationCall,
                     TxnField.application_id: self.target_app_id,
-                    TxnField.fee: self.inner_fee,
+                    TxnField.fee: inner_fee,
                 }
             )
         else:
@@ -113,12 +105,21 @@ class OpUp:
                     TxnField.on_completion: OnComplete.DeleteApplication,
                     TxnField.approval_program: ON_CALL_APP,
                     TxnField.clear_state_program: ON_CALL_APP,
-                    TxnField.fee: self.inner_fee,
+                    TxnField.fee: inner_fee,
                 }
             )
 
-    def ensure_budget(self, required_budget: Expr) -> Expr:
+    def ensure_budget(self, required_budget: Expr, inner_fee: Expr = Int(0)) -> Expr:
         """Ensure that the budget will be at least the required_budget.
+
+        Args:
+            required_budget: minimum op-code budget to ensure for the
+                upcoming execution.
+            inner_fee (optional): OpUp utility requires inner transactions fee
+                to be paid by the upper call as default. To let the
+                application pay inner transactions fee with app account's
+                funds, set this parameter to `Global.min_txn_fee` or a given
+                fixed `Int`.
 
         Note: the available budget just prior to calling ensure_budget() must be
         high enough to execute the budget increase code. The exact budget required
@@ -137,12 +138,20 @@ class OpUp:
         return Seq(
             buffered_budget.store(required_budget + buffer),
             While(buffered_budget.load() > Global.opcode_budget()).Do(
-                self._construct_itxn()
+                self._construct_itxn(inner_fee=inner_fee)
             ),
         )
 
-    def maximize_budget(self, fee: Expr) -> Expr:
+    def maximize_budget(self, fee: Expr, inner_fee: Expr = Int(0)) -> Expr:
         """Maximize the available opcode budget without spending more than the given fee.
+
+        Args:
+            fee: fee expenditure cap for the op-code budget maximization.
+            inner_fee (optional): OpUp utility requires inner transactions fee
+                to be paid by the upper call as default. To let the
+                application pay inner transactions fee with app account's
+                funds, set this parameter to `Global.min_txn_fee` or a given
+                fixed `Int`.
 
         Note: the available budget just prior to calling maximize_budget() must be
         high enough to execute the budget increase code. The exact budget required
@@ -154,7 +163,7 @@ class OpUp:
         i = ScratchVar(TealType.uint64)
         n = fee / Global.min_txn_fee()
         return For(i.store(Int(0)), i.load() < n, i.store(i.load() + Int(1))).Do(
-            self._construct_itxn()
+            self._construct_itxn(inner_fee=inner_fee)
         )
 
 
--- a/pyteal/ast/opup.py
+++ b/pyteal/ast/opup.py
@@ -58,7 +58,9 @@ class OpUp:
             )
     """
 
-    def __init__(self, mode: OpUpMode, target_app_id: Expr = None):
+    def __init__(
+        self, mode: OpUpMode, target_app_id: Expr = None, inner_fee: Expr = Int(0)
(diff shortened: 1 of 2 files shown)


pyteal/ast/opup.py
```diff
@@ -87,30 +94,27 @@ def __init__(self, mode: OpUpMode, target_app_id: Expr = None):
...
-                ),
-                InnerTxnBuilder.Submit(),
+            return InnerTxnBuilder.Execute(
+                {
+                    TxnField.type_enum: TxnType.ApplicationCall,
+                    TxnField.on_completion: OnComplete.DeleteApplication,
+                    TxnField.approval_program: ON_CALL_APP,
+                    TxnField.clear_state_program: ON_CALL_APP,
+                    TxnField.fee: self.inner_fee,
+                }
```
michaeldiamant, line 117: @cusma Opening a thread to respond to https://github.com/algorand/pyteal/pull/553#issuecomment-1269794621 so that we avoid cluttering _Conversation_ tab.  Can you let me know if you agree with the assessment?  If so, we'll resolve the thread and consider the concern addressed.

* I think you raise a good question.  
* I agree the parameter name can be confusing in the conversation's context.  Though I advocate leaving as is due to the backwards compatibility concern.

tzaffi: 👍 for backward compatability

cusma: Let's preserve backward compatibility then. I couldn't think about a better name than `inner_fee` for the Inner Transactions fee generated by `OpUp`. Open to better suggestions if any.

pyteal/ast/opup.py
```diff
@@ -58,7 +58,9 @@ class OpUp:
             )
     """
 
-    def __init__(self, mode: OpUpMode, target_app_id: Expr = None):
+    def __init__(
+        self, mode: OpUpMode, target_app_id: Expr = None, inner_fee: Expr = Int(0)
```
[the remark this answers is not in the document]
cusma: I initially chose `inner_fee` as constructor without a real strong opinion but I agree that pattern you described is the most common.

Also, since an forgetful re-use of a previously constructed `OpUp` with certain `inner_fee` could expose the Smart Contract funds, I think that having it as method parameter forces more explicit usage, which I think is what we should optimize for looking at security.

So, I'll switch `inner_fee` from constructor to method parameter.

pyteal/ast/opup.py
michaeldiamant: @cusma Resolving - Thanks for the discussion here RE:  backwards compatibility.

pyteal/ast/opup.py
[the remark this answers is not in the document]
michaeldiamant: @cusma https://github.com/algorand/pyteal/pull/553/commits/6b054a17957a22668ccc760bc3a4690b4f478d1f works for me.  

I'd like 1 more reviewer to chime in before considering the thread _resolved_.

tzaffi (commented):
~LGTM~ @michaeldiamant pointed out a question to me that I'm not sure about either, so I'm removing my previous approval.

tzaffi (approved):

pyteal/ast/opup.py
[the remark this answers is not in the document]
tzaffi: Speaking to @michaeldiamant and reading the code, I support this approach of requiring `inner_fee` in every invocation.

pyteal/ast/opup.py
```diff
@@ -133,13 +138,21 @@ def ensure_budget(self, required_budget: Expr) -> Expr:
         return Seq(
             buffered_budget.store(required_budget + buffer),
             While(buffered_budget.load() > Global.opcode_budget()).Do(
-                self._construct_itxn()
+                self._construct_itxn(inner_fee=inner_fee)
             ),
         )
 
-    def maximize_budget(self, fee: Expr) -> Expr:
+    def maximize_budget(self, fee: Expr, inner_fee: Expr = Int(0)) -> Expr:
```
michaeldiamant, line 145: @cusma + reviewers:  I see a backwards _incompatible_ change here and I'm opening the thread to solicit feedback on how to address.

* I _think_  the PR is backwards _incompatible_ due to its default behavior change:  inner transaction fee deducted from caller's account rather than app's account.
* If we agree the change _is_ backwards incompatible, then I think we must rework the PR.  I'm open to ideas and I can elaborate as is helpful.  Here's a few brainstormed thoughts (my preference **bolded**):
  * Add uniquely named methods (e.g. `maximimize_budget_with_caller_account`).  Implies deprecating the existing methods in favor of new, uniquely named methods.
  * **Define a class structure representing possible fee coverage scenarios.**  Modify existing methods to default to the type that matches existing behavior.
* Regardless of implementation approach taken, I think we need to agree on the cases to support.  Here's the cases I see:
  * Paid by caller:  `inner_fee` = 0.
  * Paid by app until app account exhausted:  `inner_fee` _not_ set (as is behavior).
  * Paid by app until max fee reached:  `inner_fee` set with value > 0.

cusma: @michaeldiamant I preferred to have `inner_fee=Int(0)` as default because this automatically preserves App's funds, differently from what happens with the _as-is` behaviour, that could potentially expose the App to unwanted fee draining.

I see your point, let me elaborate on that step by step:

- I agree we slightly change the default behaviour but I'm not sure if this _literally breaks_ the API backward compatibility.
- If we think this _is_ backward incompatible instead, then there could be another option: replace `inner_fee=Int(0)` with `inner_fee=Global.min_txn_fee()`: in this case we preserve the exact same existing behaviour but we loose the "App fund usage" as default.
- On your last point:
    - `inner_fee` = 0: I agree this implies all Inner Transactions fee are paid by the caller, either for `OpUp.ensure_budget` or `OpUp.maximize_budget`;
    - `inner_fee` not set: this implies Inner Transactions fee are "paid by app until app account exhausted" _just_ for `OpUp.ensure_budget`, since `OpUp.maximize_budget` always requires a `fee` parameter (again, not very straightforward naming) to cap the fee consumptions. The point is that we still have to decide "how much" the `OpUp` operations cost: I suppose the answer is `min_txn_fee` as default.
    - `inner_fee` > 0: this implies Inner Transactions fee are "paid by app until max fee reached" _just_ for `OpUp.maximize_budget`.

Now my question is: for the `OpUp` operations, should we enable Inner Transactions fee to be paid by the caller _just_ for the `OpUp.maximize_budget` option? Under `OpUp.ensure_budget` the Inner Transaction fee could be different depending on the execution path, while with `OpUp.maximize_budget` the caller would always know in advance how much fee must be paid with the App call. In the `OpUp.ensure_budget` case, due to execution path dependency, it could be difficult (although possible) for a client to correctly set a sufficient App call fee.

michaeldiamant: @cusma I'm responding inline below.  For now, I'm putting aside implementation concerns to see if we can agree on prerequisite topics.

>  I'm not sure if this literally breaks the API backward compatibility.

* I'd like to agree with you, but the proposal silently modifies the default behavior for an already released API.
* I understand and agree with why the PR opts for a different default.  Unfortunately, doing so is a breaking change.
* _Possible_ middle ground provided reviewers agree:  Deprecate existing methods in favor of new methods with a safer default or all required parameters.   

> replace inner_fee=Int(0) with inner_fee=Global.min_txn_fee(): in this case we preserve the exact same existing behaviour

Unfortunately, the proposal does _not_ preserve existing behavior when the caller overpays.

> Under OpUp.ensure_budget the Inner Transaction fee could be different depending

Does the following rationale convince you that both OpUp methods ought to support payor configuration?  If not, can you add more context to help me see the intuition?
* I agree with the quoted statement though I think it's orthogonal to (what I believe to be) the central question:  How can a smart contract writer configure the payor configuration appropriate for his/her smart contract?
* Ultimately, _some_ account(s) must pay the fees.  Since `OpUp` is a general purpose mechanism supporting, I think it's appropriate for all methods issuing inner-transactions to support payor configuration.

cusma: @michaeldiamant my answers:

> Possible middle ground provided reviewers agree: Deprecate existing methods in favor of new methods with a safer default or all required parameters.

Yes, this could be a way. Once we define the requirements we can keep this option in mind.

> Unfortunately, the proposal does not preserve existing behavior when the caller overpays.

Right, I neglected this corner case due to my bias towards "prioritizing caller payment should be preferable", but you are right in saying that this is not the _exact_ same behavior. So we can discard the `inner_fee=Global.min_txn_fee()` option.

> Ultimately, some account(s) must pay the fees. Since `OpUp` is a general purpose mechanism supporting, I think it's appropriate for all methods issuing inner-transactions to support payor configuration.

Agree, I was maybe overthinking on the "client side", but staying strictly on the "smart contract writer" one I would say that would be preferable to support inner-transaction payor configuration in any case.

cusma: @michaeldiamant thanks!

Due to a new little member of the family 🐣🍼 I have to wind down my commitment on keeping this PR going forward, but I'll try to  follow the discussion and implementation progress every now and then.

jasonpaulos: I just want to chime in and say my personal preference for avoiding a breaking change is to make the `inner_fee` parameter default to `None`, which would indicate `TxnField.fee` should not be set on the inner txn (the same as the current behavior in master). Only by passing in a non-`None` value would the user receive code that explicitly sets the fee.

barnjamin: I implemented the None default param for this in https://github.com/algorand/pyteal/pull/566 but since the pr is on the `cusma` repo I don't think I can target it easily.

PR #554, opened 2022-10-06
Minor improvement on testcase coverage in util.py inside pyteal.abi

ahangsu:
Extract out bfs iteration logic from `test_type_spec_from_annotation_is_exhaustive` and make it generic.

Rewrite `test_type_spec_from_annotation_is_exhaustive` to make it rely on `pytest.mark.parametrize` for better error msg.

I see this to be potentially helpful for future testing on full coverage of test.


michaeldiamant (approved):
@ahangsu Seems like a good to move to me.  Parameterizing the test cases simplifies reasoning about failure cases.  ☕ 

In addition to a visual inspection, I commented out checks in `type_spec_from_annotation` to fail `test_type_spec_from_annotation_is_exhaustive` as intended.

PR #561, opened 2022-10-13
`ExecuteMethodCall` typecheck

### LINKED ISSUE #537: ExecuteMethodCall complains about expecting `byte[]` and getting `byte[]`

ahangsu:
Resolves #537


michaeldiamant (approved):
@ahangsu Thanks for the fix here - As intended, I confirmed the added test fails in `master` without the PR's source changes.

barnjamin (approved):

Issue #547, opened 2022-09-29
Implement subroutine arguments and return values using frame pointers


michaeldiamant:
## Problem

`ABIReturnSubroutine` method parameters are represented by scratch slots in generated TEAL source.  For larger programs, scratch slot usage is dominated by `ABIReturnSubroutine` and may prevent program compilation.

https://github.com/algorand/go-algorand/pull/4319 + https://github.com/algorand/go-algorand/issues/4304 offer a set of frame pointer opcodes that simplify stack-based management of subroutine arguments.  The story requests replacing scratch slot based subroutine parameters with stack-based analogs via frame pointer opcodes.

PR #562, opened 2022-10-13
Frame pointer usage in subroutine

ahangsu:
## TODO list

- [x] Define a set of `Expr`s for frame operations: `proto`, `frame_dig`.
- [x] Define new `var_n_loaded` for frame pointer.
- [x] Modify subroutine declaration, to allow for prefix in subroutine body (for `proto`)
- [x] Figure out `type_of` and `has_return` depending on `subroutine.get_declaration()` (check `SubroutineFnWrapper`)
- [x] Decide if we want to `get_declaration` to initialize all versions of `subroutine_declaration`
    The thinking is: if we want to initialize all versions of declaration, then the scratch slot assignment will be lead to compiler test and roundtrip compile test failing. Do we want to rewrite all these testcases?
- [x] Check again on output ABI variable logic (I think we need a prefix expr in subroutine declaration)
- [x] Graviton test using frame pointer
- [x] Modify ABI-roundtrip test generated file name, should include version name
- [x] ABI integration test cases for frame pointer

~~Router related code change for frame pointer~~
 Should be solved for ticket later on router specific, but the current `Proto` modification should suffice for stack based type cast.

~~Inspired by `cpp` move function, by invoking part of decoding function, but not storing in scratch slot (or reusing them?).~~
    With the current work in #584, I believe this can be solved in a similar flavor.

~~Question: How to decrease ABI arg scratch slot usage, across subroutine and router, or is it false hope?~~
Not a false hope.

~~How to let algod happy about program version 8?~~
Yea it is happy now

Resolves https://github.com/algorand/pyteal/issues/547


barnjamin:
While working on something else I noticed something I'm a bit confused by.

```py
def test_fp_thing():
    # Confirm an equivalent router definition _without_ `name` overrides produces the same output.
    r2 = pt.Router("test")

    @r2.method
    def square(x: pt.abi.Uint64, *, output: pt.abi.Uint64):
        """handles the deposit where the input is an asset transfer"""
        return output.set(x.get() * x.get())

    ap2, _, _= r2.compile_program(version=pt.compiler.MAX_PROGRAM_VERSION)
    print(ap2)
```

The `square` method compiles to the teal:
```
// square
square_0:
proto 1 1
int 0            // ??
frame_dig -1
frame_dig -1
*
frame_bury 0     // ??
frame_dig 0      // ??
retsub
```

Is the `int 0` necessary? also can we skip the bury/dig at the end?

tests/integration/graviton_test.py
```diff
@@ -163,6 +163,8 @@ def test_stable_teal_generation(subr, mode):
     case_name = subr.name()
     print(f"stable TEAL generation test for {case_name} in mode {mode}")
 
+    # HANG NOTE: I prefer not to modify this test, for it is skipped now on thread-unsafe behavior,
+    # and I would suggest revisiting later after we have satisfied solution for #199.
```
ahangsu, line 167: MINOR NOTE. Once resolved we take this out:
```suggestion
```

tests/integration/abi_roundtrip_test.py
```diff
@@ -209,21 +209,25 @@ def test_pure_compilation(abi_type):
...
+        with open(save_to, "w") as f:
+            f.write(teal)
 
-    save_to = tealdir / filename
-    with open(save_to, "w") as f:
-        f.write(teal)
+        assert_teal_as_expected(save_to, FIXTURES / "roundtrip" / filename)
 
-    assert_teal_as_expected(save_to, FIXTURES / "roundtrip" / filename)
+    compile_and_compare(6)
```
tzaffi, line 229: **_BIG_** 👍

tests/integration/graviton_abi_test.py
```diff
@@ -299,7 +300,8 @@ def conditional_factorial(_factor: pt.abi.Uint64, *, output: pt.abi.Uint64) -> p
 # ---- integration test functions ---- #
 
 
-def test_integer65():
+@pytest.mark.parametrize("version", [6, 8])
```
tzaffi, line 303: NOICE

tzaffi (commented):
Still reading the PR, but leaving some minor suggestions.

tests/abi_roundtrip.py
```diff
@@ -120,13 +120,21 @@ def array_comp_factory(self) -> pt.ABIReturnSubroutine:
...
-        else:
+        if not self.type_spec.is_length_dynamic():  # type: ignore[attr-defined]
+            if self.length:
+                assert self.length == self.type_spec.length_static()  # type: ignore[attr-defined]
+            else:
+                self.length = self.type_spec.length_static()  # type: ignore[attr-defined]
+        elif self.length is None:
             self.length = DEFAULT_DYNAMIC_ARRAY_LENGTH
 
+        # if self.length is not None:
```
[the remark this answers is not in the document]
ahangsu: oh I need to rip that off.

tests/integration/graviton_test.py
```diff
@@ -575,22 +579,30 @@ def square(x):
     x = 9
     args = [x]
 
-    # evaluate the programs
-    app_result = PyTealDryRunExecutor(square, Mode.Application).dryrun(args)
-    lsig_result = PyTealDryRunExecutor(square, Mode.Signature).dryrun(args)
+    def evaluate_and_check(_version: int):
```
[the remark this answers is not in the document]
ahangsu: 👍

pyteal/ast/subroutine.py
```diff
@@ -744,8 +843,221 @@ def __call__(self, fn_implementation: Callable[..., Expr]) -> SubroutineFnWrappe
...
+    def fp_evaluator(cls) -> "SubroutineEval":
+        return cls(SubroutineEval.__var_n_loaded_fp, True)
+
+
+SubroutineEval.__module__ = "pyteal"
+
+
 def evaluate_subroutine(subroutine: SubroutineDefinition) -> SubroutineDeclaration:
     """
+    NOTE: THIS METHOD IS DEPRECATED, USE CLASS `SubroutineEval` INSTEAD FOR SUBROUTINE EVALUATOR
```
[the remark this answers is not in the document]
ahangsu: Sounds reasonable to me, the only reason I am concerned about ripping it off is someone outside potentially using it. But given the fact that, we are not top-level exporting the method, we *should* be fine to rip this method out.

Also good point on rewiring some tests in `subroutine_test.py`.

tests/integration/graviton_test.py
```diff
@@ -639,20 +651,24 @@ def euclid(x, y):
...
 
-    # assert that each result is that same as what Python's math.gcd() computes
-    inspectors = PyTealDryRunExecutor(euclid, Mode.Application).dryrun_on_sequence(
-        inputs
-    )
-    for i, result in enumerate(inspectors):
-        args = inputs[i]
-        assert result.stack_top() == math.gcd(*args), result.report(
-            args, f"failed for {args}"
+    def test_and_report(_verison: int):
```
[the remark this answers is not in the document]
ahangsu: 👍

pyteal/ast/subroutine.py
```diff
@@ -239,10 +337,11 @@ def _validate_annotation(
         )
 
     def get_declaration(self) -> "SubroutineDeclaration":
-        if self.declaration is None:
-            # lazy evaluate subroutine
-            self.declaration = evaluate_subroutine(self)
-        return self.declaration
+        # XXX MARK THIS METHOD TO BE DEPRECATED LATER I GUESS
```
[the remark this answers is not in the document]
ahangsu: off they go

pyteal/ast/subroutine.py
```diff
@@ -17,6 +18,102 @@
...
+                Op.callsub.min_version, FRAME_POINTER_VERSION
+            ): SubroutineEval.normal_evaluator(),
+            range(
+                FRAME_POINTER_VERSION, MAX_PROGRAM_VERSION + 1
+            ): SubroutineEval.fp_evaluator(),
+        }
+        self.has_return: Optional[bool] = None
+        self.type_of: Optional[TealType] = None
+
+    def get_declaration(self) -> "SubroutineDeclaration":
```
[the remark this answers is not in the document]
ahangsu: ripped.

pyteal/ast/subroutine.py
```diff
@@ -1014,12 +1014,12 @@ def __call__(self, subroutine: SubroutineDefinition) -> SubroutineDeclaration:
         proto = self.__proto(subroutine)
 
         args = subroutine.arguments()
-        arg_vars: list[ScratchVar] = []
+        arg_vars: list[tuple[ScratchVar, int]] = []
```
tzaffi, line 1017:
```suggestion
        argvar_n_frameidx: list[tuple[ScratchVar, int]] = []
```
nit... suggest a renaming like this that reminds us that there's now a second coordinate

ahangsu: 👍

pyteal/ast/subroutine.py
```diff
@@ -810,53 +941,146 @@ def var_n_loaded(
...
+        # - subroutine is a normal subroutine, then check subroutine body evaluates to something, rather than none
+        # - subroutine is an ABIReturnSubroutine, the type is void, and its subroutine body type of is always none
+        num_stack_outputs: int = (
+            1
+            if subroutine.has_abi_output
+            else int(subroutine.return_type != TealType.none)
         )
 
-    deferred_expr: Optional[Expr] = None
+        return Proto(subroutine.argument_count(), num_stack_outputs, mem_layout=layout)
```
jasonpaulos, line 1011: Following up from https://github.com/algorand/pyteal/pull/584#discussion_r1022155872, it appears that there _is_ a case where `num_stack_outputs != num_return_allocs`, and that is when there's a non-ABI subroutine that returns a value. Because the return value of such a subroutine is the last thing it places on the stack, it's not easy to make it use a preallocated stack return value in any interesting way.

Note that we have to solve this problem if we want to use non-return local variables in a subroutine like that, since just placing the return value on top of the stack will no longer be good enough. But that's a problem for another day.

pyteal/ast/subroutine_test.py
```diff
@@ -1153,6 +1153,8 @@ def mySubroutine():
             return returnValue
 
         definition = pt.SubroutineDefinition(mySubroutine, returnType)
+        evaluate_subroutine = SubroutineEval.normal_evaluator()
```
jasonpaulos, line 1156: For each of the 4 changes you made to these tests, I would like to see a version of that test with `SubroutineEval.fp_evaluator()`.

tests/integration/graviton_abi_test.py
```diff
@@ -513,10 +518,12 @@ def py_factorial(n):
...
 
 
-def test_conditional_factorial():
+@pytest.mark.parametrize("version", [6, 8])
+def test_conditional_factorial(version: int):
     ptdre = PyTealDryRunExecutor(conditional_factorial, pt.Mode.Application)
     inputs = [(n,) for n in range(20)]
-    inspectors = ptdre.dryrun_on_sequence(inputs)
+
+    inspectors = ptdre.dryrun_on_sequence(inputs, compiler_version=version)  # type: ignore
```
jasonpaulos, line 526: A lot of the changes in this file have `# type: ignore`. I can't see why that's needed

tests/abi_roundtrip.py
```diff
@@ -120,11 +120,12 @@ def array_comp_factory(self) -> pt.ABIReturnSubroutine:
         When the length has not been provided for a dynamic array,
         default to DEFAULT_DYNAMIC_ARRAY_LENGTH
         """
-        if self.length is not None:
-            assert self.type_spec.is_length_dynamic()  # type: ignore[attr-defined]
-        elif not self.type_spec.is_length_dynamic():  # type: ignore[attr-defined]
-            self.length = self.type_spec.length_static()  # type: ignore[attr-defined]
-        else:
+        if not self.type_spec.is_length_dynamic():  # type: ignore[attr-defined]
+            if self.length:
```
jasonpaulos, line 124: prefer `is not None` here, since `length` _could_ be 0

pyteal/compiler/compiler.py
```diff
@@ -209,7 +212,7 @@ def compileSubroutine(
     newSubroutines = referencedSubroutines - subroutine_start_blocks.keys()
     for subroutine in sorted(newSubroutines, key=lambda subroutine: subroutine.id):
         compileSubroutine(
-            subroutine.get_declaration(),
+            subroutine.get_declaration_by_version(options.version),
```
jasonpaulos, line 215: Comment relating to this file in general, not this line: I had a verbal discussion with @michaeldiamant and we thought it would be a good idea to add some way of opting out of frame pointers for v8 programs.

The reason being that we've already shipped a pyteal version capable of generating v8 programs _without_ frame pointers. Because frame pointer usages _does_ affect the runtime behavior of programs in some circumstances, it seems like a good idea to give users the option to opt-out of them, since we have to support both cases anyway.

There are a few ways this could be implemented, but I imagine we could add a boolean `frame_pointers` field to `CompileOptions` which would default to true if the version is 8 or higher, and false otherwise.

jasonpaulos (commented):
For the most part this looks good! Great job of adapting the existing patterns to this new behavior.

pyteal/ast/subroutine.py
```diff
@@ -8,15 +8,111 @@
...
 
 
+class _SubroutineDeclByVersion:
+    def __init__(self, subroutine_def: "SubroutineDefinition") -> None:
+        from pyteal.compiler.compiler import FRAME_POINTER_VERSION, MAX_PROGRAM_VERSION
+
+        self.subroutine: SubroutineDefinition = subroutine_def
+        self.version_map: dict[range, Optional[SubroutineDeclaration]] = {
+            range(Op.callsub.min_version, FRAME_POINTER_VERSION): None,
+            range(FRAME_POINTER_VERSION, MAX_PROGRAM_VERSION + 1): None,
```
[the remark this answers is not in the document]
ahangsu: Just add a testcase for that

pyteal/ast/subroutine.py
```diff
@@ -8,15 +8,111 @@
...
+        has_return, type_of = decl.has_return(), decl.type_of()
+        if not is_pre_existing:
+            self.__erase_version(version)
+        ScratchSlot.nextSlotId = current_slot_id
+        return has_return, type_of
+
+    def __info_prepare(self) -> None:
+        from pyteal.compiler.compiler import FRAME_POINTER_VERSION
+
+        if self.has_return and self.type_of:
```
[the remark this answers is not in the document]
ahangsu: good find, I am checking against `None` now

pyteal/ast/subroutine.py
```diff
@@ -475,10 +572,10 @@ def name(self) -> str:
         return self.subroutine.name()
 
     def type_of(self):
-        return self.subroutine.get_declaration().type_of()
+        return self.subroutine.declarations.versions_type_of()
 
     def has_return(self):
-        return self.subroutine.get_declaration().has_return()
+        return self.subroutine.declarations.versions_type_of()
```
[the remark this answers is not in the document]
ahangsu: fat finger copy-paste, good find

pyteal/ast/subroutine.py
```diff
@@ -789,10 +892,38 @@ def evaluate_subroutine(subroutine: SubroutineDefinition) -> SubroutineDeclarati
...
+    Usage (A) "argumentVars" - Storing pre-placed stack variables with frame pointer:
+        Type 1. (by-value) use None for we can load from stack with FrameDig() from FrameVar abstraction
+        Type 2. (by-reference) ALSO use FrameDig() from FrameVar abstraction to pick up from the stack
+            NOTE: SubroutineCall.__teal__() has placed the _SLOT INDEX_ on the stack so this is stored into the local scratch space
+        Type 3. (ABI) use None for we can load from stack with FrameDig() from FrameVar abstraction
+            NOTE: SubroutineCall.__teal__() has placed the ABI encoding on the stack, so ABI set and get methods are working over stack
+        Type 4. (ABI-output-arg) it is not really used here, but we use FrameDig/Bury to interact with it
+
+    Usage (B) "loadedArgs" - Passing through to an invoked PyTEAL subroutine AST:
+        Type 1. (by-value) use FrameDig() from FrameVar absrraction, to have an Expr that can be compiled in python by the PyTEAL subroutine.
```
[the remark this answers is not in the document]
ahangsu: 👍

tests/abi_roundtrip.py
ahangsu: 👍

pyteal/ast/subroutine_test.py
ahangsu: corresponding tests added.

pyteal/compiler/compiler.py
ahangsu: The suggested change has been pushed in this commit: https://github.com/algorand/pyteal/pull/562/commits/4b274508d3c47e48fa0700642e2d7a25d7a246b2

ahangsu: I think we might want the corresponding integration tests?

CHANGELOG.md
```diff
@@ -1,8 +1,14 @@
 # Unreleased
 
+# Added
+* Added frame pointer support for subroutine arguments, to substitute excessive use of scratch slots. ([#562](https://github.com/algorand/pyteal/pull/562))
```
tzaffi, line 4:
```suggestion
* Added frame pointer support for subroutine arguments, replacing the previous usage of scratch. ([#562](https://github.com/algorand/pyteal/pull/562))
```
Completely optional rewording that removes judgmental words like "excessive"

pyteal/ast/frame.py
```diff
@@ -89,19 +89,15 @@ def __init__(
                 f"be greater than local allocations {len(local_stack_types)}."
             )
 
-        if not all(map(lambda t: t != TealType.none, arg_stack_types)):
+        if not all(
+            map(lambda t: t != TealType.none, arg_stack_types + local_stack_types)
+        ):
```
tzaffi, line 94:
```suggestion
        if any(t == TealType.none for t in arg_stack_types + local_stack_types):
```

CHANGELOG.md
ahangsu: 👍

pyteal/ast/frame.py
ahangsu: 👍

pyteal/ast/subroutine_test.py
```diff
@@ -1138,177 +1141,202 @@ def mySubroutine(a):
...
+        expected, _ = pt.Seq([declaration.body.args[0], return_value]).__teal__(
+            options_v8
         )
 
-        actual, _ = declaration.__teal__(options)
-        options.setSubroutine(None)
-        assert actual == expected
+        actual, _ = declaration.__teal__(options_v8)
+        options_v8.setSubroutine(None)
+        with pt.TealComponent.Context.ExprEqualityContext():
```
jasonpaulos, line 1338: I don't think this is necessary/desired here

pyteal/ast/subroutine_test.py
```diff
@@ -1138,177 +1141,202 @@ def mySubroutine(a):
...
         assert isinstance(declaration.body, pt.Seq)
-        assert len(declaration.body.args) == 11
-
-        for i in range(10):
-            assert isinstance(declaration.body.args[i], pt.ScratchStackStore)
+        assert len(declaration.body.args) == 2
 
-        for i in range(10):
-            assert declaration.body.args[i].slot is argSlots[-i - 1]
+        assert isinstance(declaration.body.args[0], Proto)
```
jasonpaulos, line 1328: We should also check the proto arguments

jasonpaulos (commented):
I only have some test comments

pyteal/compiler/compiler.py
```diff
@@ -46,10 +46,22 @@ def __init__(
...
         self.version = version
         self.optimize = optimize if optimize is not None else OptimizeOptions()
+        self.use_frame_pointer: bool
+
+        if frame_pointers is None:
+            self.use_frame_pointer = self.version >= FRAME_POINTER_VERSION
+        else:
+            if frame_pointers and self.version < FRAME_POINTER_VERSION:
+                raise TealInputError(
+                    f"Try to use frame pointer with an insufficient version {self.version}."
```
jasonpaulos, line 61: I don't think there's a test that triggers this error. Could you add one?

pyteal/ast/subroutine_test.py
ahangsu: just removed the context

pyteal/ast/subroutine_test.py
ahangsu: added a check against proto here

pyteal/compiler/compiler.py
ahangsu: yes, add one test in compiler_test

jasonpaulos (approved):
Looks great!

pyteal/ast/frame.py
tzaffi: FYI - I just tried out OpenAI ChatGPT with this refactoring to see if this could have been automated. Low and behold:

@ahangsu @michaeldiamant @jasonpaulos 

image

pyteal/ast/router.py
```diff
@@ -724,6 +724,7 @@ def compile_program(
         version: int = DEFAULT_TEAL_VERSION,
         assemble_constants: bool = False,
         optimize: OptimizeOptions = None,
+        frame_pointers: Optional[bool] = None,
```
tzaffi, line 727: Apologies if we already discussed and came to a different conclusion regarding how to handle the`frame_pointers` compilation behavior. However, should we consider a slightly different API? As this is a new optimization, the following may make sense. Instead of introducing a new variable for `router`, we can update `OptimizeOptions` and have it keep track of the frame_pointers usage. The update would look something like:

```python
class OptimizeOptions:
    def __init__(
        self, 
        *, 
        maximize: bool = True,
        scratch_slots: Optional[bool] = None,
        frame_pointers: Optional[bool] = None,
    ):
        self._skip_slots: Set[ScratchSlot] = set()

        if maximize:
            other_opts = []
            if scratch_slots is not None:
                other_opts.append("scratch_slots")
            if frame_pointers is not None:
                other_opts.append("frame_pointers")
            if other_opts:
                raise TealCompileError(f"When choosing to maximize OptimizeOptions, other parameters {other_opts} should not be provided.")
            self.scratch_slots = self.frame_pointers = True
            return

        # for backwards compatibility, default to self.scratch_slots == False:
        self.scratch_slots = bool(scratch_slots)

        self.frame_pointers = frame_pointers in [None, True]
```

If this seems like a reasonable way forward, I can open a PR.

ahangsu: moving `frame_pointers` option to `OptimizeOptions` seems reasonable to me, while I have some thoughts on `maximize`.

On a syntax level, I think it would be more preferred to introduce class member to return an `OptimizeOptions` instance from a class method, like `OptimizeOptions.maximize()` or `OptimizeOptions.O3`, which feels better for me on triggering all optimizations, without introducing complication in class object initialization arguments.

Does the argument above sounds reasonable in some sense?

tzaffi: Will prepare a PR.

jasonpaulos: I just wanted to chime in and say I agree with the decision reached here. Please tag me on the PR when it's ready.

tzaffi: FYI - after discussing with @michaeldiamant , I'll make this backwards compatible. We should consider changing the behavior to default to `maximize` optimization for AVM v $\geq$ 9

PR #564, opened 2022-10-15
Minor changes for internal consistency

barnjamin:
Working on something else I noticed we didn't provide direct access to the `num_app_args` Txn pseudo-field 

Currently we can still get this information with `Txn.application_args.length` so maybe this is by design? 

Other changes are:

-  Use the full name of the global field accessors instead of shortening application, the API stays the same when using the methods.
 - `__str__` method of assert should print out each of the conditions instead of the list

### DIFF
--- a/pyteal/ast/global_.py
+++ b/pyteal/ast/global_.py
@@ -19,13 +19,13 @@ class GlobalField(Enum):
     logic_sig_version = (5, "LogicSigVersion", TealType.uint64, 2)
     round = (6, "Round", TealType.uint64, 2)
     latest_timestamp = (7, "LatestTimestamp", TealType.uint64, 2)
-    current_app_id = (8, "CurrentApplicationID", TealType.uint64, 2)
+    current_application_id = (8, "CurrentApplicationID", TealType.uint64, 2)
     creator_address = (9, "CreatorAddress", TealType.bytes, 3)
-    current_app_address = (10, "CurrentApplicationAddress", TealType.bytes, 5)
+    current_application_address = (10, "CurrentApplicationAddress", TealType.bytes, 5)
     group_id = (11, "GroupID", TealType.bytes, 5)
     opcode_budget = (12, "OpcodeBudget", TealType.uint64, 6)
-    caller_app_id = (13, "CallerApplicationID", TealType.uint64, 6)
-    caller_app_address = (14, "CallerApplicationAddress", TealType.bytes, 6)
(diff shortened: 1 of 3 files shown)


pyteal/ast/global_.py
```diff
@@ -19,13 +19,13 @@ class GlobalField(Enum):
     logic_sig_version = (5, "LogicSigVersion", TealType.uint64, 2)
     round = (6, "Round", TealType.uint64, 2)
     latest_timestamp = (7, "LatestTimestamp", TealType.uint64, 2)
-    current_app_id = (8, "CurrentApplicationID", TealType.uint64, 2)
+    current_application_id = (8, "CurrentApplicationID", TealType.uint64, 2)
```
ahangsu, line 22: I don't think we can change this without breaking anything tho... for `GlobalField` is exported in `__init__`, right?

ahangsu (commented):
I looked back to old PRs and found this one, the change in `assert_.py` and `txn.py` lgtm, only one question on breaking changes in `global_.py`.

pyteal/ast/global_.py
ahangsu: I did some experiment, and some conclusion drawn:
- `Enum` can have duplicate fields with different names and same value
- we do can deprecate a field in enum, by following example:
```python
class _DeprecateGlobalFieldMeta(EnumMeta):
    def __getattribute__(self, __name: str) -> Any:
        deprecated_fields_and_reasons: dict[str, str] = {"min_txn_fee": "reason"}
        if __name in deprecated_fields_and_reasons:
            warnings.warn(
                deprecated_fields_and_reasons[__name], DeprecationWarning, stacklevel=2
            )
        return EnumMeta.__getattribute__(self, __name)


class GlobalField(Enum, metaclass=_DeprecateGlobalFieldMeta):
```

so we can push through the change by deprecating old fields and use new fields internally.

bbroder-algo: it is clever and TIL metaclass but I wonder if this particular naming inconsistency is worth the extra mechanism.

ahangsu: The need is to align method names with enum fields:
- `current_application_id`
- `current_application_address`
- `caller_application_id`
- `caller_application_address`

While we can add new fields of these names into enum `GlobalField`, we want to deprecate fields with old names. It is still unclear to me how to deprecate a field in enum, until up to now with metaclass.

PR #566, opened 2022-10-17
OpUp: Allow specification of fee source

barnjamin:


pyteal/ast/opup.py
```diff
@@ -87,35 +87,43 @@ def __init__(self, mode: OpUpMode, target_app_id: Expr = None):
...
-                        TxnField.type_enum: TxnType.ApplicationCall,
-                        TxnField.application_id: self.target_app_id,
-                    }
-                ),
-                InnerTxnBuilder.Submit(),
+            return InnerTxnBuilder.Execute(
+                {
+                    TxnField.type_enum: TxnType.ApplicationCall,
+                    TxnField.application_id: self.target_app_id,
+                    TxnField.fee: inner_fee,
```
jasonpaulos, line 101: The idea is that the inner txn fee should specifically _not_ be set if `inner_fee` is None.

barnjamin: ah ok, will change

pyteal/ast/opup.py
```diff
@@ -87,35 +87,41 @@ def __init__(self, mode: OpUpMode, target_app_id: Expr = None):
...
         """Ensure that the budget will be at least the required_budget.
 
+        Args:
+            required_budget: minimum op-code budget to ensure for the
+                upcoming execution.
+            inner_fee (optional): OpUp utility requires inner transactions fee
+                to be paid by the upper call as default. To let the
+                application pay inner transactions fee with app account's
+                funds, set this parameter to `Global.min_txn_fee` or a given
+                fixed `Int`.
```
jasonpaulos, line 122: I don't agree with this comment anymore. It should outline the behavior that will happen if you omit `inner_fee` (first any excess fee credit from the group is used, then the caller app's account is used if needed), and what will happen if you specify it. For the latter, there are two common cases:
1. (The most common and the reason this PR was created) If you specify zero for the fee, you ensure that the group's fee credit must pay for this inner txn
2. If you specify the min txn fee, you ensure that the app's account is used and the group's fee credit is unmodified

Also, it's very important to mention that _multiple_ inner txns may be created, and the specified fee applies to every one of them individually.

pyteal/ast/opup.py
```diff
@@ -133,13 +139,21 @@ def ensure_budget(self, required_budget: Expr) -> Expr:
         return Seq(
             buffered_budget.store(required_budget + buffer),
             While(buffered_budget.load() > Global.opcode_budget()).Do(
-                self._construct_itxn()
+                self._construct_itxn(inner_fee=inner_fee)
             ),
         )
 
-    def maximize_budget(self, fee: Expr) -> Expr:
+    def maximize_budget(self, fee: Expr, inner_fee: Expr = None) -> Expr:
```
jasonpaulos, line 145: I think the change here needs more thought. We already have the user specify `fee`, so `inner_fee` doesn't make sense.

What I think we really want is a way of specifying _who_ should pay the fee. Here are the options I think make sense (notice these are the same as `ensure_budget` when `inner_fee` is None, 0, and min txn fee):
1. Fee is first taken from the group fee credit, if there is one, then drawn from the caller app's account balance if needed
2. Fee is only taken from the group fee credit -- error if the fee credit is too small
3. Fee is only taken from the caller app's account balance

jasonpaulos (commented):
I left comments about the PR as it is in this review, though I realize you may not have modified every part I commented about

pyteal/ast/opup.py
barnjamin: I think this is done with the latest push, will add docs if it looks reasonable

pyteal/ast/opup.py
```diff
@@ -87,35 +97,49 @@ def __init__(self, mode: OpUpMode, target_app_id: Expr = None):
...
+            fields |= {
+                TxnField.on_completion: OnComplete.DeleteApplication,
+                TxnField.approval_program: ON_CALL_APP,
+                TxnField.clear_state_program: ON_CALL_APP,
+            }
 
-    def ensure_budget(self, required_budget: Expr) -> Expr:
+        return InnerTxnBuilder.Execute(fields)
+
+    def ensure_budget(self, required_budget: Expr, inner_fee: Expr = None) -> Expr:
```
michaeldiamant, line 122: @barnjamin + reviewers:  In the prior PR, @cusma mentioned changing the default behavior for security benefits.  How (if at all) should the PR reflect the security concern?

* The concern is:  The default behavior (no fee specified) charges fees to the app's account when the caller does not overpay.  Such behavior may _unintentionally_ drain the app's account.
* Although the PR is a draft, I raise the note now because the outcome _may_ change PR direction.  There's at least a couple of options:
  * Make no behavior change, and update the method docs to clearly highlight the risk.
  * Agree the concern warrants different behavior.  Deprecate existing methods and offer new methods with a safer default _or_ require all parameters (avoid defaults).
* My 2c:  
  * If we started fresh, I think requiring all parameters is most robust.
  * Since we have an existing API, I _think_ doc changes are sufficient.  Though I'd like for folks more attuned to the problem frequency to chime in.

barnjamin: Right, lets give the tools to make the desired behavior easier and improve docs to make it more clear

barnjamin: Once the code seems right, I'll work on docs

michaeldiamant: @barnjamin Resolving - Thanks for the reply.  I trust your judgment that extending the docs is sufficient.

pyteal/ast/opup.py
```diff
@@ -29,6 +30,15 @@ class OpUpMode(Enum):
     OnCall = 1
 
 
+class OpUpFeePayer(Enum):
+    # Only the excess fee on the outer group should be used (set inner_tx.fee=0)
+    Excess = 0
+    # The app account will cover fees
+    Self = 1
```
michaeldiamant, line 37: @barnjamin A couple of thoughts here.  Treat optionally - If you prefer as is, I can work with the PR as shown.  I can also speak live if helpful.
* Since the enum describes a payer, I'd opt for names that reference each actor like Caller (0), App (1), and PreferCaller (2). 
* The inner transaction fee payer concern exists beyond OpUp.  Perhaps the enum can be defined more generally as an enrichment to `InnerTxnBuilder`.

pyteal/ast/opup.py
michaeldiamant: @barnjamin Checking my understanding:  Is it intentional that `ensure_budget` does _not_ use `OpUpFeePayer`?

I assume `ensure_budget` _should_ reference `OpUpFeePayer`.

pyteal/ast/opup.py
barnjamin: Maybe I should rename to `fee_source` since the outer group excess isn't exactly an actor. Open to name change but I think we should treat it as where to pull fees from rather than trying to name actors. wdyt?

I don't immediately see how the enum might be used in an InnerTxn.  It's useful here because the inner transactions are constructed outside the authors immediate control but during construction of an InnerTxn presumably the author has much more control of how they set the fees?

pyteal/ast/opup.py
barnjamin: That probably makes more sense than what it is currently, I was thinking the change to add `OpUpFeePayer` was meant to disambiguate fee/inner_fee of the `maximize_budget` but since I don't expect the inner_fee would be anything but 0/1000/None we should probably use the same logic as maximize

pyteal/ast/opup.py
barnjamin: I've made changes to the name to avoid `payer` in favor of `source`, happy to chat about this

pyteal/ast/opup.py
```diff
@@ -29,8 +33,23 @@ class OpUpMode(Enum):
...
+class OpUpFeeSource(Enum):
+    # Only the excess fee on the outer group should be used (set inner_tx.fee=0)
+    GroupExcess = 0
+    # The app's account will cover all fees (set inner_tx.fee=Global.min_tx_fee())
+    AppAccount = 1
+    # First the excess will be used, remaining fees will be taken from the app account
+    Any = 2
+
+
+def fee_by_source(source: OpUpFeeSource) -> Optional[Expr]:
```
michaeldiamant, line 45: @barnjamin Optional:  To minimize public API surface area, do you feel comfortable marking `fee_by_source` private by renaming to `_fee_by_source`?

Feel welcomed to resolve as is.

pyteal/ast/opup.py
```diff
@@ -87,41 +106,59 @@ def __init__(self, mode: OpUpMode, target_app_id: Expr = None):
 
         self.mode = mode
 
-    def _construct_itxn(self) -> Expr:
+    def _construct_itxn(self, inner_fee: Expr = None) -> Expr:
```
michaeldiamant, line 109: @barnjamin Minor:  Are you comfortable with the following _suggestion_?

With the _suggestion_, mypy prevents an unintended regression where the caller fails to supply a `inner_fee` argument.

If you strongly prefer as is, then feel welcomed to resolve without change.

```suggestion
    def _construct_itxn(self, inner_fee: Optional[Expr]) -> Expr:
```

pyteal/ast/opup.py
michaeldiamant: @barnjamin Good to resolve once the _group excess_ naming question (below) is responded to.

>  I think we should treat it as where to pull fees from 

* Seems reasonable to me.  Thanks for additional framing context.
* Is _group excess_ a term familiar to most builders?  
  * If so, then let's keep as is.  
  * I don't necessarily have a _good_ alternative.  go-algorand uses the term _fee credit_.  The docs also discuss _fee pooling_.  If one of these terms seems better, then let's take it, otherwise let's keep as is.

> don't immediately see how the enum might be used in an InnerTxn. 

* I'm good to keep as is (specific to OpUp) though I'd like to do a quick verbal to socialize the thought.
* From my perspective, configuring the inner transaction payer applies to _any_ inner transaction construction.  We can discuss it out-of-band to the PR to see if it's applicable + worth pursuing.

pyteal/ast/opup.py
```diff
@@ -87,41 +106,59 @@ def __init__(self, mode: OpUpMode, target_app_id: Expr = None):
...
-    def ensure_budget(self, required_budget: Expr) -> Expr:
+    def ensure_budget(
+        self, required_budget: Expr, fee_source: OpUpFeeSource = OpUpFeeSource.Any
+    ) -> Expr:
         """Ensure that the budget will be at least the required_budget.
 
+        Args:
+            required_budget: minimum op-code budget to ensure for the
+                upcoming execution.
+            inner_fee (optional): controls the behavior of setting the fee for
```
michaeldiamant, line 139: @barnjamin Friendly visual reminder (also for me) to update docs once we finalize naming conventions.

pyteal/ast/opup.py
barnjamin: If we use `credit` elsewhere ill tweak to use the same term, no reason to come up with new ones

pyteal/ast/opup.py
barnjamin: will do

pyteal/ast/opup.py
barnjamin: still need to get back to this but marking as ready to review to get any more feedback on code

pyteal/ast/opup.py
michaeldiamant: Resolving - Thanks for brining in the change.

pyteal/ast/opup.py
michaeldiamant: @barnjamin Resolving - _credit_ works for me.  We can pick up an enriched inner transaction builder another time.

barnjamin:
Adding a colon after the pound for the inline comments allows them to be displayed in the docs:
image

michaeldiamant (approved):
@barnjamin and @cusma Thanks for the effort here to improve OpUp!  ☕

jasonpaulos (commented):
This change looks acceptable to me.

I noticed the existing OpUp tests do not verify the IR of the produced expressions like we normally do, so I opened #574 to address this.

I'm flexible about the order of merges here, but I think it'd be nice to merge that PR into this one before it makes its way into master.

pyteal/ast/opup.py
```diff
@@ -13,6 +14,8 @@
 from pyteal.types import TealType, require_type
 from enum import Enum
 
+ON_CALL_APP = Bytes("base16", "068101")  # v6 program "int 1"
```
barnjamin, line 17: Maybe for a follow up pr but should we make this version take the program version from the compile options?

beaker has a half baked concept of a `Precompile` for things like this https://github.com/algorand-devrel/beaker/blob/boxen/beaker/precompile.py

jasonpaulos (approved):

pyteal/ast/opup.py
jasonpaulos: Currently I don't think there's a benefit to having this inner program version match the outer program version. However if we ever make it so that a new AVM version cannot call into v6 programs, this definitely must be changed.

PR #567, opened 2022-10-17
Applying `type_spec_abi_is_assignable_to` to `ABI` computed type

ahangsu:
lifting strict equality type spec check to relaxed type check

### DIFF
--- a/pyteal/ast/abi/address.py
+++ b/pyteal/ast/abi/address.py
@@ -1,5 +1,5 @@
 from enum import IntEnum
-from typing import Union, Sequence, Literal, cast
+from typing import Sequence, Literal, cast
 from collections.abc import Sequence as CollectionSequence
 
 from pyteal.errors import TealInputError
@@ -59,16 +59,14 @@ class Address(StaticArray[Byte, Literal[AddressLength.Bytes]]):
         """
         return self.stored_value.load()
 
-    def set(
+    def set(  # type: ignore[override]
         self,
-        value: Union[
-            str,
-            bytes,
-            Expr,
-            Sequence[Byte],
-            "Address",
-            ComputedValue["Address"],
-        ],
+        value: str
+        | bytes
+        | Expr
+        | Sequence[Byte]
+        | "Address"
+        | ComputedValue["Address"],
     ):
         """Set the value of this Address to the input value.
 
--- a/pyteal/ast/abi/address.py
+++ b/pyteal/ast/abi/address.py
@@ -93,6 +93,7 @@ class Address(StaticArray[Byte, Literal[AddressLength.Bytes]]):
 
         match value:
             case ComputedValue():
+                # TODO need to see again if we need to change on type system or not, _set_with_computed_type?
                 pts = value.produced_type_spec()
                 if pts == AddressTypeSpec() or pts == StaticArrayTypeSpec(
                     ByteTypeSpec(), AddressLength.Bytes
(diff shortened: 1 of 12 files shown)


pyteal/ast/abi/address.py
```diff
@@ -93,6 +93,7 @@ def set(
 
         match value:
             case ComputedValue():
+                # TODO need to see again if we need to change on type system or not, _set_with_computed_type?
                 pts = value.produced_type_spec()
                 if pts == AddressTypeSpec() or pts == StaticArrayTypeSpec(
                     ByteTypeSpec(), AddressLength.Bytes
```
ahangsu, line 99: question on this part: we let `type_spec_is_assignable_to` to be false when we want to check if `byte[32]` is assignable to `address`, while in this case, we do want to assign `byte[32]` to `address`.

What should be changed:
- the type system?
- or the implementation, say we leave a note here for special corner case?

michaeldiamant: It's a good question.  

My 2c and happy to discuss live: I favor changing the type system because it feels analogous to the bi-directional treatment of `NameTuple` and `Tuple`.

jasonpaulos: I agree the type system to be changed to not allow `Address().set(StaticArrary[Bytes, 32]())` (and the `ComputedValue` variant)

ahangsu: change reflected in [3e776f8](https://github.com/algorand/pyteal/pull/567/commits/3e776f8a2bdf00422fe5dbab865450d7b46ac906)

ahangsu: defer `unsafe_cast` implementation to a PR after this one get merged.

pyteal/ast/abi/address.py
```diff
@@ -58,18 +59,14 @@ def get(self) -> Expr:
         """
         return self.stored_value.load()
 
-    def set(
+    def set(  # type: ignore[override]
```
michaeldiamant, line 62: I'm not sure which is preferable - What do folks think?

Options:
* Keep the PR as is with mypy ignore.
* Re-add removed type parameters with usage of the to-be-added `cast` method.

During our live discussion yesterday, I didn't realize the type error will happen.  It feels to me like we should _not_ ignore the error.  

Another alternative (that's more expansive in scope) is to rework `Address` so that it extends `TypeSpec` and uses composition for value storage/retrieval.

pyteal/ast/abi/address.py
```diff
@@ -93,21 +90,9 @@ def set(
...
-                )
+                return self._set_with_computed_type(value)
             case BaseType():
-                if (
-                    value.type_spec() == AddressTypeSpec()
-                    or value.type_spec()
-                    == StaticArrayTypeSpec(ByteTypeSpec(), AddressLength.Bytes)
-                ):
+                if type_spec_is_assignable_to(value.type_spec(), AddressTypeSpec()):
                     return self.stored_value.store(value.stored_value.load())
```
michaeldiamant, line 97: @ahangsu Unrelated to the line - Based on the PR's changes, the `TealInputError` at the end of the method should remove _StaticArray_ from its message.  

Not requesting an immediate change because there's more pressing threads open.  I leave the question to avoid forgetting the concern prior to merge.

pyteal/ast/abi/address.py
michaeldiamant: @ahangsu Thanks for the changes + `unsafe_cast` in follow on PR - I'm good to resolve. 

I have a new open question, but can move discussion out of this thread on my side.

ahangsu: just a reminder, #580 carries the `unsafe_cast` implementation, we can always rebase to this PR and get all the code together.

The final question is on another thread, i.e., on the type signature of `Address.set`, and the behavior related with mypy.

ahangsu:
closing for #583 opened, and more thinking need to be done, as we realize the scope of the problem is much larger.

PR #568, opened 2022-10-17
Using Will's algod docker image

ahangsu:
Changed docker compose file, and add a small script in `Makefile` to retry querying docker until algod is ready to go.

Definitely not the best shell script, let me know how to improve.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -59,6 +59,12 @@ jobs:
           sudo chmod +x /usr/local/bin/docker-compose
           sudo ln -s /usr/local/bin/docker-compose /usr/bin/docker-compose
           docker-compose --version
+      - name: Upgrade docker compose
+        run: |
+          sudo curl -SL "https://github.com/docker/compose/releases/download/v2.11.2/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
+          sudo chmod +x /usr/local/bin/docker-compose
+          sudo ln -s /usr/local/bin/docker-compose /usr/bin/docker-compose
+          docker-compose --version
       - name: Create sandbox
         run: make sandbox-dev-up
       - name: Install python dependencies
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,12 +55,13 @@ jobs:
             $(lsb_release -cs) stable" | sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
           sudo apt update
           sudo apt -y install docker-ce docker-ce-cli containerd.io
-          sudo curl -L "https://github.com/docker/compose/releases/download/1.29.2/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
+          sudo curl -L "https://github.com/docker/compose/releases/download/v2.11.2/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
           sudo chmod +x /usr/local/bin/docker-compose
           sudo ln -s /usr/local/bin/docker-compose /usr/bin/docker-compose
           docker-compose --version
       - name: Upgrade docker compose
         run: |
+          docker-compose --version
           sudo curl -SL "https://github.com/docker/compose/releases/download/v2.11.2/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -58,8 +58,11 @@ lint-and-test: check-generate-init lint test-unit
 
 # ---- Integration Tests (algod required) ---- #
 
+sandbox-dev-up: SHELL:=/bin/bash
 sandbox-dev-up:
 	docker-compose up -d algod
+	SANDBOX_READY_REGEX="^Last committed block:*" ; \
+	while ! [[ `docker compose exec algod goal node status | head -n 1` =~ $$SANDBOX_READY_REGEX ]] ; do sleep 1 ; done
```
ahangsu, line 65: what these lines do is mainly querying `goal node status` and check first line has form `Last commit block: ...`, and keep retrying until the logic condition is satisfied.

Can it be more elegent?

Makefile
```diff
@@ -58,8 +58,11 @@ lint-and-test: check-generate-init lint test-unit
 
 # ---- Integration Tests (algod required) ---- #
 
+sandbox-dev-up: SHELL:=/bin/bash
 sandbox-dev-up:
 	docker-compose up -d algod
+	SANDBOX_READY_REGEX="^Last committed block:*" ; \
+	while ! [[ `docker compose exec algod goal node status 2>&1 | head -n 1` =~ $$SANDBOX_READY_REGEX ]] ; do echo "waiting algod..." ; sleep 1 ; done
```
ahangsu, line 65: not sure if writing shell in `Makefile` is the best practice, but this snippet of code serve as the guard for image to be prepared, before we start integration test.

barnjamin: Could the `docker-compse up` command block until the algod node is up and running?

ahangsu: We have to use detached mode, to run the container background, right?

barnjamin: There is a `--wait` flag that will wait until the container is up and running and healthy and works in detached mode.

Might have to define healthchecks for it in the `docker-compose.yml` file that does something like we're doing here

ahangsu: good point, I think we can make Makefile succinct, and move wait functionality somewhere else then

.github/workflows/build.yml
```diff
@@ -55,7 +55,14 @@ jobs:
...
           sudo apt update
           sudo apt -y install docker-ce docker-ce-cli containerd.io
-          sudo curl -L "https://github.com/docker/compose/releases/download/1.29.2/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
+          sudo curl -L "https://github.com/docker/compose/releases/download/v2.11.2/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
+          sudo chmod +x /usr/local/bin/docker-compose
+          sudo ln -s /usr/local/bin/docker-compose /usr/bin/docker-compose
+          docker-compose --version
+      - name: Upgrade docker compose
+        run: |
+          docker-compose --version
```
ahangsu, line 64:
```suggestion
```

.github/workflows/build.yml
```diff
@@ -55,7 +55,14 @@ jobs:
...
+          sudo curl -SL "https://github.com/docker/compose/releases/download/v2.11.2/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
```
ahangsu, line 65: this line upgrade CI's docker compose from 1.29.2 to 2.11.2, so I can have `--with` in `docker-compose` command line arg, to activate `healthcheck`.

Makefile
ahangsu: @barnjamin take a pass? I did something in workflow to let `docker-compose` to work with `--with` for health check.

barnjamin (approved):

.github/workflows/build.yml
ahangsu: ok, now we are moving from `docker-compose` to `docker compose`, removing these lines of manual upgrade to 2.11.2

michaeldiamant (approved):
@ahangsu Thanks for your efforts improve our correctness confidence by swapping images! ☕

Makefile
ahangsu: ok we are moving from `docker-compose` to `docker compose`, activating command line arg `--wait` should be fine.

Issue #576, opened 2022-10-21
PyTeal fails to compile ARC-4 methods with more than 15 app arguments and atleast one transaction type argument


S3v3ru5:
### Subject of the issue

Hi. I work at Trail of Bits, and we noticed an issue in the way pyteal handles ARC 004 for function with a large number of parameters: it fails to compile ABI Subroutines that takes more than 15 app type args and at least one transaction type args.

Method args are classified into TransactionType [arc-0004.md#transaction-types](https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0004.md#transaction-types) args and App args. We think the reason of this issue is that in [router.py](https://github.com/algorand/pyteal/blob/59b51dcec93e96b31a751fbb3aab968f886c7341/pyteal/ast/router.py#L357-L442), `arg_vals`, the list of all expected method arguments, is divided into `app_arg_vals` and `txn_arg_vals`:
- `app_arg_vals`: list of app type arguments.
- `txn_args_vals`: list of txn type arguments.

If len(`app_arg_vals`) is greater than `METHOD_ARG_NUM_CUTOFF = 15`, `app_arg_vals[METHOD_ARG_NUM_CUTOFF - 1:]` are `tupled` ([router.py#L382-L389](https://github.com/algorand/pyteal/blob/59b51dcec93e96b31a751fbb3aab968f886c7341/pyteal/ast/router.py#L382-L389)).

Tupled arguments are later de-tupled in ([router.py#L431-L442](https://github.com/algorand/pyteal/blob/59b51dcec93e96b31a751fbb3aab968f886c7341/pyteal/ast/router.py#L431-L442])).

While de-tupling, `arg_vals[METHOD_ARG_NUM_CUTOFF - 1 :]` is considered as the original types which are tupled. However, `app_arg_vals[METHOD_ARG_NUM_CUTOFF - 1 :]` are the original types and `app_arg_vals[METHOD_ARG_NUM_CUTOFF - 1:]` is **not** equal to `arg_vals[METHOD_ARG_NUM_CUTOFF - 1 :]` if there is a transaction type arg in `arg_vals`.

As `set(app_arg_vals) = set(arg_vals) - set(txn_arg_vals)`, `app_arg_vals[METHOD_ARG_NUM_CUTOFF - 1 :]` will have length less than `arg_vals[METHOD_ARG_NUM_CUTOFF - 1 :]`, compilation would fail with `Index out of bounds` error.

### Your environment

[...]

PR #577, opened 2022-10-21
Fix: tupled app args with transaction type

barnjamin:
addresses #576

### DIFF
--- a/pyteal/ast/router.py
+++ b/pyteal/ast/router.py
@@ -361,11 +361,6 @@ class ASTBuilder:
                 ats for ats in arg_vals if not isinstance(ats, abi.Transaction)
             ]
 
-            # only transaction args (these are omitted from app args)
-            txn_arg_vals: list[abi.Transaction] = [
-                ats for ats in arg_vals if isinstance(ats, abi.Transaction)
-            ]
-
             for aav in app_arg_vals:
                 # If we're here we know the top level isnt a Transaction but a transaction may
                 # be included in some collection type like a Tuple or Array, raise error
--- a/pyteal/ast/router.py
+++ b/pyteal/ast/router.py
@@ -383,7 +383,6 @@ class ASTBuilder:
                 last_arg_specs_grouped: list[abi.TypeSpec] = [
                     t.type_spec() for t in app_arg_vals[METHOD_ARG_NUM_CUTOFF - 1 :]
                 ]
-                tupled_app_args = app_arg_vals[METHOD_ARG_NUM_CUTOFF - 1 :]
                 app_arg_vals = app_arg_vals[: METHOD_ARG_NUM_CUTOFF - 1]
                 app_arg_vals.append(
                     abi.TupleTypeSpec(*last_arg_specs_grouped).new_instance()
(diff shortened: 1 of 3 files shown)


tonyfloatersu (commented):
Hey, it's also Hang, please consider add a line in CHANGELOG

pyteal/ast/router.py
```diff
@@ -431,13 +432,10 @@ def wrap_handler(
             # de-tuple into specific values using `store_into` on
             # each element of the tuple'd arguments
             if tuplify:
```
ahangsu, line 434:
```suggestion
            if tuplify:
                tupled_app_args = app_arg_vals[METHOD_ARG_NUM_CUTOFF - 1 :]
```

pyteal/ast/router.py
```diff
@@ -361,6 +361,11 @@ def wrap_handler(
                 ats for ats in arg_vals if not isinstance(ats, abi.Transaction)
             ]
 
+            # only transaction args (these are omitted from app args)
+            txn_arg_vals: list[abi.Transaction] = [
+                ats for ats in arg_vals if isinstance(ats, abi.Transaction)
+            ]
+
```
ahangsu, line 368:
```suggestion
```

pyteal/ast/router.py
```diff
@@ -373,16 +378,12 @@ def wrap_handler(
             # assign to a var here since we modify app_arg_vals later
             tuplify = len(app_arg_vals) > METHOD_ARG_NUM_CUTOFF
```
ahangsu, line 380:
```suggestion
            tuplify = len(app_arg_vals) > METHOD_ARG_NUM_CUTOFF

            # only transaction args (these are omitted from app args)
            txn_arg_vals: list[abi.Transaction] = [
                ats for ats in arg_vals if isinstance(ats, abi.Transaction)
            ]

```

ahangsu (commented):
minor comments to minimize the change, the move at this point seems reasonable to me.

Good to have the issue reproducing code in test case, with transaction being the last arg.

ahangsu:
oh snap something goes wild, looking into it

pyteal/ast/router.py
```diff
@@ -383,6 +383,7 @@ def wrap_handler(
                 last_arg_specs_grouped: list[abi.TypeSpec] = [
                     t.type_spec() for t in app_arg_vals[METHOD_ARG_NUM_CUTOFF - 1 :]
                 ]
+                tupled_app_args = app_arg_vals[METHOD_ARG_NUM_CUTOFF - 1 :]
```
ahangsu, line 386:
```suggestion
                tupled_app_args = app_arg_vals[METHOD_ARG_NUM_CUTOFF - 1 :]
                last_arg_specs_grouped: list[abi.TypeSpec] = [
                    t.type_spec() for t in tupled_app_args
                ]
```

ahangsu (approved):
I think it is in a good shape, also requesting a pass from @jasonpaulos

michaeldiamant (approved):
@barnjamin Thanks for the fix - looks correct to me.  

Can address out-of-band to the PR:  I'd feel more confident if we can generate arbitrary test examples capturing all `BaseType`s.

jasonpaulos (approved):
Looks correct, thanks for fixing this

PR #579, opened 2022-10-21
More strict MethodCall args type checking

barnjamin:


ahangsu (commented):
I think the logic behind `require_exact_type` is roughly the same as `require_type`? What is the motivation of introducing another type checking function?

barnjamin:
@ahangsu i dont remember exactly, but I _think_ it had to do with `anytype` matching for bytes or int.

Closing this though since I cant recall why I filed it

PR #580, opened 2022-10-21
Type cast between ABI values, if they have same encoding underneath

ahangsu:
related with #567, for the `cast` implementation.


ahangsu:
closing for #583 opened, and more thinking need to be done, as we realize the scope of the problem is much larger.

PR #584, opened 2022-10-25
ABI Data Schema

ahangsu:
After discussion with @jasonpaulos in 2022/10/24, we decide to provide a set of data retrieval interface for ABI values, not just scratch slots. This will hopefully decrease (or eliminate) the ABI scratch slot use in subroutine.

~~Based on #562 for now.~~ #562 is based on this one.


JGomezST:
Looking forward to this PR. Scratch slot use currently is very inefficient and I hit the cap with only a 2 page application.

tzaffi (commented):
I haven't yet read the final PR in this trio, so I don't yet see the application of these changes, but they look _very promising_ in promoting code re-usage and better API's.

I made suggestions to consolidate even further by subclassing `ScratchVar` under the new abstract class and keeping the same methods. Again, because I haven't yet read the final PR, I'm not 100% this actually works (but I would be surprised if it doesn't). This would also cut the surface area of changes down drastically.

pyteal/ast/abi/type.py
```diff
@@ -65,6 +65,64 @@ def __str__(self) -> str:
 TypeSpec.__module__ = "pyteal.abi"
 
 
+class DataStorageSchema(ABC):
```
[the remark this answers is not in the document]
tzaffi: If `FrameVar`'s only end up working in the context of a subroutine, maybe they should be called `SubroutineVar`'s or `FrameParam` or `SubroutineParam`...

pyteal/ast/abi/type.py
```diff
@@ -65,6 +65,64 @@ def __str__(self) -> str:
...
+        pass
+
+
+DataStorageSchema.__module__ = "pyteal"
+
+
+class ScratchStorage(DataStorageSchema):
+    def __init__(self, storage_type: TealType) -> None:
+        super().__init__()
+        self.scratchvar: Final = ScratchVar(storage_type)
```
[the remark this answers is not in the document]
ahangsu: good find, I am also applying the fix to the one in type spec decl in `BaseType`

pyteal/ast/abi/type.py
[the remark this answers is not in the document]
ahangsu: The thinking makes sense to me, I would propose `AbstractVar` and `FrameVar` for now.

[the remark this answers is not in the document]
ahangsu: actually this is a good thinking, for we should move the declaration of such storage model out of ABI, for it serves a wider range or generality.

The thinking of renaming it to `FrameVar` is that, it can help in router rewrite work, though we haven't reach yet. Although the work still need the help of subroutine, we use `FrameVar` for it suggests the data source is coming from frame pointer, which is pertinent to me.

pyteal/ast/abi/address.py
```diff
@@ -56,7 +56,7 @@ def get(self) -> Expr:
 
         The expression will have the type TealType.bytes.
         """
-        return self.stored_value.load()
+        return self._stored_value.load()
```
tzaffi, line 59: I agree that `stored_value` should have been private, but unfortunately it wasn't. Making it private is a breaking change. I doubt many if any people are actually using it as opposed to the clearly preferable `get()` methods. However, we should decide how to handle this situation deliberately.

@barnjamin @michaeldiamant @jasonpaulos any thoughts?

tzaffi: We probably could get away with it given our previous warnings.

barnjamin: agree, if we're going to break it we should break it now

pyteal/ast/scratchvar.py
```diff
@@ -1,11 +1,30 @@
...
 from pyteal.errors import TealInputError
-from pyteal.types import TealType, require_type
+from pyteal.types import TealType, require_type, types_match
 
 from pyteal.ast.expr import Expr
 from pyteal.ast.scratch import ScratchSlot, ScratchLoad, ScratchStore
 
 
-class ScratchVar:
+class AbstractVar(ABC):
```
tzaffi, line 10: Nit: since `AbstractVar` may (and in fact _is_) subclassed by non-scratch related classes, I recommend a new file `abstractvar.py` to contain its source.

ahangsu: that's also good point, do we need a set of test for that? I suppose no, for it is pure abstract class.

pyteal/ast/scratchvar.py
```diff
@@ -42,7 +61,12 @@ def store(self, value: Expr) -> Expr:
         Args:
             value: The value to store. Must conform to this ScratchVar's type.
         """
-        require_type(value, self.type)
+        from pyteal.ast.subroutine import SubroutineCall
```
tzaffi, line 64: I'm trying to understand why we need this workaround for Subroutines. Maybe there exists a more straightforward approach?

ahangsu: Let me provide a little more context about the design back when we were working on `ABIReturnSubroutine`.

The `ABIReturnSubroutine` returns the `output` value encoding on the stack, then the `deferred_expr` picks it up and store the encoding into a scratch slot, when we have only scratch slot. Anyways, the stack remains nothing, and surprisingly the `ABIReturnSubroutine` evaluates to nothing, i.e., `TealType.none`.

Now what, what is it related with `BaseType.store_value`? There's some hand waving between `deferred_expr` and `BaseType.store_value`: such that after subroutine call, we initialize an ABI value. The ABI value stores the result of `ABIReturnSubroutine` into itself, which is the execution of `deferred_expr`.

So why it is using `slot` rather than just `store` into `scratchvar`? Recall we discussed that `ABIReturnSubroutine` has type `TealType.none`, we here use `slot` to bypass the type checking, which is not quite good.

Now we are back to this work around, so for subroutine call with abi return, we need to check the type of output, such that we can perform sort of type check against stack

tzaffi: Is there a specific test that you know of that is failing without introducing this code?

Assuming we need this code, I think a slight improvement (in terms of legibility) would be to consider the `else` case first.

```python
        if isinstance(value, SubroutineCall) and value.output_kwarg:
            assert types_match(value.output_kwarg.abi_type.storage_type(), self.type)
        else:
            require_type(value, self.type)
```

ahangsu: I opened PR: https://github.com/algorand/pyteal/pull/588 to illustrate the results of lifting such checks.

Essentially, on all of build test failing cases: they are attempting to store results of `ABIReturnSubroutine` into another ABI value (with underlying scratch var). The result of `ABIReturnSubroutine` should be of type `ReturnedType`, and the `type_of` method is returning `TealType.none`.

If we want to match the `ScratchVar`'s type against `ReturnedValue`'s "wrapped" result type, we need to go into `value.output_kwarg` to see the type of ABI value encodings.

Hopefully there are more succinct ways to deal with the type checking... 🤔

ahangsu: In any case, this piece of code only perform type checking, I should do a better job in raising error message.

pyteal/ast/abi/address.py
ahangsu: I suppose the consensus we reached in the live discussion is that, we want to make `stored_value` hidden as `_stored_value`, and probably it is a good time to have breaking change early?

pyteal/ast/scratchvar.py
tzaffi: I'm investigating and close to prepare a proposed child PR.

tzaffi: Here is the child PR #590. There are pros and cons to it. I am not convinced that it is actually an improvement over the current approach. However, if you feel that it is feel free to merge it in. Or if you see some possible improvements that would get it mergeable, feel free to suggest. However, given the very modest effect on overall code quality, we shouldn't spend much more time in this rabbit hole which I have dug for us.

ahangsu: I am merging in the change, for I think it is a reasonable move for code type check.

pyteal/ast/abi/type.py
```diff
@@ -1,11 +1,11 @@
 from typing import TypeVar, Generic, Callable, Final, cast
 from abc import ABC, abstractmethod
 
-from pyteal.errors import TealInputError
-from pyteal.types import TealType
 from pyteal.ast.expr import Expr
-from pyteal.ast.scratchvar import ScratchVar
+from pyteal.ast.scratchvar import ScratchVar, AbstractVar
```
tzaffi, line 5:
```suggestion
from pyteal.ast.abstractvar import AbstractVar
from pyteal.ast.scratchvar import ScratchVar
```

though you may wish to re-sort the imports in this section

ahangsu: good find, oh boy 🤦

pyteal/ast/frame.py
```diff
@@ -101,6 +102,31 @@ def has_return(self) -> bool:
...
+        self.frame_index = frame_index
+
+    def storage_type(self) -> TealType:
+        return self.stack_type
+
+    def store(self, value: Expr, validate_types: bool = True) -> Expr:
+        if not validate_types:
+            # validation always happens inside of FramBury's initializer
+            raise TealInternalError(
+                f"DynamicScratchVar's must validate_types but {validate_types=}"
```
tzaffi, line 118:
```suggestion
                f"FrameVar's must validate_types but {validate_types=}"
```

woopsies

tzaffi (approved):
I'm going ahead and approving the PR with the understanding that the minor suggestions (import `AbstractVar` from its own file + correct the wrong error message which I recently introduced) are addressed.

pyteal/ast/abi/type.py
ahangsu: There's actually more wrong imports of `AbstractVar`, and I got them fixed.

pyteal/ast/frame.py
```diff
@@ -1,16 +1,18 @@
...
+from pyteal.errors import TealInputError, TealInternalError, verifyProgramVersion
 from pyteal.ir import TealBlock, TealSimpleBlock, TealOp, Op
 
 if TYPE_CHECKING:
     from pyteal.compiler import CompileOptions
 
 
 class Proto(Expr):
-    def __init__(self, num_args: int, num_returns: int):
+    def __init__(self, num_args: int, num_returns: int, /, *, num_allocs: int = 0):
```
tzaffi, line 15:
```suggestion
    def __init__(self, num_args: int, num_returns: int, *, num_allocs: int = 0):
```

In my opinion, it is convenient to continue allowing the naming of arguments `num_args` and `num_returns`. However, there could be a compelling reason to disallow this so feel free to ignore this suggestion.

pyteal/ast/frame.py
```diff
@@ -74,8 +87,16 @@ def has_return(self) -> bool:
 
 class FrameBury(Expr):
     def __init__(self, value: Expr, frame_index: int):
+        from pyteal.ast.subroutine import SubroutineCall
+
         super().__init__()
-        require_type(value, TealType.anytype)
+        if not isinstance(value, SubroutineCall) or not value.output_kwarg:
```
tzaffi, line 93: Nit: this looks like the 2nd time that we're using this logic. So we can get a little more D.R.Y. with a common helper function. However, feel free to ignore the suggestion if, for example, it's too awkward to consolidate the logic across python packages.

pyteal/ast/frame.py
ahangsu: the interface changed by introduction of #593 , resolving

pyteal/ast/frame.py
ahangsu: the implementation changed by introduction of #593 , resolving

pyteal/ast/frame.py
```diff
@@ -73,9 +225,26 @@ def has_return(self) -> bool:
 
 
 class FrameBury(Expr):
-    def __init__(self, value: Expr, frame_index: int):
+    """An expression that burys a value to a position around frame pointer.
```
jasonpaulos, line 228: nit: spelling, burys -> buries

jasonpaulos (commented):
The actual changes seem pretty straightforward, but I do have a few questions about specifics.

pyteal/ast/frame.py
ahangsu: 👍

pyteal/ast/abi/type.py
```diff
@@ -76,8 +77,14 @@ class BaseType(ABC):
...
         super().__init__()
-        self._type_spec: Final = spec
-        self.stored_value: Final = ScratchVar(spec.storage_type())
+        self._type_spec: Final[TypeSpec] = spec
+        self._stored_value: AbstractVar = ScratchVar(spec.storage_type())
+
+    def _load_value(self) -> Expr:
+        return self._stored_value.load()
+
+    def _store_value(self, value: Expr) -> Expr:
```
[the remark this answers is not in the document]
ahangsu: 👍 removed

pyteal/ast/abstractvar.py
```diff
@@ -0,0 +1,34 @@
...
+    * (Strong) Typed: can indicate its own type.
+
+    ScratchVar and FrameVar inherits from this class, representing the load and storage of value
+    against scratch slots or stack based on frame pointer.
+
+    This class is intentionally hidden because it's too basic to directly expose.
+    """
+
+    @abstractmethod
+    def store(self, value: Expr, validate_types: bool = True) -> Expr:
```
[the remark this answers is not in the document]
ahangsu: just add some explanation for motivation of `validate_type`, something similar to PR explanation.

pyteal/ast/frame.py
```diff
@@ -1,27 +1,167 @@
...
-                f"the number of arguments provided to Proto must be >= 0 but {num_args=}"
+                f"The number of arguments provided to Proto must be >= 0 but {num_args=}."
             )
         if num_returns < 0:
             raise TealInputError(
-                f"the number of return values provided to Proto must be >= 0 but {num_returns=}"
+                f"The number of returns provided to Proto must be >= 0 but {num_returns=}."
             )
+        if mem_layout:
+            if mem_layout.num_return_allocs > num_returns:
```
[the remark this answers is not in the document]
ahangsu: `num_return_allocs` are applied for ABI return value, while `num_returns` in `proto` only matters about how many remaining things on the stack. So not all return on stack are ABI values, `num_return_allocs` <= `num_returns`.

pyteal/ast/frame.py
```diff
@@ -1,27 +1,167 @@
...
+    def __init__(
+        self,
+        arg_stack_types: list[TealType],
+        local_stack_types: list[TealType],
+        num_return_allocs: int,
+    ):
+        super().__init__()
+        if num_return_allocs < 0:
+            raise TealInternalError("Return allocation number should be non-negative.")
+        elif num_return_allocs > len(local_stack_types):
```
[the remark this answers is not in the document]
ahangsu: return values are subset of `local_stack_types`, and they are the 0st from frame pointer. I will note the convention in docstrong.

pyteal/ast/abi/type.py
```diff
@@ -231,7 +240,22 @@ def store_into(self, output: BaseType) -> Expr:
...
+
+        if (
+            isinstance(self.computation, SubroutineCall)
+            and self.computation.output_kwarg
+        ):
+            assert types_match(
+                self.computation.output_kwarg.abi_type.storage_type(),
+                output._stored_value.storage_type(),
+            )
+            validate_in_store = False
```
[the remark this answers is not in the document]
ahangsu: What you are point at is the reason that `ABIReturnSubroutine` evaluates to `TealType.none`, but recall the syntax of `ABIReturnSubroutine`, which returns an `Expr` that:
- either has type `void`, which is `TealType.none`
- or does have an output ABI value, but the `Expr` evaluates to `TealType.none`
  - in scratch slot based compilation, we use deferred expr to load to stack
  - in frame based compilation, we do nothing for result is already placed on stack

  and then we wrap the computation with `ReturnedValue`.

In either case, I don't feel right about changing this line...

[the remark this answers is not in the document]
ahangsu: After some more thoughts, I think what determines here is the scratch slot behavior, that takes the encoding off the stack.

I would propose to let return type to be concrete, only when the return value is in the context of frame pointer, rather than being stored in a scratch slot.

pyteal/ast/frame.py
```diff
@@ -1,27 +1,172 @@
...
+                raise TealInternalError(
+                    "Local variable in subroutine initialization must be typed."
+                )
+
+    def __teal__(self, options: "CompileOptions") -> tuple[TealBlock, TealSimpleBlock]:
+        if self.count == 1:
+            inst_srt, inst_end = self.auto_instance.__teal__(options)
+            return inst_srt, inst_end
+        else:
+            dupn_srt, dupn_end = DupN(self.auto_instance, self.count).__teal__(options)
```
barnjamin, line 48: while trying this out I found that DupN adds an extra element to the stack, since we've already got one from the initial instance we should use `self.count - 1` I think, but please double check me.

prior to change:
```
127 |69  |callsub label9                    |        |[]
203 |123 |proto 0 0                         |        |[]
206 |124 |intc_0 // 0                       |        |[]
207 |125 |dupn 3                            |        |[0]
209 |126 |txna ApplicationArgs 1            |        |[0, 0, 0, 0]
212 |127 |btoi                              |        |[0, 0, 0, 0, 0x000000000000000...
213 |128 |frame_bury 1                      |        |[0, 0, 0, 0, 2]
215 |129 |txna ApplicationArgs 2            |        |[0, 2, 0, 0]
218 |130 |btoi                              |        |[0, 2, 0, 0, 0x000000000000000...
219 |131 |frame_bury 2                      |        |[0, 2, 0, 0, 4]
221 |132 |callsub label12                   |        |[0, 2, 4, 0]
```
After:
```
127 |69  |callsub label9                    |        |[]
203 |123 |proto 0 0                         |        |[]
206 |124 |intc_0 // 0                       |        |[]
207 |125 |dupn 2                            |        |[0]
209 |126 |txna ApplicationArgs 1            |        |[0, 0, 0]
212 |127 |btoi                              |        |[0, 0, 0, 0x0000000000000002]
213 |128 |frame_bury 1                      |        |[0, 0, 0, 2]
215 |129 |txna ApplicationArgs 2            |        |[0, 2, 0]
218 |130 |btoi                              |        |[0, 2, 0, 0x0000000000000004]
219 |131 |frame_bury 2                      |        |[0, 2, 0, 4]
221 |132 |callsub label12                   |        |[0, 2, 4]
```


```suggestion
            dupn_srt, dupn_end = DupN(self.auto_instance, self.count - 1).__teal__(options)
```

ahangsu: right, that makes sense, I was unaware of the dupn doesn't consume the initial one on stack, and your finding is correct.

pyteal/ast/abi/type.py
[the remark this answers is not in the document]
ahangsu: resolved by introducing change of `type_of` in `SubrouineCall` object, removing all of `validate_type` argument.

pyteal/ast/scratchvar.py
```diff
@@ -36,13 +37,15 @@ def storage_type(self) -> TealType:
         """Get the type of expressions that can be stored in this ScratchVar."""
         return self.type
 
-    def store(self, value: Expr) -> Expr:
+    def store(self, value: Expr, validate_type: bool = True) -> Expr:
```
jasonpaulos, line 40: Should this be removed now?

ahangsu: weird, it's not included

ahangsu: pushed.

pyteal/ast/frame.py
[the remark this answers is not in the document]
jasonpaulos: I think based on previous discussion we concluded that ABI return values _can_ be placed on the stack. So should the condition here be changed?

pyteal/ast/frame.py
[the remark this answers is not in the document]
jasonpaulos: Ok, thanks for clarifying.

I _think_ I'd prefer to see separate lists for return values and local ones, but I don't feel strongly about changing that the moment, so I'll resolve.

pyteal/ast/frame.py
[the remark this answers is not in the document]
ahangsu: if we are fine with return value to be in local variables, then the condition should be fine without changing?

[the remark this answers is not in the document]
jasonpaulos: The right thing to do here depends on what happens in #562, so I'm fine with merging this as it is currently and then revisiting.

jasonpaulos (approved):
Looks good, thanks for clarifying and taking suggestions

PR #585, opened 2022-10-26
Frame Ops to `avm8` branch PR

ahangsu:
Introduction to frame pointer ops in avm8 PR


ahangsu:
A few things to debate about:
- on using frame pointer in subroutine, we only used `proto/frame_bury/frame_dig`, while `bury/dupn/popn` are not used.
  I am happy to hide/delete `bury/dupn/popn` expressions, and revisit until we need them.
- `proto` can only be used at the beginning of subroutine, and I don't feel good about exposing it. Shall we remove it in export list, namely `__init__.py`?

pyteal/ast/frame.py
```diff
@@ -0,0 +1,188 @@
...
+from pyteal.types import TealType, require_type
+from pyteal.errors import TealInputError, verifyProgramVersion
+from pyteal.ir import TealBlock, TealSimpleBlock, TealOp, Op
+
+if TYPE_CHECKING:
+    from pyteal.compiler import CompileOptions
+
+
+class Proto(Expr):
+    def __init__(self, arg_num: int, ret_num: int):
```
tzaffi, line 13:
```suggestion
    def __init__(self, num_args: int, num_returns: int):
```
slight rename suggested

tzaffi (commented):
I still have more suggestions, but I'm having cookie-related-issues (?) seeing what my suggestions were so am going to add a preliminary review now

pyteal/ast/frame.py
```diff
@@ -0,0 +1,188 @@
...
+        super().__init__()
+        if arg_num < 0:
+            raise TealInputError(f"subroutine arg number {arg_num} must be >= 0")
+        if ret_num < 0:
+            raise TealInputError(f"return value number {ret_num} must be >= 0")
+        self.arg_num = arg_num
+        self.ret_num = ret_num
```
tzaffi, line 20:
```suggestion
    def __init__(self, num_args: int, num_returns: int):
        super().__init__()
        if num_args < 0:
            raise TealInputError(f"the number of arguments provided to Proto must be >= 0 but {num_args=}")
        if num_returns < 0:
            raise TealInputError(f"the number of return values provided to Proto must be >= 0 but {num_returns=}")
        self.num_args = num_args
        self.num_returns = num_returns
```

one possible improvement - pls. take with a grain of salt

ahangsu: looks right, I am taking

pyteal/ast/frame.py
```diff
@@ -0,0 +1,192 @@
...
+
+    def has_return(self) -> bool:
+        return False
+
+
+Proto.__module__ = "pyteal"
+
+
+class FrameDig(Expr):
+    def __init__(self, depth: int):
```
tzaffi, line 49: According to [Martin Fowler and Phil Karlton](https://martinfowler.com/bliki/TwoHardThings.html):

> There are only two hard things in Computer Science: cache invalidation and naming things.

We've seen how hard the first problem is in our recent boxes release. The second problem occurs more often.

So what should we name `depth`? There's a lot to argue for keeping this name as is. I'll leave the positives for the counter-arguments. But here are what to my ear strikes me as non-ideal about "depth".
* Usually, depth is an unsigned value (though here's a [counter-example](https://www.usgs.gov/observatories/hvo/news/volcano-watch-why-do-some-earthquakes-have-negative-depths) )
* Using at in a natural way, conforms with the volcano link above. So that positive depth -to me- would mean going _below_ the frame pointer into the args, and negative would mean going above, to the return values. This is the opposite of how we would be using it
* There is already an upcoming example where we provide a different name for the argument: [frame_index](https://github.com/algorand/pyteal/blob/6ba772a69721fe43409a7fa03f8235741dd38d8b/pyteal/ast/subroutine.py#L975)

So I propose we rename the parameter to one of:
* `frame_index`
* `displacement`
* `index`
* `delta`
* something else?

tzaffi: BTW: `height` is off the table, as it already means the height of the frame pointer.

tzaffi: `depth` still seems like a good name for the argument of the non frame `Bury` and `Dig`

pyteal/ast/frame.py
```diff
@@ -0,0 +1,192 @@
...
+        return TealType.none
+
+    def has_return(self) -> bool:
+        return False
+
+
+FrameBury.__module__ = "pyteal"
+
+
+class Bury(Expr):
```
tzaffi, line 104: Should we take this as an opportunity to expose `Dig` as well? I don't have a strong opinion here, but it's worth a quick convo.

pyteal/ast/frame.py
```diff
@@ -0,0 +1,192 @@
...
+
+    def has_return(self) -> bool:
+        return False
+
+
+FrameDig.__module__ = "pyteal"
+
+
+class FrameBury(Expr):
+    def __init__(self, what: Expr, depth: int):
```
tzaffi, line 76: Nit: In my brief perusal of our codebase, I've found one other example of `what` used to refer to a parameter of type `Expr`, and that only appears in a test script. On the other hand, the most common such variable name appears to be `value` and so I recommend reconsider renaming `what` to `value`.

tzaffi (commented):
Functionally everything looks great. I'm just quibbling over the namings. That's what part of what makes computer science so challenging and fun.

pyteal/ast/frame.py
ahangsu: yes good point! This is my sinful past in writing cpp, and there are a lot of stuffs named in `what` 😅

pyteal/ast/frame.py
ahangsu: yea I agree, any of proposed naming should be more pertinent than `depth` for it suggests a number >= 0.

I would rename it to `frame_index`, feels most making sense to me.

pyteal/ast/frame.py
ahangsu: I see two (or three) options here:
- expose `Dig` for currently exposed `Bury`
- eliminate `Bury` for `Dig` is not currently exposed here
- do nothing about `Dig`

I prefer first 2, for I preferring providing all or none, rather than giving something in the middle.

Currently I feel hard to use `Dig/Bury` in code, for it strictly require one knowing what is there some `depth` down the stack. But exposing both opcode does not seem a harm to me.

tzaffi: @michaeldiamant @jasonpaulos - Any thoughts? Do we see any immediate use cases for `Bury`? And would those use cases not also necessitate providing `Dig`?

If not, I propose eliminating `Bury` from the PR.

jasonpaulos: I don't see any reason to expose `Dig` and/or `Bury` to users. Directly stack manipulation is too low-level for PyTeal.

ahangsu: got it, I am opt to take that out

tzaffi:
> A few things to debate about:
> 
> * on using frame pointer in subroutine, we only used `proto/frame_bury/frame_dig/Dupn`, while `bury/popn` are not used.
>   I am happy to hide/delete `bury/popn` expressions, and revisit until we need them.
> * `proto` can only be used at the beginning of subroutine, and I don't feel good about exposing it. Shall we remove it in export list, namely `__init__.py`?

Thanks for deleting `bury`. You are mentioning that `popn` isn't used either. Therefore, I recommend removing it as well, unless there is a clear use case (nothing _POPS OUT_ to me 🤣 )

ahangsu:
> You are mentioning that popn isn't used either. Therefore, I recommend removing it as well, unless there is a clear use case (nothing POPS OUT to me 🤣 )

pop popn

tzaffi (approved):
LGTM

ahangsu:
reminder: CHANGELOG.md

pyteal/__init__.py
```diff
@@ -8,6 +8,7 @@
     MIN_TEAL_VERSION,
     DEFAULT_TEAL_VERSION,
     MAX_PROGRAM_VERSION,
+    FRAME_POINTER_VERSION,
```
jasonpaulos, line 11: I'd prefer not to export `FRAME_POINTER_VERSION` before we actually support frame pointer usage. Otherwise it seems misleading.

ahangsu: removed from `__init__.py`

jasonpaulos (approved):

PR #586, opened 2022-10-27
Wrong Comments for GetBit

dabinlee708:
Currently, the comment says
    GetBit(Bytes("base16", "0xf0"), Int(0)) # get the 0th bit of 0xf0, produces 1
    GetBit(Bytes("base16", "0xf0"), Int(7)) # get the 7th bit of 0xf0, produces 0
    GetBit(Bytes("base16", "0xf0"), Int(8)) # get the 8th bit of 0xf0, invalid index
However, base 16 "0xF0" is 11110000 in binary (ignoring leading zeroes). The 0th bit should be 0 while the 7th bit should be 1 instead.

### DIFF
--- a/docs/arithmetic_expression.rst
+++ b/docs/arithmetic_expression.rst
@@ -117,8 +117,8 @@ The :any:`GetBit` expression can extract individual bit values from integers and
     GetBit(Int(16), Int(63)) # get the 63rd bit of 16, produces 0
     GetBit(Int(16), Int(64)) # get the 64th bit of 16, invalid index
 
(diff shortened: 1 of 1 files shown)


ahangsu:
According to the same file at line 99-102:

> * For byte strings, bit indexing begins at the first bit. For example, the bit at index 0 of the base16
>  byte string :code:`0xf0` (:code:`11110000` in binary) is 1. Any index less than 4 has a bit
>  value of 1, and any index 4 or greater has a bit value of 0. Any index less than 8 times the length
>  of the byte string is valid.

I wonder if the change is correct...?

PR #589, opened 2022-11-02
Scratch branch

tzaffi:
Not yet sure this actually works, but let's let CI give us a clue.

I'm proposing allowing for skipping validation in `AbstractVar.store()` with the assumption that the caller overrides with a custom validation at the site of the call.


ahangsu:
Oh no I am making a huge mistake... the previous `scratch-branch` is not rebased to `avm8`, and that is why there were ghost commits... Now I tried to rebase to `abi-value-data-scheme`, but annihilated all your changes... 🤦

ahangsu:
It seems very hard to get stuffs back, I would suggest we close this branch, and branch from current [abi-value-data-schema](https://github.com/algorand/pyteal/tree/abi-value-data-schema), does that feel okay with you?

tzaffi:
> It seems very hard to get stuffs back, I would suggest we close this branch, and branch from current [abi-value-data-schema](https://github.com/algorand/pyteal/tree/abi-value-data-schema), does that feel okay with you?

no worries... I'm less and less convinced that my idea is actually an improvement... Closing and possibly reopening another.

PR #590, opened 2022-11-02
Validate types option

tzaffi:
There are pros and cons to this PR. We can talk further but we should only merge if there is a consensus that this is a net benefit.

* Cons:
  *  requiring the handling of a new `validate_types` param of `store()` in all subclasses of AbstractVar even though only a single class currently needs it
  
* Pros
  * increased clarity of when validation is actually happen
  * delegate custom validation to the code location that is most concerned with it
  * remove apparently unrelated code from `ScratchVar`


examples/application/abi/algobank.json
```diff
@@ -1,5 +1,6 @@
 {
     "name": "AlgoBank",
+    "desc": null,
```
tzaffi, line 3: locally, I was expecting the `desc` field. I'm not sure why this new failure.

ahangsu (approved):

PR #592, opened 2022-11-03
allow app id to be none since it may be called during create

barnjamin:
While answering a question in the forum, I realized I'd not considered the case where a MethodCall may be used to create an application. 

Since the AVM doesn't like the manual setting of AppId 0, we should allow app_id to be None and prevent setting the app id field

### DIFF
--- a/pyteal/ast/itxn.py
+++ b/pyteal/ast/itxn.py
@@ -1,7 +1,7 @@
 from enum import Enum
 import algosdk
 
-from typing import TYPE_CHECKING, cast
+from typing import TYPE_CHECKING, cast, Optional
 from pyteal.ast.int import EnumInt
 from pyteal.ast.for_ import For
 from pyteal.ast.int import Int
@@ -253,10 +253,10 @@ class InnerTxnBuilder:
     def ExecuteMethodCall(
         cls,
         *,
-        app_id: Expr,
+        app_id: Optional[Expr],
         method_signature: str,
         args: list[abi.BaseType | Expr | dict[TxnField, Expr | list[Expr]]],
-        extra_fields: dict[TxnField, Expr | list[Expr]] = None,
+        extra_fields: Optional[dict[TxnField, Expr | list[Expr]]] = None,
     ) -> Expr:
         """Performs a single app call transaction formatted as an ABI method call.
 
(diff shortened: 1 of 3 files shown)


jasonpaulos:
Good catch. The solution looks good but it'd be nice to see a regression test added to `test_InnerTxnBuilder_method_call`

tzaffi (commented):
Thanks for the fix. I'm expecting `mypy` to flag the fact that `app_id` could now be `None`. So, I recommend changing its signature to:

```python
app_id: Optional[Expr]
```

ahangsu:
looks good, I will approve after a test on app-id == None gets added

ahangsu:
also add a line in CHANGELOG `Fixed` section about this fix

ahangsu (approved):
looks right to me

pyteal/ast/itxn.py
```diff
@@ -1,7 +1,7 @@
 from enum import Enum
 import algosdk
 
-from typing import TYPE_CHECKING, cast
+from typing import TYPE_CHECKING, cast, Optional
```
tzaffi, line 4:
```suggestion
from typing import TYPE_CHECKING, cast
```

pyteal/ast/itxn.py
```diff
@@ -253,10 +253,10 @@ def SetFields(cls, fields: dict[TxnField, Expr | list[Expr]]) -> Expr:
     def ExecuteMethodCall(
         cls,
         *,
-        app_id: Expr,
+        app_id: Optional[Expr],
```
tzaffi, line 256:
```suggestion
        app_id: Expr | None,
```

Do you want to allow default `= None` or have the user explicitly provide `None` ?

pyteal/ast/itxn.py
```diff
@@ -253,10 +253,10 @@ def SetFields(cls, fields: dict[TxnField, Expr | list[Expr]]) -> Expr:
...
         method_signature: str,
         args: list[abi.BaseType | Expr | dict[TxnField, Expr | list[Expr]]],
-        extra_fields: dict[TxnField, Expr | list[Expr]] = None,
+        extra_fields: Optional[dict[TxnField, Expr | list[Expr]]] = None,
```
tzaffi, line 259:
```suggestion
        extra_fields: dict[TxnField, Expr | list[Expr]] | None = None,
```

pyteal/ast/itxn.py
```diff
@@ -310,10 +311,10 @@ def ExecuteMethodCall(
     def MethodCall(
         cls,
         *,
-        app_id: Expr,
+        app_id: Optional[Expr],
```
tzaffi, line 314:
```suggestion
        app_id: Expr | None,
```

pyteal/ast/itxn.py
```diff
@@ -310,10 +311,10 @@ def ExecuteMethodCall(
...
         method_signature: str,
         args: list[abi.BaseType | Expr | dict[TxnField, Expr | list[Expr]]],
-        extra_fields: dict[TxnField, Expr | list[Expr]] = None,
+        extra_fields: Optional[dict[TxnField, Expr | list[Expr]]] = None,
```
tzaffi, line 317:
```suggestion
        extra_fields: dict[TxnField, Expr | list[Expr]] | None = None,
```

tzaffi (approved):

PR #593, opened 2022-11-03
Memory layout for frame var and frame ops

ahangsu:
After a discussion with @barnjamin on 2022/11/03, we think it would help in providing fine grained type checking over the stack, by introducing a way to depicting memory like type stack.

This should benefit further work down the line such as using local variables and using frame pointer in router code generation.


pyteal/ast/frame.py
```diff
@@ -11,8 +14,101 @@
...
+
+    def __post_init__(self):
+        if self.has_output and len(self.local_stack_types) == 0:
+            raise TealInternalError(
+                "ProtoStackLayout initialization error: cannot output without local variable allocs."
+            )
+
+        self.output_index = 0 if self.has_output else None
+        if not all(map(lambda t: types_match(t, TealType.none), self.arg_stack_types)):
+            raise TealInternalError("Variables in frame memory layout must be typed.")
```
tzaffi, line 66: The error message makes me believe that really we should be asserting:
```python
all(map(lambda t: t != TealType.none, self.arg_stack_types)):
```

pyteal/ast/frame.py
```diff
@@ -11,8 +14,101 @@
...
+
+        # Type check of local variables are performed over LocalTypeSegments
+        self.succinct_repr = [
+            LocalTypeSegment(t_type, len(list(dup_seg)))
+            for t_type, dup_seg in groupby(self.local_stack_types)
+        ]
+
+        # anytype at i + 1 merge to i
+        for i in reversed(range(len(self.succinct_repr) - 1)):
```
tzaffi, line 75:
```suggestion
        for i in range(len(self.succinct_repr) - 2, -1, -1)):
```

not sure this is actually better

pyteal/ast/frame.py
```diff
@@ -11,8 +14,101 @@
...
+            if self.succinct_repr[i + 1].local_type == TealType.anytype:
+                self.succinct_repr[i].cnt += self.succinct_repr[i + 1].cnt
+                self.succinct_repr.pop(i + 1)
```
tzaffi, line 78: My gut is telling me that this will cause gaps and therefore be faulty in the case that there are consecutive non-`anytype`s in the middle of the sequence. However, a test case such as the following added to `SuccinctReprTestCase` could prove my gut wrong:

```python
        SuccinctReprTestCase(
            [
                pt.TealType.anytype,
                pt.TealType.bytes,
                pt.TealType.bytes,
                pt.TealType.uint64,
                pt.TealType.uint64,
                pt.TealType.uint64,
                pt.TealType.anytype,
                pt.TealType.anytype,
                pt.TealType.anytype,
                pt.TealType.anytype,
                pt.TealType.bytes,
                pt.TealType.bytes,
                pt.TealType.uint64,
                pt.TealType.uint64,
                pt.TealType.uint64,
            ],
            [
                LocalTypeSegment(pt.TealType.bytes, 3),
                LocalTypeSegment(pt.TealType.uint64, 7),
                LocalTypeSegment(pt.TealType.bytes, 2),
                LocalTypeSegment(pt.TealType.uint64, 3),
            ],
        ),
```

pyteal/ast/frame.py
```diff
@@ -11,8 +14,101 @@
...
+
+        if (
+            len(self.succinct_repr) > 1
+            and self.succinct_repr[0].local_type == TealType.anytype
+        ):
+            self.succinct_repr[1].cnt += self.succinct_repr[0].cnt
+            self.succinct_repr.pop(0)
+
+    def __getitem__(self, index: int):
```
tzaffi, line 87:
```suggestion
    def __getitem__(self, index: int) -> TealType:
```

pyteal/ast/frame.py
```diff
@@ -11,8 +14,101 @@
...
+
+LocalTypeSegment.__module__ = "pyteal"
+
+
+@dataclass
+class ProtoStackLayout:
+    arg_stack_types: list[TealType]
+    local_stack_types: list[TealType]
+    has_output: bool
+    output_index: Optional[int] = field(init=False)
```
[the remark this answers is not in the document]
tzaffi: I'm still fuzzy with how all of this is going to work but it seems like there could be some benefit in having another field `local_stack_index` which demarcates where the local variables will be stored after the output variables.

But probably this is an uneeded complication.

pyteal/ast/frame.py
```diff
@@ -86,16 +178,19 @@ def has_return(self) -> bool:
...
 class FrameBury(Expr):
-    def __init__(self, value: Expr, frame_index: int):
-        from pyteal.ast.subroutine import SubroutineCall
-
+    def __init__(
+        self,
+        value: Expr,
+        frame_index: int,
+        *,
+        validate_types: bool = True,
```
tzaffi, line 186: I don't yet see the value of exposing the `validate_types` parameter. Perhaps it is utilized in the next PR?

tzaffi (commented):
I left several questions and suggestions.

Also, I'm trying to understand the cost/benefit of this approach.

I believe we have the following pro:
* better type safety

and the following cons:
* more complex logic (therefore a greater bug surface area)
* slightly costlier? (I think we may be creating more calls to`dupn` than otherwise)

pyteal/ast/frame.py
```diff
@@ -11,8 +14,101 @@
     from pyteal.compiler import CompileOptions
 
 
+@dataclass
+class LocalTypeSegment:
```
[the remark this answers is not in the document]
ahangsu: yes, that are all good points to me:
- `__teal__()` method seems like a `Expr`: After some thoughts during the weekend, I believe we can attempt to make both `LocalTypeSegment` and `ProtoStackLayout` inherit from `Seq`, for they serve a similar purpose without revealing `type_of`. I hope this may help in sourcemapping!
- `dataclass` is just out of laziness for me unwilling to declare `__init__` function. For the simplicity and increasing complexity, I opt to change the class declaration to vanilla flavor of python class decl.

pyteal/ast/frame.py
```diff
@@ -11,8 +14,101 @@
...
+    succinct_repr: list[LocalTypeSegment] = field(init=False)
+
+    def __post_init__(self):
+        if self.has_output and len(self.local_stack_types) == 0:
+            raise TealInternalError(
+                "ProtoStackLayout initialization error: cannot output without local variable allocs."
+            )
+
+        self.output_index = 0 if self.has_output else None
+        if not all(map(lambda t: types_match(t, TealType.none), self.arg_stack_types)):
```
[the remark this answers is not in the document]
ahangsu: yes good point, I take the change.

pyteal/ast/frame.py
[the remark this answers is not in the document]
ahangsu: I just took out the dependencies over `dataclasses` from `LocalTypeSegment` and `ProtoStackLayout`, and I hope (yet unsure how) it would help in sourcemapping.

pyteal/ast/frame.py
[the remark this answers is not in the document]
ahangsu: I agree that `output_index` is not quite useful, and actually it is somewhat a drawback.

When I was creating these declaration, I was thinking about `ABIReturnSubroutine`, but the purpose of `ProtoStackLayout` should be more than this, and encoding output index to be 0 is not quite interesting.

Now I am refeeding this `ProtoStackLayout` with `num_returns`, such that it perform some requirements over the stack, and removing the hard-coded output index 0.

pyteal/ast/frame.py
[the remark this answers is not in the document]
ahangsu: after a quick thought, I am taking out the dependencies to `Seq`, and let them directly inherit from `Expr`, just to decrease the complication.

pyteal/ast/frame.py
ahangsu: good point, I hope I changed the code to what you just suggested here.

pyteal/ast/frame.py
ahangsu: good testcase, add to frame_test and seems to be working!

ahangsu: Just to leave this thread open: I think I am tryin' to act smart here by merging `TealType.anytype` to nearest `TealType.byte` or `TealType.uint64`. But does that worth it? People feel free to persuade me on not worth it.

tzaffi: I would benefit from an explanation of the motivation for the algorithm. This has to do with my "pros/cons" comment above. But if the pro is strong enough, and the tests give confidence that there are no bugs, then I would argue that it's "worth it".

pyteal/ast/frame_test.py
```diff
@@ -105,3 +113,180 @@ def test_dupn_invalid():
 
     with pytest.raises(pt.TealInputError):
         DupN(pt.Int(1), 1).__teal__(avm7Options)
+
+
+def test_local_type_segment_invalid():
```
tzaffi, line 118: would be nice to assert a bit about the expected error. For example, the last case could be rewritten as:

```python
    with pytest.raises(pt.TealInternalError) as tie:
        LocalTypeSegment(pt.TealType.none, 2)
    assert "frame memory layout must be typed" in str(tie)
```

tzaffi (commented):
My main comments have been addressed. A couple of points are worth discussing further, but basically the shape of the PR is GTG from my perspective.

pyteal/ast/frame.py
barnjamin: I'm somewhat confused by this sort of "casting" into a type based on adjacent types, why not keep them anytype?

ahangsu: Uh I missed discussing the algorithm, my bad.

The motivation in applying the algorithm is to generate smallest code for local var init.

The expected IO of the algorithm follows:
- given a "sucinct" expression of local variable type description, e.g.,
  `[(TealType.uint64, 3), (TealType.anytype, 4), (TealType.bytes, 5)]`
- output a "shrinked" succinct expression of local variable type description, e.g., 
  `[(TealType.uint64, 7), (TealType.bytes, 5)]`

Now I am thinking out loud:
- How much it saved? Probably at most 2 ops for each occurrence of `TealType.anytype`.
- Is it somewhat confusing? I somehow see the confusion in maintaining the code.

Given the complexity (maybe unnecessary) and profit gain from smaller code, I am slightly prone to taking the algorithm out.

pyteal/ast/frame_test.py
ahangsu: good point, just add all `tie` error message check

pyteal/ast/frame.py
ahangsu: I just take the algorithm out, at a point I am unconvinced that the algorithm brings more benefit.

pyteal/ast/frame.py
ahangsu: resolving for I removed the whole algorithm...

pyteal/ast/frame.py
ahangsu: okay, circling back to this discussion.

My thinking is, `FrameBury` should be only used by `FrameVar`, through `store` method.
- `FrameBury` used to be oblivious of the type of the stack position to bury to, now we can use `ProtoStackLayout` to tell the exact `TealType`.
- The additional knowledge of `TealType` allows `FrameBury` to validate type, so I extend the option `validate_types` to `FrameBury`, so it validates the type of argument that it construct from.
- The final type checking does not need to replicate 2 pieces of similar logic as you pointed out before: https://github.com/algorand/pyteal/pull/584#discussion_r1014341184, so I see it as somehow an improvement. The final implementation of `FrameVar.store` should look somehow like `ScratchVar.store`.

barnjamin (approved):

tzaffi (approved):
LGTM

PR #594, opened 2022-11-03
Fail Releasing when version isn't set correctly in `setup.py`

tzaffi:
# Use `make fail-if-unreleasable`

### When `setup.py` version _is GREATER THAN_ PyPi's version
```sh
❯ make fail-if-unreleasable
... some annoying ignorable warnings ...
You're all good to go. Enjoy releasing new_version=0.20.2 (> pypi=0.20.1)
```

### When `setup.py` version _is NOT GREATER THAN_ PyPi's version

```sh
❯ make fail-if-unreleasable
... some annoying ignorable warnings ...
  File "<string>", line 1, in <module>
AssertionError: cannot release as local=[0, 20, 1] v. pipy=[0, 20, 1]
make: *** [fail-if-unreleasable] Error 1
```

## Testing Locally via `act` simulator:
Before running the following, comment out the line `if: ...` in the `validate-release:` section of `.github/workflows/build.yml`

```sh
❯ make local-gh-job ACT_JOB=validate-release
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -77,10 +77,26 @@ jobs:
           name: pyteal.docset
           path: docs/pyteal.docset.tar.gz
 
+  validate-release:
+    runs-on: ubuntu-20.04
+    container: python:3.10
+    needs: ['build-test']
+    strategy:
+      matrix:
+        python: [ "3.10" ]
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Validate Release
+        run: make fail-if-unreleasable
+
+
   upload-to-pypi:
     runs-on: ubuntu-20.04
     container: python:3.10
-    needs: ['build-test', 'run-integration-tests', 'build-docset']
+    needs: ['build-test', 'validate-release', 'run-integration-tests', 'build-docset']
     if: ${{ github.event_name == 'push' && startsWith(github.ref, 'refs/tags') }}
     steps:
       - name: Check out code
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -7,7 +7,7 @@
 
 setuptools.setup(
     name="pyteal",
-    version="0.20.1",
+    version="0.20.2",
```
tzaffi, line 10:
```suggestion
    version="0.20.1",
```

.github/workflows/build.yml
```diff
@@ -77,10 +77,23 @@ jobs:
...
+          fetch-depth: 0
+      - name: Validate Release
+        run: make fail-if-unreleasable
+
+
   upload-to-pypi:
     runs-on: ubuntu-20.04
     container: python:3.10
-    needs: ['build-test', 'run-integration-tests', 'build-docset']
+    needs: ['validate-release', 'build-test', 'run-integration-tests', 'build-docset']
```
michaeldiamant, line 96: @tzaffi Let me know if I've understood the proposed process?

* My understanding:  The PR catches a version mismatch _after_ git tag creation, but _before_ pypi upload.
* If that's the case, I think the PR provides _some_ improvement because it more directly highlights the error.  But, I am interested in an approach (either via process or tooling) to either _prevent_ or _minimize_ errant tag creation.  I have some thoughts here more easily discussed live.
* For reference - Here's an example build failure when `setup.py` version was _not_ changed during a release:  https://github.com/algorand/pyteal/actions/runs/3387261940/jobs/5627959460.

tzaffi:
After discussing with @michaeldiamant , I'm closing as this fun hack doesn't provide a good enough solution to the original problem that was encountered.

PR #596, opened 2022-11-07
Update voting example

jasonpaulos:


examples/application/abi/poll/contract.json
```diff
@@ -0,0 +1,65 @@
+{
+    "name": "OpenPollingApp",
+    "methods": [
+        {
+            "name": "create",
+            "args": [
+                {
+                    "type": "string[]",
+                    "name": "options",
+                    "desc": "A list of options for the poll. This list should not contain duplicate entries."
```
tzaffi, line 10: Nice. This could replace our `Polly` with a slack hook!

examples/application/abi/poll/contract.json
```diff
@@ -0,0 +1,65 @@
...
+            "returns": {
+                "type": "void"
+            },
+            "desc": "Marks this poll as closed.\nThis will fail if the poll is already closed."
+        },
+        {
+            "name": "submit",
+            "args": [
+                {
+                    "type": "uint16",
```
tzaffi, line 43: nit: `uint8` (or even smaller) probably suffices. But there's no real requirement to limit user choices to 256.

tests/unit/compile_test.py
```diff
@@ -35,6 +35,34 @@ def test_abi_algobank():
         assert contract.dictify() == expected_contract
 
 
+def test_abi_vote():
```
tzaffi, line 38:
```suggestion
def test_abi_poll():
```

nit - slightly more in line with the current naming

tzaffi (commented):
looking really promising!

examples/application/abi/poll/contract.py
```diff
@@ -0,0 +1,107 @@
...
+        options: A list of options for the poll. This list should not contain duplicate entries.
+        resubmission: A boolean value indicating whether this poll allows accounts to change their
+            answer after their first submission.
+        can_resubmit: Whether this poll allows accounts to change their submissions or not.
+    """
+    return Seq()
+
+
+@router.method(name="open")
+def open_poll() -> Expr:
```
ahangsu, line 33: good to see naming overriding being used 👍

examples/application/abi/poll/contract.py
```diff
@@ -0,0 +1,182 @@
...
+option_count_keys = [
+    Bytes(option_count_prefix + b"\x00"),
+    Bytes(option_count_prefix + b"\x01"),
+    Bytes(option_count_prefix + b"\x02"),
+]
+
+
+@router.method(no_op=CallConfig.CREATE)
+def create(
+    options: abi.StaticArray[abi.String, Literal[3]], can_resubmit: abi.Bool
```
tzaffi, line 37: If we really want more choices, we could have a factory that takes `N` = the number of choices as a parameter and returns the router. But it seems overcomplicates for the demo, so I would prefer it to be kept this way. But we could still mention that it's easily achievable.

examples/application/abi/poll/contract.py
```diff
@@ -0,0 +1,182 @@
...
+    """Create a new polling application.
+
+    Args:
+        options: A list of options for the poll. This list should not contain duplicate entries.
+        can_resubmit: Whether this poll allows accounts to change their submissions or not.
+    """
+    return Seq(
+        App.globalPut(open_key, Int(0)),
+        App.globalPut(resubmit_key, can_resubmit.get()),
+        App.globalPut(option_name_keys[0], options[0].use(lambda value: value.get())),
```
tzaffi, line 48: sanity checking that a more direct approach for setting the value isn't available

PR #600, opened 2022-11-09
Frame Pointer usage in router

ahangsu:
Last piece of #547. Aiming to remove as much scratch slot usage in router as possible.

A rough idea is that, the frame pointer thing only apply in ABI return subroutine. We should create dynamically a subroutine that takes `Txn.application_args` in, internally use local variables (from frame pointer) as ABI value storage, and return results in log.

Need some manual tests and compiler test.


pyteal/ast/router.py
```diff
@@ -459,34 +566,90 @@ def wrap_handler(
...
-                    CondNode(
-                        walk_in_cond,
-                        Seq(Assert(cond), self.wrap_handler(True, handler)),
-                    )
-                )
-            case 1:
-                self.conditions_n_branches.append(
-                    CondNode(walk_in_cond, self.wrap_handler(True, handler))
-                )
-            case 0:
```
barnjamin, line 479: noticed this one has a condition for 0, the new one doesnt?

ahangsu: The case has been ruled out in the new code snippet
```python
        if isinstance(cond, int) and cond == 0:
            return
```

tzaffi:
Before I review the details of these changes, it strikes me as a good use case for graviton tests. Unfortunately, the only existing graviton tests for the router are in graviton itself. This might be a good juncture to bring these tests directly into PyTeal so we could then assert that functionality remains identical regardless of version that is being compiled. However, this may slow down the release of the feature. It's worth a bit of discussion.

pyteal/compiler/compiler_test.py
```diff
@@ -3413,3 +3413,1214 @@ def approve_if_odd(condition_encoding: pt.abi.Uint32) -> pt.Expr:
...
+        "QuestionableRouterGenerateCodeWithFramePointer", on_completion_actions
+    )
+    add_methods_to_router(_router_with_oc)
+    (
+        actual_ap_with_oc_compiled,
+        actual_csp_with_oc_compiled,
+        _,
+    ) = _router_with_oc.compile_program(version=8)
+
+    expected_ap_with_oc = """#pragma version 8
```
tzaffi, line 3427: Wow. Not a single `store` !

bbroder-algo: I dig it.

ahangsu:
> the only existing graviton tests for the router are in graviton itself.

Yes, I remember seeing a set of `Questionable.teal` file in graviton, I guess that is the router integration test inside graviton.

And I agree it might be good to enhance graviton s.t., we can write integration test for router generated code inside PyTeal rather than inside graviton.

pyteal/ast/router.py
```diff
@@ -274,13 +280,207 @@ class CondNode:
...
+                )
+            case 1:
+                return CondNode(
+                    walk_in_cond,
+                    ASTBuilder.wrap_handler(
+                        True, self.method, use_frame_pt=use_frame_pt
+                    ),
+                )
+            case _:
+                raise TealInputError("Invalid condition input for CondWithMethod")
```
tzaffi, line 319:
```suggestion
        if not (isinstance(self.condition, Expr) or self.condition) == 1:
            raise TealInputError("Invalid condition input for CondWithMethod")
            
        res = ASTBuilder.wrap_handler(True, self.method, use_frame_pt=use_frame_pt)
        if is_instance(self.condition, Expr):
            res = Seq(Assert(self.condition), res)
        return CondNode(walk_in_cond, res)
```

tzaffi: Just a suggestion, since now we're down to only 2 cases.

tzaffi (commented):
I'm still reading the PR, but I wanted to put up some comments before standup.

pyteal/ast/router.py
```diff
@@ -274,13 +280,194 @@ class CondNode:
...
+    - condition: Expr | int, which matches the on_completion conditions
+    - method: ABIReturnSubroutine, assuming that the method is ABI routable
+    """
+
+    method_sig: str
+    condition: Expr | int
+    method: ABIReturnSubroutine
+
+    def to_cond_node(self, use_frame_pt: bool = False) -> CondNode:
+        walk_in_cond = Txn.application_args[0] == MethodSignature(self.method_sig)
```
tzaffi, line 298: I should have suggested to but the assertion first...

pyteal/ast/router.py
```diff
@@ -459,34 +560,91 @@ def wrap_handler(
...
-                return
-            case _:
-                raise TealInputError("Invalid condition input for add_method_to_ast")
+        if isinstance(cond, int) and cond == 0:
+            return
+        self.methods_with_conds.append(CondWithMethod(method_signature, cond, handler))
+
+    def program_construction(self, use_frame_pt: bool = False) -> Expr:
+        self.conditions_n_branches += [
+            i.to_cond_node(use_frame_pt=use_frame_pt) for i in self.methods_with_conds
```
tzaffi, line 637: `i` seems like an int, but maybe you're in a fight against `black` not breaking the line up?

ahangsu: oh, I think I am just lazy about naming it, probably renaming to something more meaningful

ahangsu: renaming to `method_with_cond`.

pyteal/ast/router.py
```diff
@@ -696,7 +867,9 @@ def contract_construct(self) -> sdk_abi.Contract:
 
         return sdk_abi.Contract(self.name, self.methods, self.descr)
 
-    def build_program(self) -> tuple[Expr, Expr, sdk_abi.Contract]:
+    def build_program(
```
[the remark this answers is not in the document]
ahangsu: I think it works, but I want to propose using both `version` and `optimize` to generate the boolean of `use_frame_pointer`, what do you think?

pyteal/compiler/compiler.py
```diff
@@ -264,6 +264,7 @@ def compileTeal(
             the compiled program's size. Enabling this option requires a minimum program version of 3.
             Defaults to false.
         optimize (optional): OptimizeOptions that determine which optimizations will be applied.
+        frame_pointers (optional): An option to opt out of building with frame pointer after version 8.
```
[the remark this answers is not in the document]
ahangsu: That is leftovers from previous PRs, good catch for calling out, I am removing this line.

pyteal/ast/router.py
```diff
@@ -274,13 +280,194 @@ class CondNode:
...
+CondWithMethod.__module__ = "pyteal"
+
+
 @dataclass
 class ASTBuilder:
     conditions_n_branches: list[CondNode] = field(default_factory=list)
+    methods_with_conds: list[CondWithMethod] = field(default_factory=list)
+
+    @staticmethod
+    def __handler_validity_check(
```
tzaffi, line 318: It was a bit confusing to me to read this method name and try to understand why it would return a value. So maybe a renaming to something like `_filter_invalid_and_typecast()` would be a bit clearer? (essentially this is a type cast, even though now `mypy` is smart enough to figure this out without you explicitly using `cast()`)

pyteal/ast/router.py
```diff
@@ -459,34 +560,92 @@ def wrap_handler(
...
                 )
+        else:
+            subroutine_ify = ASTBuilder.__de_abify_subroutine(handler)
+            return Seq(
+                subroutine_ify(),
+                Approve(),
+            )
+
+    @staticmethod
+    def __de_abify_subroutine(
```
tzaffi, line 571: I think since this logic is meant for `frame_pointers` usage, it would help to evoke this by adding `fp` or `frame_pointers` to the method name. Also, out of symmetry, it's probably preferable to summarize the non-frame pointers code generation in lines 527-562 with a function as well.

tzaffi: Once that is done, there might be other "obvious" points where the logic can be unified.

tzaffi: or it may be the case that you can pass `use_frame_pointers` into this function and use in both places.

pyteal/ast/router.py
```diff
@@ -459,34 +560,92 @@ def wrap_handler(
...
+            app_arg_vals,
+            txn_arg_vals,
+            handler,
+            use_frame_pt=True,
+        )
+
+        subroutine_caster = Subroutine(TealType.none, f"{handler.name()}_caster")
+        return_expr: Expr
+
+        if not proto or not proto.mem_layout:
```
tzaffi, line 596: Is there a case that `ASTBuilder.__decode_constructions_and_args()` may return an invalid such `proto` ? If so, it should probably in charge of this assertion (for failing fast's sake) especially since it has access to the `use_frame_ptr` parameter.

pyteal/ast/router.py
```diff
@@ -459,34 +560,92 @@ def wrap_handler(
...
+            subroutine_call: abi.ReturnedValue = cast(
+                abi.ReturnedValue, handler(*arg_vals)
+            )
+
+            return_expr = Seq(
+                *proto.mem_layout._succinct_repr(),
+                *decode_instructions,
+                subroutine_call.store_into(output_temp),
+                abi.MethodReturn(output_temp),
+            )
```
tzaffi, line 621:
```suggestion
        decoding = [*proto.mem_layout._succinct_repr(), *decode_instructions]
        if handler.type_of() == sdk_abi.Returns.VOID:
            returning = cast(Expr, handler(*arg_vals))
        else:
            output_temp: abi.BaseType = cast(
                OutputKwArgInfo, handler.output_kwarg_info
            ).abi_type.new_instance()
            output_temp._stored_value = FrameVar(proto, 0)
            subroutine_call: abi.ReturnedValue = cast(
                abi.ReturnedValue, handler(*arg_vals)
            )
            returning = [subroutine_call.store_into(output_temp), abi.MethodReturn(output_temp)]
        return Seq(*decoding, *returning)
```
please ignore this suggestion if you disagree

tzaffi (commented):
Looks solid. I made some ignorable suggestions regarding some code streamlining. And I still would like to come up with a strategy around graviton testing.

pyteal/ast/router.py
ahangsu: I am taking with some minor changes.

pyteal/ast/router.py
ahangsu: sure, I think this method serves following purpose: filter out unroutable ABIReturnSubroutine + Expr + normal subroutine, leave only routable ABIReturnSubroutine passing. So your naming makes the most sense, probably I will take `__filter_invalid_handlers_and_typecast`.

pyteal/ast/router.py
ahangsu: The suggestion on renaming the method name to be related with frame pointers makes a lot of sense.

I prefer packing vanilla flavor of code generation to be another function, rather controlled by passing `use_frame_pointer`, for that will be way too complex and hard to maintain the resulting code.

pyteal/ast/router.py
ahangsu: I somehow doubt why this check is where, generally `ASTBuilder.__decode_constructions_and_args()` should not return an invalid `proto`, for in our case, we want `proto 0 0` with some memory layouts.

I will simply annihilate this unnecessary check.

pyteal/ast/router.py
```diff
@@ -131,6 +135,9 @@ def approval_cond(self) -> Expr | int:
 ActionType = Expr | SubroutineFnWrapper | ABIReturnSubroutine
 
 
+MethodConfig.__module__ = "pyteal"
```
bbroder-algo, line 138: duped?

ahangsu: yes, bad merge.

pyteal/ast/router.py
```diff
@@ -259,13 +266,203 @@ class CondNode:
...
+    """
+
+    method_sig: str
+    condition: Expr | int
+    method: ABIReturnSubroutine
+
+    def to_cond_node(self, use_frame_pt: bool = False) -> CondNode:
+        walk_in_cond = Txn.application_args[0] == MethodSignature(self.method_sig)
+
+        if not (isinstance(self.condition, Expr) or self.condition == 1):
```
bbroder-algo, line 283: 1 is an enum?

pyteal/ast/router.py
```diff
@@ -259,13 +263,203 @@ class CondNode:
...
+        # Only args that appear in app args
+        app_arg_vals: list[abi.BaseType] = [
+            ats for ats in arg_vals if not isinstance(ats, abi.Transaction)
+        ]
+
+        # only transaction args (these are omitted from app args)
+        txn_arg_vals: list[abi.Transaction] = [
+            ats for ats in arg_vals if isinstance(ats, abi.Transaction)
+        ]
+
```
bbroder-algo, line 344: current transaction types:
    TransactionTypeSpec(),
    PaymentTransactionTypeSpec(),
    KeyRegisterTransactionTypeSpec(),
    AssetConfigTransactionTypeSpec(),
    AssetFreezeTransactionTypeSpec(),
    AssetTransferTransactionTypeSpec(),
    ApplicationCallTransactionTypeSpec(),

pyteal/ast/router.py
```diff
@@ -259,13 +263,203 @@ class CondNode:
...
+        subroutine: ABIReturnSubroutine,
+        use_frame_pt: bool = False,
+    ) -> tuple[list[Expr], list[abi.BaseType], Optional[Proto]]:
+        """
+        Assumption: arg_vals = app_args_vals union with txn_arg_vals
+        """
+
+        # if subroutine has ABI output, then local variables start from 1
+        # otherwise local variables start from 0
+        index_start_from = int(subroutine.output_kwarg_info is not None)
```
bbroder-algo, line 370: because it is returning an index
maybe `index_start_from = 0 if subroutine.output_kwarg_info is None else 1`

ahangsu: [7ade42b](https://github.com/algorand/pyteal/pull/600/commits/7ade42b68df7399ce389d9891265192b2d0e5502)

pyteal/ast/router.py
ahangsu: no, `self.condition` can be `Expr | int`:
- `Expr` stands for some on-call conditions, e.g.,:
  `Txn.on_completion() == OnComplete.NoOp and Txn.on_completion() == OnComplete.OptIn`.
- 1 or 0 stands for 2 extreme cases of on-call conditions:
  - 1 stands for allowed for any on call option
  - 0 stands for no on call option is allowed, which is a dead branch.

so here, it means if `self.condition` is not valid, do something. Validity decided by `Expr` or == 1.

ahangsu: also explained in more documentation [52a9cd6](https://github.com/algorand/pyteal/pull/600/commits/52a9cd6335ded1cf340f492c7b2cbf23bdadfe98)

pyteal/ast/router.py
```diff
@@ -259,13 +266,203 @@ class CondNode:
 CondNode.__module__ = "pyteal"
 
 
-@dataclass
+@dataclass(frozen=True)
+class CondWithMethod:
+    """CondWithMethod unifies the generation to CondNode from compilation w/wo frame pointer.
```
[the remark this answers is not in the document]
ahangsu: [52a9cd6](https://github.com/algorand/pyteal/pull/600/commits/52a9cd6335ded1cf340f492c7b2cbf23bdadfe98)

pyteal/ast/router.py
ahangsu: right, I think `TransactionTypeSpec` is rather generic, while the others are specific (concrete).

pyteal/ast/router.py
```diff
@@ -605,9 +611,50 @@ def __de_abify_subroutine_vanilla(
...
+            |  +--------------------------------------------------------------------------------+
+            |  | At this point, all of the handler arguments (ABI typed) are prepared.          |
+            |  | We call the handler with all these handler arguments.                          |
+            |  | If handler has output, we dig the result to top of the stack, and log it.      |
+            |  |                                                                                |
+            |  | This section represents the expressions in `returning_steps`.                  |
+            +  +--------------------------------------------------------------------------------+
+            |                                                                                   |
+            | Now that handler return (if exists) is handled, we `Approve()` and exit prog.     |
+            +-----------------------------------------------------------------------------------+
```
tzaffi, line 656: Nice 👍

.github/workflows/build.yml
```diff
@@ -45,7 +45,7 @@ jobs:
           echo $installed
           [[ $installed =~ "Python ${expected}" ]] && echo "Configured Python" || (echo "Failed to configure Python" && exit 1)
       - name: Start algod
-        run: make algod-start
+        run: make algod-start-report
```
tzaffi, line 48: This prints out the go-algorand hash summary after bringing algod up

pyteal/ast/router.py
```diff
@@ -256,26 +259,247 @@ def approval_construction(self) -> Optional[Expr]:
             )
         return Cond(*[[n.condition, n.branch] for n in conditions_n_branches])
 
+    def get_method_config(self) -> MethodConfig:
```
tzaffi, line 262: Introduced for `abi_router_test.py` integration test

pyteal/ast/router.py
```diff
@@ -715,12 +999,31 @@ def build_program(self) -> tuple[Expr, Expr, sdk_abi.Contract]:
...
+            bare_call_approval = self.bare_call_actions.approval_construction()
+            if bare_call_approval:
+                self.approval_ast.bare_calls = [
+                    CondNode(
+                        Txn.application_args.length() == Int(0),
+                        cast(Expr, bare_call_approval),
+                    )
+                ]
+
+        optimize = optimize if optimize else OptimizeOptions()
```
tzaffi, line 1012:
```suggestion
        optimize = optimize or OptimizeOptions()
```

ignoreable suggestion

tests/integration/abi_router_test.py
```diff
@@ -0,0 +1,540 @@
...
+    )
+    assert (sim_results := results.results) and all(
+        sim.succeeded for meth in sim_results.values() for sim in meth.values()
+    )
+    scenario_assert_stats(scenario, results, totals)
+
+    print("SUMMARY STATS: ", json.dumps(totals, indent=2))
+
+
+IDENTITY_PREDICATES = {
```
tzaffi, line 429: Graviton tests for asserting that v6 router is identical to v8 router

tzaffi (approved):
All my suggestions are small nits, so I am approving again.

pyteal/ast/router.py
ahangsu: yes, nice move

tzaffi:
A question came up whose answer is probably relevant to the work in this PR. I observed that when I introduced the following lines of code https://github.com/algorand/pyteal/blob/a73f4df2ebe553282ba2ba86e85d3a55ed65ab90/pyteal/stack_frame.py#L411-L414
the resulting teal actually changed. I suspect this has something to do with the caching behavior of the subroutine declarations, but I haven't investigated it very deeply.

tzaffi:
> A question came up whose answer is probably relevant to the work in this PR. I observed that when I introduced the following lines of code
> 
> https://github.com/algorand/pyteal/blob/a73f4df2ebe553282ba2ba86e85d3a55ed65ab90/pyteal/stack_frame.py#L411-L414
> 
> 
> the resulting teal actually changed. I suspect this has something to do with the caching behavior of the subroutine declarations, but I haven't investigated it very deeply.

I have a hunch that delaying `Router._build_program()` to happen during compilation may solve this problem as with all the compilation options in possession, there is no need to "hedge our bets" regarding whether to frame-pointers need to be utilized.

pyteal/ast/router.py
```diff
@@ -715,12 +1000,31 @@ def build_program(self) -> tuple[Expr, Expr, sdk_abi.Contract]:
...
             self.clear_state,
             self.contract_construct(),
         )
 
+    @contextmanager
+    def _cleaning_context(self):
+        starting_slot_id = ScratchSlot.nextSlotId
+        yield
+        self._clean()
+        ScratchSlot.reset_slot_numbering(starting_slot_id)
```
tzaffi, line 1026:
```suggestion
        try:
            yield
        finally:
            self._clean()
            ScratchSlot.reset_slot_numbering(starting_slot_id)
```

After playing around with my own context manager, I believe this pattern may be safer, in case of an exception during the compilation.

ahangsu: thanks for the suggestion!

jasonpaulos (approved):
Expected test code in `tests/teal/router` look correct to me

Issue #599, opened 2022-11-08
Improve subroutine evaluation code (post frame pointers)


jasonpaulos:
This issue deals with code implemented in #584 and #562.

In those PRs, to probe info for `has_return` and `type_of` of subroutine declaration, subroutines are evaluated twice: once with ScratchVar arguments, and once with frame pointer arguments. Then during compilation the appropriate version of subroutine evaulator is chosen and used in code generation.

Generally speaking it would be nice to avoid having to evaluate every subroutine twice. Once way of doing that could be with something like a `ProxyExpr`:

```python
class ProxyExpr(Expr):
    
    def __init__(self, type_of: TealType):
        super().__init__()
        self.representing: Expr | None = None
    
    def resolve_to(self, expr: Expr):
        self.representing = expr
    
    def __teal__(self, options: "CompileOptions"):
        if self.representing is None:
            raise TealInternalError("Not assigned....")
        return self.representing.__teal__(options)
```

The idea would be to use this as a placeholder before compilation to evaluate the subroutine, then assign the `ProxyExpr` to its actual value once that's known at compile time.

PR #606, opened 2022-11-18
Using local variable in subroutine through frame pointer

### LINKED ISSUE #559: Support local variables with frame pointers
## Problem

Local variables are implemented using scratch slots because it's been impractical to manage stack-based allocations in PyTeal.  With the introduction of frame pointer opcodes (https://github.com/algorand/go-algorand/pull/4319), there's less friction.  

The story requests defining + implementing a local variable scheme based on stack allocated values.  It's understood that the API must be fleshed out during development.

ahangsu:
closes #559

### DIFF
--- a/pyteal/ast/abi/type.py
+++ b/pyteal/ast/abi/type.py
@@ -3,7 +3,7 @@ from abc import ABC, abstractmethod
 
 from pyteal.ast.expr import Expr
 from pyteal.ast.abstractvar import AbstractVar
-from pyteal.ast.frame import FrameVar, FRAME_LOCAL_NUM
+from pyteal.ast.frame import FrameVar, MAX_FRAME_LOCAL_VARS
 from pyteal.ast.scratchvar import ScratchVar
 from pyteal.ast.seq import Seq
 from pyteal.errors import TealInputError
@@ -92,7 +92,7 @@ class BaseType(ABC):
             # NOTE: you can have at most 128 local variables.
             # len(local_types) + 1 computes the resulting length,
             # should be <= 128
-            if len(local_types) + 1 <= FRAME_LOCAL_NUM:
+            if len(local_types) + 1 <= MAX_FRAME_LOCAL_VARS:
                 local_types.append(self._type_spec.storage_type())
                 self._stored_value = FrameVar(proto, len(local_types) - 1)
                 return
--- a/pyteal/ast/abi/type.py
+++ b/pyteal/ast/abi/type.py
@@ -83,8 +83,7 @@ class BaseType(ABC):
         self._type_spec: Final[TypeSpec] = spec
         self._stored_value: AbstractVar
 
-        if SubroutineEval.Context.config_use_frame_pointer:
-            assert SubroutineEval.Context.proto
+        if SubroutineEval.Context.proto:
             proto = SubroutineEval.Context.proto
 
             assert proto.mem_layout
--- a/pyteal/ast/abi/type.py
+++ b/pyteal/ast/abi/type.py
@@ -3,6 +3,7 @@ from abc import ABC, abstractmethod
 
 from pyteal.ast.expr import Expr
 from pyteal.ast.abstractvar import AbstractVar
+from pyteal.ast.frame import FrameVar
 from pyteal.ast.scratchvar import ScratchVar
 from pyteal.ast.seq import Seq
 from pyteal.errors import TealInputError
@@ -76,9 +77,28 @@ class BaseType(ABC):
 
     def __init__(self, spec: TypeSpec) -> None:
         """Create a new BaseType."""
+        from pyteal.ast.subroutine import SubroutineEval
+
         super().__init__()
         self._type_spec: Final[TypeSpec] = spec
-        self._stored_value: AbstractVar = ScratchVar(spec.storage_type())
+        self._stored_value: AbstractVar
+
+        if SubroutineEval.Context.config_use_frame_pointer:
+            assert SubroutineEval.Context.proto
+            proto = SubroutineEval.Context.proto
+
+            assert proto.mem_layout
+            local_types = proto.mem_layout.local_stack_types
+
+            # NOTE: you can have at most 128 local variables.
+            # len(local_types) + 1 computes the resulting length,
+            # should be <= 128
+            if len(local_types) + 1 <= 128:
+                local_types.append(self._type_spec.storage_type())
+                self._stored_value = FrameVar(proto, len(local_types) - 1)
+                return
+
+        self._stored_value = ScratchVar(spec.storage_type())
 
     def type_spec(self) -> TypeSpec:
         """Get the TypeSpec for this ABI type instance."""
@@ -221,17 +241,23 @@ class ReturnedValue(ComputedValue):
                 f"expected type_spec {self.produced_type_spec()} but get {output.type_spec()}"
             )
 
-        declaration = self.computation.subroutine.get_declaration()
+        # HANG NOTE! This get_declaration check applies only for pre frame pointer case
+        # the post frame pointer case should not apply
+        # need to somehow expose the context of evaluation
+        try:
+            declaration = self.computation.subroutine.get_declaration()
 
-        if declaration.deferred_expr is None:
-            raise TealInputError(
-                "ABI return subroutine must have deferred_expr to be not-None."
-            )
-        if declaration.deferred_expr.type_of() != output.type_spec().storage_type():
-            raise TealInputError(
-                f"ABI return subroutine deferred_expr is expected to be typed {output.type_spec().storage_type()}, "
-                f"but has type {declaration.deferred_expr.type_of()}."
-            )
+            if declaration.deferred_expr is None:
+                raise TealInputError(
+                    "ABI return subroutine must have deferred_expr to be not-None."
+                )
+            if declaration.deferred_expr.type_of() != output.type_spec().storage_type():
+                raise TealInputError(
+                    f"ABI return subroutine deferred_expr is expected to be typed {output.type_spec().storage_type()}, "
+                    f"but has type {declaration.deferred_expr.type_of()}."
+                )
+        except Exception:
+            pass
 
         return output._stored_value.store(self.computation)
 
--- a/pyteal/ast/abi/type.py
+++ b/pyteal/ast/abi/type.py
@@ -4,7 +4,7 @@ from abc import ABC, abstractmethod
 from pyteal.ast.expr import Expr
 from pyteal.ast.abstractvar import AbstractVar, alloc_abstract_var
 from pyteal.ast.seq import Seq
-from pyteal.errors import TealInputError
+from pyteal.errors import TealInputError, TealTypeError
 from pyteal.types import TealType
 
 
@@ -215,6 +215,8 @@ class ReturnedValue(ComputedValue):
         return self.type_spec
 
     def store_into(self, output: BaseType) -> Expr:
+        from pyteal.ast.subroutine import SubroutineDeclaration
+
         if output.type_spec() != self.produced_type_spec():
             raise TealInputError(
                 f"expected type_spec {self.produced_type_spec()} but get {output.type_spec()}"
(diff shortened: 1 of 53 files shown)


ahangsu:
Maybe we want to handle https://github.com/algorand/pyteal/pull/562#discussion_r1026086697 in this PR

pyteal/ast/abi/type.py
```diff
@@ -221,17 +240,23 @@ def store_into(self, output: BaseType) -> Expr:
...
-        if declaration.deferred_expr.type_of() != output.type_spec().storage_type():
-            raise TealInputError(
-                f"ABI return subroutine deferred_expr is expected to be typed {output.type_spec().storage_type()}, "
-                f"but has type {declaration.deferred_expr.type_of()}."
-            )
+        # HANG NOTE! This get_declaration check applies only for pre frame pointer case
+        # the post frame pointer case should not apply
+        # need to somehow expose the context of evaluation
+        try:
+            declaration = self.computation.subroutine.get_declaration()
```
ahangsu, line 247: this line is somewhat off: the reason is that, we are not *version aware* when we call `store_into`. This `get_declaration` always return version 6 implementation, while in version 8 this would cause error.

pyteal/ast/subroutine.py
```diff
@@ -1025,6 +1036,14 @@ def __call__(self, subroutine: SubroutineDefinition) -> SubroutineDeclaration:
             if not self.use_frame_pt:
                 deferred_expr = output_carrying_abi._stored_value.load()
 
+        if self.use_frame_pt:
+            assert proto.mem_layout
+            depth = len(proto.mem_layout.local_stack_types)
+            # only when we have 1 return, and with other local variables
+            # we use bury to bury the result to 0 index against frame pointer
+            if not abi_output_kwargs and 0 < proto.num_returns < depth:
+                deferred_expr = Bury(depth)
```
tzaffi, line 1045: Could `FrameBury` be used here instead by using the offset from the frame pointer? I ask because it appears that this is the only usage of `Bury` so if the answer is "yes", it's an argument against introducing `class Bury` and sticking to `FrameBury` for frame pointer related work.

pyteal/ast/subroutine.py
```diff
@@ -1006,7 +1007,17 @@ def __call__(self, subroutine: SubroutineDefinition) -> SubroutineDeclaration:
...
 
         # Arg usage "B" supplied to build an AST from the user-defined PyTEAL function:
-        subroutine_body = subroutine.implementation(*loaded_args, **abi_output_kwargs)
+        subroutine_body: Expr
+        if not self.use_frame_pt:
+            subroutine_body = subroutine.implementation(
+                *loaded_args, **abi_output_kwargs
+            )
+        else:
+            with SubroutineEval.Context.CompileWithFrameContext(proto):
```
tzaffi, line 1016: Is this a lot cleaner than using a more direct approach that avoids python contexts? As cool as I find contexts to be, I think they're most useful for consumers of an API. In library code, they may cause unnecessary confusion.

tzaffi (commented):
Another great improvement in our war against scratch slots. I left a couple of questions.

pyteal/ast/frame.py
```diff
@@ -13,6 +13,9 @@
     from pyteal.compiler import CompileOptions
 
 
+FRAME_LOCAL_NUM: Final[int] = 128
```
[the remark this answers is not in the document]
ahangsu: 👍 makes sense, I will add to my changes

pyteal/ast/abi/type.py
```diff
@@ -76,9 +77,27 @@ class BaseType(ABC):
...
+
         super().__init__()
         self._type_spec: Final[TypeSpec] = spec
-        self._stored_value: AbstractVar = ScratchVar(spec.storage_type())
+        self._stored_value: AbstractVar
+
+        if SubroutineEval.Context.proto:
+            proto = SubroutineEval.Context.proto
+
+            assert proto.mem_layout
```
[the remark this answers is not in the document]
ahangsu: Another good point, I have to admit that letting `proto.mem_layout` to be okay with `None` is not so well considered...

> Perhaps the proto member of Context object is constructed in a way that guarantees that it's well-formed?

yes, that is correct, for the proto generated in subroutine is guaranteed to be well-formed. But still, I think it might be better to let `proto.mem_layout` to be not `None`.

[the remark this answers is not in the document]
ahangsu: Pushed the change in [06774f6](https://github.com/algorand/pyteal/pull/606/commits/06774f675f208f5085760014ed20929a9bd7129d)

pyteal/ast/subroutine.py
ahangsu: I see, let me see if there are some easy syntax to achieve same functionality of swapping scratch-based ABI locals to frame-based ABI locals. I was unaware of the purpose of context, and just make things easiest for me at this place.

ahangsu: I figured out a way to get away with context syntax in this library internal implementation, pushed here: [7514b14](https://github.com/algorand/pyteal/pull/606/commits/7514b14397978e881e2354e014e4f9d5ca51cee6).

pyteal/ast/subroutine.py
```diff
@@ -1006,7 +1007,19 @@ def __call__(self, subroutine: SubroutineDefinition) -> SubroutineDeclaration:
...
         # Arg usage "B" supplied to build an AST from the user-defined PyTEAL function:
-        subroutine_body = subroutine.implementation(*loaded_args, **abi_output_kwargs)
+        subroutine_body: Expr
+        if not self.use_frame_pt:
+            subroutine_body = subroutine.implementation(
+                *loaded_args, **abi_output_kwargs
+            )
+        else:
+            prev_proto: Optional[Proto] = SubroutineEval.current_proto
+            SubroutineEval.current_proto = proto
```
tzaffi, line 1017: Now that you have followed my suggestion and removed the python context-manager I see better the original purpose. We need a way to transmit to ABI Types that they are in a particular proto context and then use the frame opcodes accordingly, instead of scratch var's. Since we want the ABI Types to detect the proto context without explicitly being told about this, we need to pass this information through a global (or static) variable. Without a python context-manager, we are left to directly manipulate the static variable before and after subroutine evaluation, as is happening here. However, this manual manipulation is exactly what context-managers are meant to relieve us from.
So I must backtrack from my original suggestion and suggest that context managers be employed. I would accept reverting to the original code.
However, my original remarks about context managers being confusing still stand. So I wonder if there is a way to make the intent simpler. Perhaps it's just a matter of variable naming.

The original usage pattern looked like:
```python
with SubroutineEval.Context.CompileWithFrameContext(proto):
    subroutine_body = subroutine.implementation(...)
```

Perhaps the following would be easier to grok?

```python
with frame_pointer_context(proto):
    ...
```
The implementation might look like the following (which I basically cribbed from [this RealPython article](https://realpython.com/python-with-statement/#creating-function-based-context-managers) ) but you could also using a class-based context and keep the running `proto` as a static variable, instead of the global that I'm using:

```python
from contextlib import contextmanager

_fp_context_proto: Optional[Proto] = None

@contextmanager
def frame_pointer_context(proto: Proto):
    global _fp_context_proto
    tmp, _fp_context_proto = _fp_context_proto, proto
    yield proto # might as well yield the proto
    _fp_context_proto = tmp
```


But again, reverting to the original implementation is good 'nuf. 

And apologies for asking you to revert the exact thing I previously asked you to do.

tzaffi (commented):
Upon closer evaluation, I think the original approach using python context-managers is preferred.

Apart from that, there was only one question remaining: how blocked is the PR by #599 ?

pyteal/ast/subroutine.py
```diff
@@ -1025,6 +1044,13 @@ def __call__(self, subroutine: SubroutineDefinition) -> SubroutineDeclaration:
             if not self.use_frame_pt:
                 deferred_expr = output_carrying_abi._stored_value.load()
 
+        if self.use_frame_pt:
+            depth = len(proto.mem_layout.local_stack_types)
+            # only when we have 1 return, and with other local variables
+            # we use bury to bury the result to 0 index against frame pointer
```
ahangsu, line 1050:
```suggestion
            # we use frame_bury to bury the result to 0 index against frame pointer
```

tzaffi (approved):
Approving as all my major remarks have been addressed.

pyteal/ast/abi/type.py
```diff
@@ -221,17 +240,23 @@ def store_into(self, output: BaseType) -> Expr:
...
-                "ABI return subroutine must have deferred_expr to be not-None."
-            )
-        if declaration.deferred_expr.type_of() != output.type_spec().storage_type():
-            raise TealInputError(
-                f"ABI return subroutine deferred_expr is expected to be typed {output.type_spec().storage_type()}, "
-                f"but has type {declaration.deferred_expr.type_of()}."
-            )
+        # HANG NOTE! This get_declaration check applies only for pre frame pointer case
+        # the post frame pointer case should not apply
+        # need to somehow expose the context of evaluation
```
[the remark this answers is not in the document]
ahangsu: no longer a blocker I think, changed my mind

pyteal/ast/abi/type.py
ahangsu: For status quo, we cannot know compile to version while evaluating the pyteal source code. We have to use `get_declaration_by_option(False)` here, and we can deprecate `get_declaration` methods.

pyteal/ast/subroutine.py
```diff
@@ -304,6 +306,7 @@ def _validate_annotation(
         )
 
     def get_declaration(self) -> "SubroutineDeclaration":
+        """MARK AS DEPRECATED, FOR WE ARE NOW GET DECLARATION BY COMPILE OPTIONS"""
```
michaeldiamant, line 309: @ahangsu Checking if it's been considered:  Since the docstring mentions deprecation, have you also considered adding a `warnings.warn(..., ..., DeprecationWarning`) to formally announce deprecation? 

Presumably, adding a formal deprecation paves the way for us to remove these methods in a future release.

pyteal/ast/abi/type.py
```diff
@@ -221,17 +220,23 @@ def store_into(self, output: BaseType) -> Expr:
...
+                raise TealInputError(
+                    "ABI return subroutine must have deferred_expr to be not-None."
+                )
+            if declaration.deferred_expr.type_of() != output.type_spec().storage_type():
+                raise TealInputError(
+                    f"ABI return subroutine deferred_expr is expected to be typed {output.type_spec().storage_type()}, "
+                    f"but has type {declaration.deferred_expr.type_of()}."
+                )
+        except Exception:
+            pass
```
michaeldiamant, line 239: @ahangsu Request for clarification:  For scratch slot-based usage, is it _correct_ to ignore the raised exception?

I think I'm missing some intuition.  I'm imagining if an exception was thrown prior to the PR (with scratch slots), then it's _still_ an error to raise to the user.

Feel welcomed to suggest a live discussion if there's too much context to convey in writing.

pyteal/ast/abstractvar.py
```diff
@@ -38,3 +38,32 @@ def storage_type(self) -> TealType:
...
+def alloc_abstract_var(stack_type: TealType) -> AbstractVar:
+    """Allocate abstract var over stack, or over scratch.
+
+    This unexported function takes a TealType as value type representation over stack (or scratch),
+    and generates an AbstractVar instance.
+    It infers the proto currently being used in context of subroutine evaluation,
+    and swap to FrameVar to save the use of scratch slots.
+
+    Arg:
+        stack_type: An TealType that represents stack type.
```
michaeldiamant, line 52: @ahangsu Minor verbiage suggestion.  Alternatively, changing from _An_ to _A_ works too.

```suggestion
        stack_type: TealType that represents stack type.
```

pyteal/ast/subroutine.py
ahangsu: That makes sense, just like py-sdk.

pyteal/ast/abi/type.py
ahangsu: Refactored in https://github.com/algorand/pyteal/pull/606/commits/0f1fe4f384c0a829c3aa9188bd548f54b16e4821, for the only error throwing is running `get_declaration_by_option`, with no access to current compile option, such that compiling against scratch slot with a evaluated to frame pt body will error.

I refactored out the coarse scope of error catching in latest commit.

pyteal/ast/subroutine.py
```diff
@@ -304,6 +311,10 @@ def _validate_annotation(
         )
 
     def get_declaration(self) -> "SubroutineDeclaration":
+        warnings.warn(
+            "`get_declaration` is being deprecated: Please use `get_declaration_by_option` instead.",
+            DeprecationWarning,
+        )
         return self.declarations.get_declaration()
```
michaeldiamant, line 318: @ahangsu Minor:  Since `self.declarations.get_declaration()` is deprecated, I'm wondering if it's possible to change the implementation to a non-deprecated variant?

I think `return self.get_declaration_by_option()` is also a possible option here.

```suggestion
        return self.declarations.get_declaration_by_option(False)
```

ahangsu: good point, I take the change, or API caller will take double warning.

jasonpaulos (approved):

Issue #608, opened 2022-11-30
PyTeal examples folder ships with the package itself


CiottiGiorgio:
Not sure if this was intended or not but I think the examples shouldn't be shipped with the package.
I observed the same behaviour in `beaker` and it is currently being addressed by [this](https://github.com/algorand-devrel/beaker/issues/132) issue.

jasonpaulos:
Hi @CiottiGiorgio, why do you think examples should not ship with the package?

CiottiGiorgio:
I think it can be nice if you are working directly with PyTeal but if I'm just importing it as a dependency (of Beaker for example) I don't need PyTeal examples.
As it is named right now, it's just an `example` folder within `site-packages` and it could collide with other examples from other packages (should they also ship them).
The best place is probably in a separate python package and then listed in the dev dependencies of pyteal (alongside testing, typing and formatting utilities that are currently written in `requirements.txt` and not `install_requires`).

At the very least, the folder should be called `pyteal_examples` so it does not collide with other dependency in the project.

CiottiGiorgio:
While I was figuring out the structure of the project and how it installs dev/non-dev dependencies, I found found out that we have the same issue for `tests` in `PyTeal` package and `tests` in `py-algorand-sdk` package. Both `tests` folders end up in `site-packages` colliding with each other.
Personally, I would set `packages` into `setup.py` to just `packages=["pyteal"]` and do the same for `py-algorand-sdk`.

CiottiGiorgio:
#610 is my proposal

CiottiGiorgio:
#610 was merged so we can close this.

michaeldiamant:
For completeness - A bug in #610 was corrected via https://github.com/algorand/pyteal/pull/614.

PR #610, opened 2022-12-03
Proposal to refactor examples name and exclude tests from being shipped.

CiottiGiorgio:


jasonpaulos (commented):
Thanks for the explanation in #608 and for creating this PR to remedy the situation.

I wasn't aware that the examples folder was being installed as a standalone package -- I think this behavior is not intentional nor desired.

So instead of renaming, could we keep the folder as `examples`, but change `setup.py` to only include the `pyteal` package?

ahangsu:
^ Up Jason's point, I think it would be nice to not involve too much change in package structure itself, but rather specifying the only package data is from `pyteal` directory, by specifying in `setup.py`.

CiottiGiorgio:
Sounds good. I'll modify this PR

ahangsu (approved):
lgtm

CiottiGiorgio:
I can't merge this as I don't have write privileges

jasonpaulos (approved):

PR #613, opened 2022-12-08
`OptimizeOptions` Takes Over `frame_pointers` Option

tzaffi:
# Summary

Removing the `frame_pointers` parameter in `Router.compile_program()` and `CompileOptions.__init__()` and adding it to `OptimizeOptions`

### DIFF
--- a/pyteal/compiler/compiler.py
+++ b/pyteal/compiler/compiler.py
@@ -51,7 +51,7 @@ class CompileOptions:
         self.mode: Mode = mode
         self.version: int = version
         self.optimize: OptimizeOptions = optimize or OptimizeOptions()
-        self.use_frame_pointer: bool = self.optimize.use_frame_pointers(self.version)
+        self.use_frame_pointers: bool = self.optimize.use_frame_pointers(self.version)
 
         self.currentSubroutine: Optional[SubroutineDefinition] = None
 
@@ -158,14 +158,14 @@ def compileSubroutine(
     if (
         currentSubroutine
         and currentSubroutine.get_declaration_by_option(
-            options.use_frame_pointer
+            options.use_frame_pointers
         ).deferred_expr
     ):
         # this represents code that should be inserted before each retsub op
         deferred_expr = cast(
             Expr,
             currentSubroutine.get_declaration_by_option(
-                options.use_frame_pointer
+                options.use_frame_pointers
             ).deferred_expr,
         )
 
@@ -218,7 +218,7 @@ def compileSubroutine(
     newSubroutines = referencedSubroutines - subroutine_start_blocks.keys()
     for subroutine in sorted(newSubroutines, key=lambda subroutine: subroutine.id):
         compileSubroutine(
-            subroutine.get_declaration_by_option(options.use_frame_pointer),
+            subroutine.get_declaration_by_option(options.use_frame_pointers),
             options,
             subroutineGraph,
             subroutine_start_blocks,
--- a/pyteal/compiler/compiler.py
+++ b/pyteal/compiler/compiler.py
@@ -46,22 +46,20 @@ class CompileOptions:
         mode: Mode = Mode.Signature,
         version: int = DEFAULT_PROGRAM_VERSION,
         optimize: OptimizeOptions = None,
-        frame_pointers: Optional[bool] = None,
     ) -> None:
         self.mode = mode
         self.version = version
-        self.optimize = optimize if optimize is not None else OptimizeOptions()
+        self.optimize = (
+            optimize if optimize else OptimizeOptions.default_for_version(self.version)
+        )
         self.use_frame_pointer: bool
 
-        if frame_pointers is None:
-            self.use_frame_pointer = self.version >= FRAME_POINTER_VERSION
+        if self.optimize.frame_pointers and self.version < FRAME_POINTER_VERSION:
+            raise TealInputError(
+                f"Try to use frame pointer with an insufficient version {self.version}."
+            )
         else:
-            if frame_pointers and self.version < FRAME_POINTER_VERSION:
-                raise TealInputError(
-                    f"Try to use frame pointer with an insufficient version {self.version}."
-                )
-            else:
-                self.use_frame_pointer = frame_pointers
+            self.use_frame_pointer = self.optimize.frame_pointers
 
         self.currentSubroutine: Optional[SubroutineDefinition] = None
 
@@ -257,7 +255,6 @@ def compileTeal(
     version: int = DEFAULT_PROGRAM_VERSION,
     assembleConstants: bool = False,
     optimize: OptimizeOptions = None,
-    frame_pointers: Optional[bool] = None,
 ) -> str:
     """Compile a PyTeal expression into TEAL assembly.
 
@@ -291,9 +288,7 @@ def compileTeal(
             )
         )
 
-    options = CompileOptions(
-        mode=mode, version=version, optimize=optimize, frame_pointers=frame_pointers
-    )
+    options = CompileOptions(mode=mode, version=version, optimize=optimize)
 
     subroutineGraph: Dict[SubroutineDefinition, Set[SubroutineDefinition]] = dict()
     subroutine_start_blocks: Dict[Optional[SubroutineDefinition], TealBlock] = dict()
--- a/pyteal/compiler/compiler.py
+++ b/pyteal/compiler/compiler.py
@@ -26,7 +26,7 @@ from pyteal.compiler.subroutines import (
 from pyteal.compiler.constants import createConstantBlocks
 
 MAX_PROGRAM_VERSION = 8
-FRAME_POINTER_VERSION = 8
+FRAME_POINTERS_VERSION = 8
 DEFAULT_SCRATCH_SLOT_OPTIMIZE_VERSION = 9
 MIN_PROGRAM_VERSION = 2
 DEFAULT_PROGRAM_VERSION = MIN_PROGRAM_VERSION
--- a/pyteal/compiler/compiler.py
+++ b/pyteal/compiler/compiler.py
@@ -246,7 +246,7 @@ def compileTeal(
     *,
     version: int = DEFAULT_PROGRAM_VERSION,
     assembleConstants: bool = False,
-    optimize: OptimizeOptions = None,
+    optimize: Optional[OptimizeOptions] = None,
 ) -> str:
     """Compile a PyTeal expression into TEAL assembly.
 
--- a/pyteal/compiler/compiler.py
+++ b/pyteal/compiler/compiler.py
@@ -27,6 +27,7 @@ from pyteal.compiler.constants import createConstantBlocks
 
 MAX_PROGRAM_VERSION = 8
 FRAME_POINTER_VERSION = 8
+DEFAULT_SCRATCH_SLOT_OPTIMIZE_VERSION = 9
 MIN_PROGRAM_VERSION = 2
 DEFAULT_PROGRAM_VERSION = MIN_PROGRAM_VERSION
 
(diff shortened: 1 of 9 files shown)


pyteal/compiler/optimizer/optimizer.py
```diff
@@ -22,12 +23,28 @@ class OptimizeOptions:
 
         scratch_slots (optional): cancel contiguous store/load operations
             that have no load dependencies elsewhere.
+        frame_opinters
```
tzaffi, line 26:
```suggestion
```

pyteal/compiler/optimizer/optimizer.py
```diff
@@ -24,10 +25,25 @@ class OptimizeOptions:
...
+    def maximize(cls) -> "OptimizeOptions":
+        return cls(scratch_slots=True, frame_pointers=True)
+
+    @classmethod
+    def default_for_version(cls, version: int) -> "OptimizeOptions":
+        """TODO: consider maximizing for default starting with Teal v. 9"""
+        from pyteal.compiler.compiler import FRAME_POINTER_VERSION
+
+        if version < FRAME_POINTER_VERSION:
+            return cls(scratch_slots=False, frame_pointers=False)
```
tzaffi, line 43: Is it too late to make the default behavior for v8 use frame pointers? If so, then I'm afraid that I'll have to default `frame_pointers = False` in the case of `version $\geq$ FRAME_POINTER_VERSION`.

Also, should I already aggressively force `maximize` for v9 and higher?

ahangsu: It might be okay to make default v8 use of frame pointer, haven't thought through the consequences, simply feeling okay with me.

I am slightly against `maximize` for v9 and higher, for it might make debugging somehow harder?

ahangsu (commented):
generally lgtm, just one question on configuring `default_for_versions` to be settled down.

tzaffi:
So you are advocating for _not_ defaulting to scratch_slots = True staring
with v9?

ahangsu:
Correct, I am advocating that scratch_slots = True not starting with v9 as
default.

pyteal/compiler/optimizer/optimizer.py
jasonpaulos: Default use of frame pointers is v8+ code is good with me.

I do not think we should default to `maximize` for any version at the moment. The risk of backwards incompatibility when new optimizations are introduced is too great to make this the default behavior in my opinion.

pyteal/compiler/optimizer/optimizer.py
```diff
@@ -24,10 +25,25 @@ class OptimizeOptions:
             that have no load dependencies elsewhere.
     """
 
-    def __init__(self, *, scratch_slots: bool = False):
-        self.scratch_slots = scratch_slots
+    def __init__(self, *, scratch_slots: bool = False, frame_pointers: bool = False):
```
jasonpaulos, line 28: I think it would be a reasonable property to make using `None` or using `OptimizeOptions()` when creating a `CompileOptions()` behave identically.

Right now they won't because `OptimizeOptions()` _always_ has `frame_pointers=False`.

This is also a problem because existing code is likely to use `OptimizeOptions(scratch_slots=True)`, and it would be undesirable to force such code to have `frame_pointers=False`.

pyteal/compiler/optimizer/optimizer.py
```diff
@@ -24,10 +25,25 @@ class OptimizeOptions:
...
+        self.scratch_slots: Final[bool] = scratch_slots
+        self.frame_pointers: Final[bool] = frame_pointers
         self._skip_slots: Set[ScratchSlot] = set()
 
+    @classmethod
+    def maximize(cls) -> "OptimizeOptions":
+        return cls(scratch_slots=True, frame_pointers=True)
```
jasonpaulos, line 35: Here's a potential problem with maximize: v7 and lower programs can't use it because they don't support frame pointers.

If we want this method to be useful, it should enable every optimization that's possible for every program version.

jasonpaulos (commented):
Here's a suggestion to avoid a problem we are grappling with in the JS SDK:

Generally speaking, when creating a class with complex configuration (`CompileOptions` in this case) we want constructor argument types to be as convenient and flexible as possible, yet we want the class members to be as specific and unambiguous as possible, so that the implementation of the class and places where it's used are simple.

If you attempt to use the same types as constructor arguments and class fields, it's difficult to satisfy both of those properties. So here's my suggestion: make the argument types (`OptimizeOptions` in this case) capture what the user wants, and only inside of the class (`CompileOptions`), you transform this information into what actually needs to be done.

For frame pointers, I see it like this:

| What the user wants  | What actually needs to be done |
| ------------- | ------------- |
| Enabled  | Use frame pointers  |
| Disabled  | Don't use frame pointers  |
| Enabled if available (this is for `maximize` to work) | Use frame pointers only if v8 or higher |
| Default | Use frame pointers only if v8 or higher |

In this case "Enabled if available" happens to be the same as "Default", but this isn't necessarily true, and I still think it's valuable to differentiate between the two.

pyteal/compiler/optimizer/optimizer.py
tzaffi: Removing for now

pyteal/compiler/optimizer/optimizer.py
tzaffi: Reverted to `None` having the same behavior as `OptimizeOptions()`.  Also, introducing `use_frame_pointer(version)` and `optimize_scratch_slots(version)` to provide transparent behavior as well as be backwards compatible to <= v7 compilationn.

pyteal/compiler/optimizer/optimizer.py
tzaffi: I've removed `maximize()`. However, please note that as the PR currently stands, I'm actually defaulting to `optimize_scratch_slots() == True` starting with v9. This is open to discussion.

tzaffi:
> Here's a suggestion to avoid a problem we are grappling with in the JS SDK:
> ...

Under the current state, we have the following behaviors. Once we come to consensue, I will also modify [the optimize options docs](https://pyteal.readthedocs.io/en/stable/compiler_optimization.html) accordingly.

| Option | What the user specifies  | Version | What is actually done |
| ------------- | ------------- |
| `scratch_slots` | Enabled (`True`)  | _any_ | Optimize scratch slots  |
|   | Disabled (`False`)  | _any_ | **Do not** optimize scratch slots  |
|   | Default (`None`)  | $\leq 8$ | **Do not** optimize scratch slots  |
|   | Default (`None`)  | $\geq 9$ | Optimize scratch slots  |
| `frame_pointers` | Enabled (`True`)  | $\leq 8$ | `TealInputError` during compilation  |
|  | Disabled (`False`)  | $\leq 8$ | Don't use frame pointers |
|  | Enabled (`False`)  | $\geq 8$ | Use frame pointers |
|  | Default (`None`)  | $\leq 8$ | Don't use frame pointers |
|  | Default (`None`)  | $\geq 8$ | Use frame pointers |

pyteal/compiler/compiler.py
```diff
@@ -27,6 +27,7 @@
 
 MAX_PROGRAM_VERSION = 8
 FRAME_POINTER_VERSION = 8
+DEFAULT_SCRATCH_SLOT_OPTIMIZE_VERSION = 9
```
tzaffi, line 30: This value is open to discussion.

pyteal/compiler/compiler.py
```diff
@@ -307,12 +294,12 @@ def compileTeal(
...
     # is necessary for the dependency checking of local slots. Global slots, slots
     # used by DynamicScratchVar, and reserved slots are not optimized.
-    if options.optimize.scratch_slots:
+    if options.optimize.optimize_scratch_slots(version):
         options.optimize._skip_slots = collect_unoptimized_slots(
             subroutine_start_blocks
         )
         for start in subroutine_start_blocks.values():
-            apply_global_optimizations(start, options.optimize)
+            apply_global_optimizations(start, options.optimize, version)
```
tzaffi, line 302: We need to start passing `version` since `optimize.optimize_scratch_slots()` is now requiring version.

pyteal/compiler/optimizer/optimizer.py
```diff
@@ -24,10 +25,43 @@ class OptimizeOptions:
...
+        *,
+        scratch_slots: Optional[bool] = None,
+        frame_pointers: Optional[bool] = None,
+    ):
+        self._scratch_slots: Final[Optional[bool]] = scratch_slots
+        self._frame_pointers: Final[Optional[bool]] = frame_pointers
+
         self._skip_slots: Set[ScratchSlot] = set()
 
+        # deprecated field kept for backwards compatibility - not used. Shall I remove altogether?
```
tzaffi, line 39: Feedback appreciated

pyteal/compiler/optimizer/optimizer.py
```diff
@@ -87,13 +121,15 @@ def _apply_slot_to_stack(
     _remove_extraneous_slot_access(start, slots_to_remove)
 
 
-def apply_global_optimizations(start: TealBlock, options: OptimizeOptions) -> TealBlock:
+def apply_global_optimizations(
```
tzaffi, line 124: This is a breaking change, but I'm comfortable with it because `apply_global_optimization` is internal

pyteal/compiler/compiler.py
```diff
@@ -46,22 +47,11 @@ def __init__(
         mode: Mode = Mode.Signature,
         version: int = DEFAULT_PROGRAM_VERSION,
         optimize: OptimizeOptions = None,
```
tzaffi, line 49:
```suggestion
        optimize: Optional[OptimizeOptions] = None,
```

tzaffi:
> Correct, I am advocating that scratch_slots = True not starting with v9 as default. On Thu, Dec 8, 2022 at 12:22 PM Zeph Grunschlag ***@***.***> wrote:

FYI - The current state of the PR still proposes to default to `optimize_scratch_slots() == True` starting v9. It's easy enough to not do this however.

pyteal/compiler/optimizer/optimizer.py
ahangsu: I vote for removing it.

ahangsu (commented):
So far looks very close to what we expect! The only conflict is the `scratch_slot` optimization scheme, so long as we settle down to one agreement, I will approve.

pyteal/compiler/optimizer/optimizer.py
```diff
@@ -21,13 +22,48 @@ class OptimizeOptions:
     Args:
 
         scratch_slots (optional): cancel contiguous store/load operations
-            that have no load dependencies elsewhere.
+            that have no load dependencies elsewhere. Starting TEAL version 9, defaults to optimizing.
+        frame_pointers (optional): employ frame pointers instead of scratch slots during compilation.
+            Available only starting in version 8. Defaults to optimizing starting TEAL version 8.
```
michaeldiamant, line 27: @tzaffi I'm only seeking to comment on a minor verbiage concern.  We've moved towards referencing _program versions_/_AVM versions_ rather than _TEAL versions_.

```suggestion
            that have no load dependencies elsewhere. Starting with program version 9, defaults to optimizing.
        frame_pointers (optional): employ frame pointers instead of scratch slots during compilation.
            Available only starting in program version 8. Defaults to optimizing starting in program version 8.
```

pyteal/compiler/optimizer/optimizer.py
```diff
@@ -21,13 +22,48 @@ class OptimizeOptions:
...
+        from pyteal.compiler.compiler import FRAME_POINTERS_VERSION
+
+        if self._frame_pointers is None:
+            return version >= FRAME_POINTERS_VERSION
+
+        if self._frame_pointers:
+            verifyProgramVersion(
+                FRAME_POINTERS_VERSION,
+                version,
+                f"Frame pointers aren't available when compiling to TEAL version {version}",
```
michaeldiamant, line 62:
```suggestion
                f"Frame pointers aren't available when compiling to program version {version}",
```

pyteal/compiler/optimizer/optimizer.py
tzaffi: removed

ahangsu (approved):

PR #614, opened 2022-12-13
Correctly searching for subpackages only in the intended folder.

CiottiGiorgio:
Oopsie

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,7 @@ setuptools.setup(
     long_description=long_description,
     long_description_content_type="text/markdown",
     url="https://github.com/algorand/pyteal",
-    packages=setuptools.find_packages(include=('pyteal*',)),
+    packages=setuptools.find_packages(include=("pyteal*",)),
     install_requires=[
         # when changing this list, also update docs/requirements.txt
         "py-algorand-sdk>=1.9.0,<2.0.0",
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,7 @@ setuptools.setup(
     long_description=long_description,
     long_description_content_type="text/markdown",
     url="https://github.com/algorand/pyteal",
-    packages=["pyteal"],
+    packages=setuptools.find_packages(include=('pyteal*',)),
     install_requires=[
         # when changing this list, also update docs/requirements.txt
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -14,7 +14,7 @@
     long_description=long_description,
     long_description_content_type="text/markdown",
     url="https://github.com/algorand/pyteal",
-    packages=["pyteal"],
+    packages=setuptools.find_packages(include=("pyteal*",)),
```
michaeldiamant, line 17: @CiottiGiorgio Can you share how was the change tested?

I'm not immediately familiar with this part of Python's tooling ecosystem and I'm hoping you can accelerate my ability to validate the PR.

CiottiGiorgio: Yes sorry I didn't describe appropriately how the issue was found and how it was solved (or what the issue even is).

So by explicitly setting `packages` to `["pyteal"]`, I excluded all submodules.
I found this out because I needed a specific commit for PyTeal after the latest release. I just pulled from master and I realized that no submodule was being packaged.
With the help of , I changed the inclusion rule for the packaging, deleted my venv, started fresh and reinstalled with `pip install .`.
This seems to have fixes this issue while still only importing `pyteal` root package + submodules.

michaeldiamant, line 17: @CiottiGiorgio Optional - I think the PR is correct though I believe the following _suggestion_ is more specific.

As is, additional top-level packages (e.g. _pyteal-foo_) will be included.  Based on usage patterns, I _think_ additional top-level packages should opt-in to packaging.

```suggestion
    packages=setuptools.find_packages(include=("pyteal", "pyteal.*",)),
```

michaeldiamant: @CiottiGiorgio Resolving - Thanks for the explanation.  I reproduced the problem with these instructions.

Unfortunately, I didn't find a reasonable CI check to install during my several minute review.

CiottiGiorgio: Absolutely agree. Thanks!

jasonpaulos (approved):
@CiottiGiorgio thanks for finding this issue and submitting a fix. I tested locally and this looks like a good solution to me

PR #615, opened 2022-12-15
`OpType(Enum)` falls down on python 3.11.1

tzaffi:
# The Problem

There is [a regression on python 3.11.1](https://github.com/python/cpython/issues/100098) which breaks the way we're using `NamedTuple`s as an `OpType(Enum)` base type.
Even though it is expected that python 3.11.2 will fix the regression, there is a simple solution: swap in a dataclass instead.

## Testing
I'm also introducing python 3.11 workflows to [demonstrate the problem](https://github.com/algorand/pyteal/actions/runs/3708935444/jobs/6287006734#step:6:21536) and ensure confidence in the fix. I propose that we _keep these workflows in place_ to help detect future regressions. Please note that the example error occurred _without any code changes_, i.e. all I did was introduce the python 3.11.1 workflow. You can validate with this commit: https://github.com/algorand/pyteal/pull/615/commits/df9219719ceba76e8f1bcd5c097788c651e8b7c0

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,7 @@ jobs:
     container: python:${{ matrix.python }}
     strategy:
       matrix:
-        python: ["3.10"]
+        python: ["3.10", "3.11"]
     steps:
       - run: python3 --version
       - name: Check out code
@@ -29,7 +29,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        python: [ "3.10" ]
+        python: ["3.10", "3.11"]
     steps:
       - name: Check out code
         uses: actions/checkout@v2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * Introducing `AbstractVar` to abstract value access: store, load, and stack type. ([#584](https://github.com/algorand/pyteal/pull/584))
   * NOTE: a backwards incompatable change was imposed in this PR: previous ABI value's public member `stored_value` with type `ScratchVar`, is now changed to protected member `_stored_value` with type `AbstractVar`.
 * Starting with program version 9, when `scratch_slots` flag isn't provided to `OptimizeOptions`, default to optimizing. For versions 8 and earlier the default is and remains to _not_ optimize. ([#613](https://github.com/algorand/pyteal/pull/613))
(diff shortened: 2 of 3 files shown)


ahangsu (approved):
looks good, would be nice to add a line in `CHANGELOG`

tzaffi:
> looks good, would be nice to add a line in `CHANGELOG`

See the most recent commit for the CHANGELOG

ahangsu (approved):

bbroder-algo:
Hey this is my 5th minute looking at PyTeal and I can't really find any places that use the specific typing information that is/was provided by NamedTuple, but I did see some places that ask for the str() representation of a NamedTuple (for example, in tealop.py) and I verified that the str() for a dataclasses looks the same as a str() for NamedTuple). 
`>>> from typing import NamedTuple
>>> bob = NamedTuple ("BobType", [("value", str), ("mode", int)])
>>> a = bob("Vlaue", 3)
>>> from dataclasses import dataclass
>>> @dataclass
... class BobDC:
...   value:str
...   mode:int
... 
>>> b = BobDC ("vvalue", "4")
>>> print (str(b))
BobDC(value='vvalue', mode='4')
>>> print (str(a))
BobType(value='Vlaue', mode=3)
>>> 
`

jasonpaulos (approved):
Looks good, thanks for catching this

tzaffi:
> Hey this is my 5th minute looking at PyTeal and I can't really find any places that use the specific typing information that is/was provided by NamedTuple, but I did see some places that ask for the str() representation of a NamedTuple (for example, in tealop.py) and I verified that the str() for a dataclasses looks the same as a str() for NamedTuple). `>>> from typing import NamedTuple
> 
> > > > bob = NamedTuple ("BobType", [("value", str), ("mode", int)])
> > > > a = bob("Vlaue", 3)
> > > > from dataclasses import dataclass
> > > > @DataClass
> > > > ... class BobDC:
> > > > ...   value:str
> > > > ...   mode:int
> > > > ...
> > > > b = BobDC ("vvalue", "4")
> > > > print (str(b))
> > > > BobDC(value='vvalue', mode='4')
> > > > print (str(a))
> > > > BobType(value='Vlaue', mode=3)
> 
> `

Thanks for the quick investigation. You add to my confidence that this change is safe. The main usages of `OpType` are:
* switching logic (e.g. `if op == Op.add: ...`)
* TEAL generation via stringifying (e.g. `str(Op.add)` )
* and compatibility validation via the `mode` and `min_version` properties

All of these behaviors are remain the same under this proposed change, and all tests pass giving my high confidence.

PR #618, opened 2022-12-19
Upgrade to latest mypy

achidlow:
PEP-484 has changed to recommend against implicit optionals based on default arguments, and mypy has accordingly flipped the default in a recent release.

ref: https://peps.python.org/pep-0484/#union-types

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,9 +40,8 @@ black:
 flake8:
 	flake8 $(ALLPY)
 
-MYPY = pyteal scripts tests
 mypy:
-	mypy --show-error-codes $(MYPY)
+	mypy
 
 lint: black flake8 mypy
 
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,4 +1,6 @@
 [mypy]
+files = pyteal,scripts,tests
+python_version = 3.10
 
 [mypy-semantic_version.*]
 ignore_missing_imports = True
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ black==22.3.0
 flake8==5.0.4
 flake8-tidy-imports==4.6.0
 graviton@git+https://github.com/algorand/graviton@v0.5.0
-mypy==0.950
+mypy==0.991
 pytest==7.2.0
 pytest-cov==3.0.0
 pytest-timeout==2.1.0
--- a/pyteal/ast/abi/array_base.py
+++ b/pyteal/ast/abi/array_base.py
@@ -82,9 +82,9 @@ class Array(BaseType, Generic[T]):
         self,
         encoded: Expr,
         *,
-        start_index: Expr = None,
-        end_index: Expr = None,
-        length: Expr = None
+        start_index: Expr | None = None,
+        end_index: Expr | None = None,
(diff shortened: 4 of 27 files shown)


barnjamin:
@tzaffi I recall seeing a recent commit from you switching the `thing | None` type hint to `Optional[thing]`, was there a specific reason for that?

tzaffi:
> @tzaffi I recall seeing a recent commit from you switching the `thing | None` type hint to `Optional[thing]`, was there a specific reason for that?

These are [semantically equivalent](https://peps.python.org/pep-0604/). I think we ought to come to an agreement about what style we ought to adopt going forward -so we don't need to spend any mental calories on it. I propose:

* use `Optional[TheType]` instead of `TheType | None
* Use `TheFirstType | TheSecondType` instead of `Union[TheFirstType, TheSecondType]`

Why? I don't really have a good reason. I could be easily swayed to another viewpoint. But we should adopt _some_ style.

requirements.txt
```diff
@@ -2,7 +2,7 @@ black==22.3.0
 flake8==5.0.4
 flake8-tidy-imports==4.6.0
 graviton@git+https://github.com/algorand/graviton@v0.5.0
-mypy==0.950
+mypy==0.991
```
tzaffi, line 5: 👍

barnjamin:
@achidlow noted that the use of Union/Optional are required if the type must be quoted

tzaffi:
> @achidlow noted that the use of Union/Optional are required if the type must be quoted

That's an argument for not using `|` at all.

mypy.ini
```diff
@@ -1,4 +1,6 @@
 [mypy]
+files = pyteal,scripts,tests
```
tzaffi, line 2: Should we remove the `MYPY` variable in the `Makefile` as well, now that the config is taking over this decision?

tzaffi (commented):
Thanks for upgrading mypy!

I had a small nit regarding the `Makefile`.

Also, I won't hold up the PR on account of this, but I created #619 and suggested that we disallow `|` for annotations going forward.

mypy.ini
achidlow: Yes, that makes sense. Done ✅

achidlow:
@tzaffi pending the outcome of #619, I've at least made this PR internally consistent in the use of `|` over `Optional`. If you'd prefer it the other way around for now I'm happy to flip that for all the methods changed in this PR.

Makefile
```diff
@@ -40,9 +40,8 @@ black:
 flake8:
 	flake8 $(ALLPY)
 
-MYPY = pyteal scripts tests
 mypy:
-	mypy --show-error-codes $(MYPY)
+	mypy
```
tzaffi, line 44: Why drop the `--show-error-codes` arg? It's been handy on occasion.

achidlow: Not sure if this got flipped at some point, but that's [the default](https://mypy.readthedocs.io/en/stable/error_codes.html#displaying-error-codes) and assumed. You have to disable it with `--hide-error-codes`

tzaffi (approved):
Looks great, thank you!

PR #605, opened 2022-11-15
Test identical functionality across versions

tzaffi:
## Summary of changes
Use Graviton's new capability ([Graviton #36](https://github.com/algorand/graviton/pull/36)) for asserting that two separate teal programs behave exactly the same relative a set of invariants to test v8 versions (with frame pointers) vs. v6 (without).

## Graviton Dependency

- [x] [Graviton #36](https://github.com/algorand/graviton/pull/36) should be merged in first
- [x] [Graviton #42](https://github.com/algorand/graviton/pull/42) should be merged in first
- [x] And the graviton dependency point to `v0.7.0`
- [x] NO! ~Shall we merge in the following first?~
  - [x] ~#628 into this PR~
  - [x] ~[Graviton #44](https://github.com/algorand/graviton/pull/44)~

PR #621, opened 2022-12-22
Prevent blackbox_test_runner from passing without invariant assertions

michaeldiamant:
Adds an assertion to confirm `blackbox_test_runner` tests supply _at least_ 1 invariant.  The intent is to prevent a test from passing when it either does not declare invariants or structures invariant declaration incorrectly.

Notes:
* While reviewing #559, I discovered that _no_ invariant assertions happen in `test_blackbox_subroutines_as_apps` and `test_blackbox_subroutines_as_logic_sigs` because the tests configure invariants with key = `assertions` while `blackbox_test_runner` looks for key = `invariants`.
* Changing the test keys from `assertions` to `invariants` results in failed tests.  I open the PR to request help in fixing the broken assertions.  Once fixed, I'd like to add the defensive check shown here.
* Reviewing git history, I confirmed the key mismatch exists since the functionality was originally merged in https://github.com/algorand/pyteal/pull/249/.

### DIFF
--- a/tests/integration/graviton_test.py
+++ b/tests/integration/graviton_test.py
@@ -513,6 +513,7 @@ def blackbox_test_runner(
     inputs, predicates = Invariant.inputs_and_invariants(
         scenario, exec_mode, raw_predicates=True
     )
+    assert predicates, "Must configure >= 1 predicate"
 
     # 3. execute dry run sequence:
     execute = DryRunExecutor.execute_one_dryrun


tests/integration/graviton_test.py
```diff
@@ -513,6 +513,7 @@ def blackbox_test_runner(
     inputs, predicates = Invariant.inputs_and_invariants(
         scenario, exec_mode, raw_predicates=True
     )
+    assert predicates, "Must configure >= 1 predicate"
```
tzaffi, line 516: This proves a graviton bug. It should assert (by default) that predicates are actually provided.

tzaffi: I've incorporated this change in #605. Here's the commit that prove it: https://github.com/algorand/pyteal/pull/605/commits/2dc41daf7875c90a536c0bde577b9167b4d4f2fa

michaeldiamant: @tzaffi Thanks for bringing in the defensive check.  I'll close this PR.

michaeldiamant:
Closing - The defensive check will be brought in via https://github.com/algorand/pyteal/pull/605.

PR #623, opened 2022-12-23
Optimize `DupN` code generation & fix testing bug

jasonpaulos:
Optimize code generated by the `DupN` expression.

Prior to this PR, that expression always used the `dupn` op. However, if only 1 additional value is needed, the `dup` op can be used to save a byte. And if no additional values are needed and you still use `DupN`, no additional op is needed.

While testing this change, I noticed a bug in the `assert_teal_as_expected` function from `tests/compile_asserts.py`. This function used to only check that the expected line of code prefixes the actual line, and unfortunately `dup` prefixes `dupn 1`. I decided to upgrade the comparison function by using the `difflib` package ~~and processing each line to strip away all auxiliary information.~~

### DIFF
--- a/pyteal/ast/frame.py
+++ b/pyteal/ast/frame.py
@@ -41,14 +41,7 @@ class LocalTypeSegment(Expr):
                 )
 
     def __teal__(self, options: "CompileOptions") -> tuple[TealBlock, TealSimpleBlock]:
-        if self.count == 1:
-            inst_srt, inst_end = self.auto_instance.__teal__(options)
-            return inst_srt, inst_end
-        else:
-            dupn_srt, dupn_end = DupN(self.auto_instance, self.count - 1).__teal__(
-                options
-            )
-            return dupn_srt, dupn_end
+        return DupN(self.auto_instance, self.count - 1).__teal__(options)
 
     def __str__(self) -> str:
         return f"(LocalTypeSegment: (type: {self.local_type}) (count: {self.count}))"
@@ -322,6 +315,14 @@ class DupN(Expr):
     """
 
     def __init__(self, value: Expr, repetition: int):
+        """Create a DupN expression.
+
+        Args:
+            value: The value to be duplicated.
+            repetition: How many additional times the value should be added to the stack. At the end
+                of this operation, `repetition+1` elements will be added to the stack. Zero can be
+                specified here to indicate no duplication.
+        """
         super().__init__()
         require_type(value, TealType.anytype)
         if repetition < 0:
@@ -330,6 +331,15 @@ class DupN(Expr):
         self.repetition = repetition
 
     def __teal__(self, options: "CompileOptions") -> tuple[TealBlock, TealSimpleBlock]:
+        if self.repetition == 0:
+            # no duplication required
+            return self.value.__teal__(options)
+
+        if self.repetition == 1:
+            # use normal dup op for just 1 duplication
+            op = TealOp(self, Op.dup)
+            return TealBlock.FromOp(options, op, self.value)
+
         verifyProgramVersion(
             Op.dupn.min_version,
             options.version,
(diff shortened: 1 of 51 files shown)


jasonpaulos:
I believe the fact that 538f784 passes tests is a bug in our test code. I'm investigating further

pyteal/ast/frame.py
```diff
@@ -44,14 +44,7 @@ def __init__(self, local_type: TealType, count: int):
...
     def __teal__(self, options: "CompileOptions") -> tuple[TealBlock, TealSimpleBlock]:
-        if self.count == 1:
-            inst_srt, inst_end = self.auto_instance.__teal__(options)
-            return inst_srt, inst_end
-        else:
-            dupn_srt, dupn_end = DupN(self.auto_instance, self.count - 1).__teal__(
-                options
-            )
-            return dupn_srt, dupn_end
+        return DupN(self.auto_instance, self.count - 1).__teal__(options)
```
tzaffi, line 47: 👍

tests/compile_asserts.py
```diff
@@ -1,4 +1,5 @@
 from pathlib import Path
+from difflib import unified_diff
```
tzaffi, line 2: 👍

tests/compile_asserts.py
```diff
@@ -24,43 +25,57 @@ def compile_and_save(approval, version, test_name):
...
+    quote_types = ("'", '"')
+    comment_start = "/"
 
+    processed_end = len(line) - 1
 
-def assert_teal_as_expected(path2actual, path2expected):
+    current_quote: str | None = None
+    for i, char in enumerate(line):
+        for qt in quote_types:
+            if char == qt:
```
jasonpaulos, line 37: Actually this probably needs to take into account escaped quotes, e.g. `\"`

tzaffi: I'm not sure the purpose of processing each line, versus just keeping the raw original. IMO, for this test code it's okay (even preferrable?) to fail because the generated teal is non-identical, even if it's semantically equivalent. I think we would want to know when this happens, and in that case we could decide to simply replace the old fixtures with the new ones.

tzaffi (commented):
I agree with the approach to replace `dupn` with `dup` or nothing whenever possible. I also support bringing in the more standard `united_diff()` into `compile_asserts.py` vs. the previous ad-hoc diffing function. 

I left a question regarding processing of the lines from Teal file fixtures, vs. just reading them raw.

tests/compile_asserts.py
tzaffi: Ok, I remember now. When I originally wrote this test, I had some hand-commented code which I didn't want to fail the diff. I think at this point, the usage has grown much larger than I originally anticipated. So the comments I just wrote, convince me that it's better to _not_ try and pre-process. I.e., don't ignore comments. If we agree that this approach is reasonable, I'm more than happy to replace the original fixtures with fixtures which will pass the tests.

jasonpaulos: I agree that it's preferable to fail if the generated code is different (otherwise we shouldn't be checking in any code for this).

Some checked in code is augmented with comments, so a direct line by line comparison will fail, even though the ops are the same. This processing function is responsible for stripping away the comments plus any difference in whitespace.

I haven't looked in detail at all the code, but my impression is we'd want to keep the additional user-generated comments if at all possible.

jasonpaulos: Ah, looks like we replied at the same time. If you're ok with getting rid of the comments, I'm happy to make that change to simplify this comparison code

tzaffi: Yeah for sure. Those original old fixtures don't need their comments anymore.

jasonpaulos: Updated in 27f530e

tzaffi (approved):
Looks great!

PR #624, opened 2022-12-23
add method to encode tuple element

barnjamin:
I just want bytes, no need for other scratch vars or lambdas

diff from a program where i switched to this:

image

### DIFF
--- a/pyteal/ast/abi/array_base_test.py
+++ b/pyteal/ast/abi/array_base_test.py
@@ -161,3 +161,105 @@ def test_ArrayElement_store_into():
 
         with pytest.raises(pt.TealInputError):
             element.store_into(abi.Tuple(abi.TupleTypeSpec(elementType)))
+
+
+def test_ArrayElement_encoding():
+    from pyteal.ast.abi.util import substring_for_decoding
+
+    for elementType in STATIC_TYPES + DYNAMIC_TYPES:
+        staticArrayType = abi.StaticArrayTypeSpec(elementType, 100)
+        staticArray = staticArrayType.new_instance()
+        index = pt.Int(9)
+
+        element = abi.ArrayElement(staticArray, index)
+        expr = element.encode()
+
+        encoded = staticArray.encode()
+        stride = pt.Int(staticArray.type_spec()._stride())
+        expectedLength = staticArray.length()
+        if elementType == abi.BoolTypeSpec():
+            expectedExpr = pt.SetBit(
+                pt.Bytes(b"\x00"),
+                pt.Int(0),
+                pt.GetBit(encoded, index),
+            )
+        elif not elementType.is_dynamic():
+            expectedExpr = substring_for_decoding(
+                encoded, start_index=stride * index, length=stride
+            )
+        else:
+            expectedExpr = substring_for_decoding(
+                encoded,
+                start_index=pt.ExtractUint16(encoded, stride * index),
+                end_index=pt.If(index + pt.Int(1) == expectedLength)
+                .Then(pt.Len(encoded))
+                .Else(pt.ExtractUint16(encoded, stride * index + pt.Int(2))),
+            )
+
+        expected, _ = expectedExpr.__teal__(options)
+        expected.addIncoming()
+        expected = pt.TealBlock.NormalizeBlocks(expected)
+
+        actual, _ = expr.__teal__(options)
+        actual.addIncoming()
+        actual = pt.TealBlock.NormalizeBlocks(actual)
+
+        with pt.TealComponent.Context.ignoreExprEquality():
+            assert actual == expected
+
+        with pytest.raises(pt.TealInputError):
+            element.store_into(abi.Tuple(abi.TupleTypeSpec(elementType)))
+
+    for elementType in STATIC_TYPES + DYNAMIC_TYPES:
+        dynamicArrayType = abi.DynamicArrayTypeSpec(elementType)
+        dynamicArray = dynamicArrayType.new_instance()
+        index = pt.Int(9)
+
+        element = abi.ArrayElement(dynamicArray, index)
+        expr = element.encode()
+
+        encoded = dynamicArray.encode()
+        stride = pt.Int(dynamicArray.type_spec()._stride())
+        expectedLength = dynamicArray.length()
+        if elementType == abi.BoolTypeSpec():
+            expectedExpr = pt.SetBit(
+                pt.Bytes(b"\x00"),
+                pt.Int(0),
+                pt.GetBit(encoded, index + pt.Int(16)),
+            )
+        elif not elementType.is_dynamic():
+            expectedExpr = substring_for_decoding(
+                encoded, start_index=stride * index + pt.Int(2), length=stride
+            )
+        else:
+            expectedExpr = substring_for_decoding(
+                encoded,
+                start_index=pt.ExtractUint16(encoded, stride * index + pt.Int(2))
+                + pt.Int(2),
+                end_index=pt.If(index + pt.Int(1) == expectedLength)
+                .Then(pt.Len(encoded))
+                .Else(
+                    pt.ExtractUint16(encoded, stride * index + pt.Int(2) + pt.Int(2))
+                    + pt.Int(2)
+                ),
+            )
+
+        expected, _ = expectedExpr.__teal__(options)
+        expected.addIncoming()
+        expected = pt.TealBlock.NormalizeBlocks(expected)
+
+        actual, _ = expr.__teal__(options)
+        actual.addIncoming()
+        actual = pt.TealBlock.NormalizeBlocks(actual)
+
+        with pt.TealComponent.Context.ignoreExprEquality():
+            with pt.TealComponent.Context.ignoreScratchSlotEquality():
+                assert actual == expected
+
+        assert pt.TealBlock.MatchScratchSlotReferences(
+            pt.TealBlock.GetReferencedScratchSlots(actual),
+            pt.TealBlock.GetReferencedScratchSlots(expected),
+        )
+
+        with pytest.raises(pt.TealInputError):
+            element.store_into(abi.Tuple(abi.TupleTypeSpec(elementType)))
--- a/pyteal/ast/abi/tuple.py
+++ b/pyteal/ast/abi/tuple.py
@@ -1,4 +1,5 @@
 from inspect import get_annotations
+from dataclasses import dataclass
 from typing import (
     List,
     Sequence,
@@ -14,15 +15,14 @@ from typing import (
 )
 from collections import OrderedDict
 
-from pyteal.types import TealType
+from pyteal.types import TealType, require_type
 from pyteal.errors import TealInputError, TealInternalError
 from pyteal.ast.expr import Expr
 from pyteal.ast.seq import Seq
 from pyteal.ast.int import Int
 from pyteal.ast.bytes import Bytes
 from pyteal.ast.unaryexpr import Len
-from pyteal.ast.binaryexpr import ExtractUint16, GetBit
-from pyteal.ast.ternaryexpr import SetBit
+from pyteal.ast.binaryexpr import ExtractUint16
 from pyteal.ast.naryexpr import Concat
 from pyteal.ast.abstractvar import alloc_abstract_var
 
@@ -37,7 +37,11 @@ from pyteal.ast.abi.bool import (
     _bool_aware_static_byte_length,
 )
 from pyteal.ast.abi.uint import NUM_BITS_IN_BYTE, Uint16
-from pyteal.ast.abi.util import substring_for_decoding, type_spec_from_annotation
+from pyteal.ast.abi.util import (
+    substring_for_decoding,
+    type_spec_from_annotation,
+    _GetAgainstEncoding,
+)
 
 
 def _encode_tuple(values: Sequence[BaseType]) -> Expr:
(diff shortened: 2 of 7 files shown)


pyteal/ast/abi/tuple.py
```diff
@@ -117,6 +117,99 @@ def _encode_tuple(values: Sequence[BaseType]) -> Expr:
...
+
+    if type(valueType) is Bool:
+        if ignoreNext > 0:
+            # value is in the middle of a bool sequence
+            bitOffsetInBoolSeq = lastBoolLength - ignoreNext
+            bitOffsetInEncoded = lastBoolStart * NUM_BITS_IN_BYTE + bitOffsetInBoolSeq
+        else:
+            # value is the beginning of a bool sequence (or a single bool)
+            bitOffsetInEncoded = offset * NUM_BITS_IN_BYTE
+        return GetBit(encoded, Int(bitOffsetInEncoded))
```
barnjamin, line 158: bug here but also where(?) were checking return type

Would it help to have Expr[bytes|uint64] at least for mypy typechecking?

tzaffi: I asked Chat GPT to rewrite the function. It came up with the following, which still passes our unit tests. Does this one also have the bug?

```python

def _index_tuple_bytes(value_types, encoded, index):
    if not (0 <= index < len(value_types)):
        raise ValueError("Index outside of range")

    offset = 0
    for i, value_type in enumerate(value_types[:index]):
        if value_type == BoolTypeSpec():
            bool_length = _consecutive_bool_type_spec_num(value_types, i)
            offset += _bool_sequence_length(bool_length)
        elif value_type.is_dynamic():
            offset += 2
        else:
            offset += value_type.byte_length_static()

    value_type = value_types[index]

    if value_type == BoolTypeSpec():
        bit_offset = offset * NUM_BITS_IN_BYTE
        return GetBit(encoded, Int(bit_offset))
    elif value_type.is_dynamic():
        has_next_dynamic_value = False
        next_dynamic_value_offset = offset + 2
        for i, type_after in enumerate(value_types[index + 1 :], start=index + 1):
            if type_after.is_dynamic():
                has_next_dynamic_value = True
                break
            if type_after == BoolTypeSpec():
                bool_length = _consecutive_bool_type_spec_num(value_types, i)
                next_dynamic_value_offset += _bool_sequence_length(bool_length)
            else:
                next_dynamic_value_offset += type_after.byte_length_static()

        start_index = ExtractUint16(encoded, Int(offset))
        if not has_next_dynamic_value:
            return substring_for_decoding(encoded, start_index=start_index)
        end_index = ExtractUint16(encoded, Int(next_dynamic_value_offset))
        return substring_for_decoding(
            encoded, start_index=start_index, end_index=end_index
        )
    else:
        start_index = Int(offset)
        length = Int(value_type.byte_length_static())
        if index + 1 == len(value_types):
            if offset == 0:
                return encoded
            return substring_for_decoding(encoded, start_index=start_index)
        if offset == 0:
            return substring_for_decoding(encoded, length=length)
        return substring_for_decoding(encoded, start_index=start_index, length=length)
```

barnjamin: only a bug in that im not checking the return type at least in the `encode` function is getting `Bytes`

It does return ```        return GetBit(encoded, Int(bit_offset))
``` but im not too mad since it couldnt know GetBit would put a uint64 on the stack

Comment (tzaffi) [reply pyteal/ast/abi/tuple.py:158]: Right. I see now what you're saying. Yeah, maybe generics could be helpful here. Given the fact that such a type check would get ignored at runtime, I still see us needing our `type_of()` assertions.

File: pyteal/ast/abi/tuple.py [hunk 2]
```diff
@@ -405,6 +499,13 @@ def store_into(self, output: T) -> Expr:
             output,
         )
 
+    def encode(self) -> Expr:
+        return _index_tuple_bytes(
+            self.tuple.type_spec().value_type_specs(),
+            self.tuple.encode(),
+            self.index,
+        )
```
ahangsu, line 507: nit: I don't think this should be a special case for only tuple, i.e., this should be a public method in virtual class `ComputedType`, and extend to static/dynamic-array-element. It is non-uniform to let only tuple has such element-wise encoding retrieval.

pyteal/ast/abi/tuple.py
```diff
@@ -117,6 +118,99 @@ def _encode_tuple(values: Sequence[BaseType]) -> Expr:
     return Concat(*toConcat)
 
 
+def _index_tuple_bytes(
+    value_types: Sequence[TypeSpec], encoded: Expr, index: int
+) -> Expr:
+    if not (0 <= index < len(value_types)):
```
ahangsu, line 124:
```suggestion
    if index not in range(len(value_types)):
```

pyteal/ast/abi/tuple.py
```diff
@@ -117,6 +118,99 @@ def _encode_tuple(values: Sequence[BaseType]) -> Expr:
...
+
+    if type(valueType) is Bool:
+        if ignoreNext > 0:
+            # value is in the middle of a bool sequence
+            bitOffsetInBoolSeq = lastBoolLength - ignoreNext
+            bitOffsetInEncoded = lastBoolStart * NUM_BITS_IN_BYTE + bitOffsetInBoolSeq
+        else:
+            # value is the beginning of a bool sequence (or a single bool)
+            bitOffsetInEncoded = offset * NUM_BITS_IN_BYTE
+        return SetBit(Bytes(b"\x00"), Int(0), GetBit(encoded, Int(bitOffsetInEncoded)))
```
ahangsu, line 159: 👍

ahangsu (commented):
overall looks good and correct in implementation, need changelog entry and resolve discussion thread on `encode` method.

pyteal/ast/abi/tuple.py
```diff
@@ -117,6 +118,99 @@ def _encode_tuple(values: Sequence[BaseType]) -> Expr:
     return Concat(*toConcat)
 
 
+def _index_tuple_bytes(
```
jasonpaulos, line 121: If at all possible it would be preferable not to maintain tuple indexing code in two separate places. Can the majority of the complexity duplicated in `_index_tuple_bytes` and `_index_tuple` be shared somewhere? Maybe there should be a helper function which does the heavy lifting for these?

pyteal/ast/abi/tuple.py
jasonpaulos: +1 to this. A less-efficient but still correct implementation that could be added to `ComputedValue` could be something like this:

```python
def encode(self) -> Expr:
    return self.use(lambda value: value.encode())
```

And `TupleElement` (and maybe `ArrayElement` eventually) can override this with a more efficient approach, like the one in this PR.

jasonpaulos (commented):
Generally makes sense, but I'd like to see sufficient test coverage too. (Check out `test_indexTuple` in `pyteal/ast/abi/tuple_test.py` for coverage of `_index_tuple` for inspiration)

pyteal/ast/abi/tuple.py
ahangsu: I made an attempt to attack in this direction, and a PR up here: https://github.com/algorand/pyteal/pull/629/, please consider a quick pass

pyteal/ast/abi/tuple.py
```diff
@@ -117,6 +118,99 @@ def _encode_tuple(values: Sequence[BaseType]) -> Expr:
...
+
+        if typeBefore.is_dynamic():
+            offset += 2
+            continue
+
+        offset += typeBefore.byte_length_static()
+
+    valueType = value_types[index]
+
+    if type(valueType) is Bool:
```
ahangsu, line 151: this is wrong, should be following
```suggestion
    if type(valueType) is BoolTypeSpec:
```

ahangsu: should be fixed by https://github.com/algorand/pyteal/pull/629

pyteal/ast/abi/tuple.py
ahangsu: I also attempted to unify the largely same logic into a same function by https://github.com/algorand/pyteal/pull/629, please take a look?

pyteal/ast/abi/tuple.py
ahangsu: resolving for #629 is now merged in

ahangsu:
I am still opinionated on this one tho, after some more thoughts. Maybe I am wrong.

This `encode` method applies to all `ComputedType`, so it also applies to `ReturnedType` which wraps up a computation returning something on stack. What if the `encode` called twice for a `ReturnedType` instance with a computation that has side effect? That would not be ideal.

pyteal/ast/abi/array_base_test.py
```diff
@@ -161,3 +161,105 @@ def test_ArrayElement_store_into():
...
+            )
+        elif not elementType.is_dynamic():
+            expectedExpr = substring_for_decoding(
+                encoded, start_index=stride * index, length=stride
+            )
+        else:
+            expectedExpr = substring_for_decoding(
+                encoded,
+                start_index=pt.ExtractUint16(encoded, stride * index),
+                end_index=pt.If(index + pt.Int(1) == expectedLength)
```
barnjamin, line 194: Will this ever be true? wouldn't this always eval to `9 + 1 == 100`?

pyteal/ast/abi/util.py
```diff
@@ -593,3 +598,50 @@ def type_spec_is_assignable_to(a: TypeSpec, b: TypeSpec) -> bool:
...
+    start_index: Expr | None = field(kw_only=True, default=None)
+    end_index: Expr | None = field(kw_only=True, default=None)
+    length: Expr | None = field(kw_only=True, default=None)
+
+    def __post_init__(self):
+        from pyteal.ast.abi import BoolTypeSpec
+
+        require_type(self.full_encoding, TealType.bytes)
+        if self.type_spec == BoolTypeSpec():
+            require_type(self.start_index, TealType.uint64)
```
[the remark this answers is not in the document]
ahangsu: I probably should state it clear here, so I want to get bit number here for boolean in encoding, and start index should be not none, with teal type uint64.

barnjamin:
@ahangsu correctly notes that if someone were to call something like `encode` multiple times on a `ReturnedValue` the `computation` Expr would be run again.

However, this behavior is already present in the current release. While it isn't ideal I think it shouldn't be a show stopped for this pr.  Perhaps another pr that improves how a ReturnedValue caches its computation (via scratch var or otherwise?) is warranted.

pyteal/ast/abi/util.py
ahangsu, line 616:
```suggestion
        # The check is to see the existence of `start_index` for bool case
        # Bool case need only start_index, and must have type uint64
        if self.type_spec == BoolTypeSpec():
            require_type(self.start_index, TealType.uint64)
```

pyteal/ast/abi/tuple.py
```diff
@@ -117,97 +122,128 @@ def _encode_tuple(values: Sequence[BaseType]) -> Expr:
...
+        require_type(self.encoded, TealType.bytes)
 
-        if typeBefore == BoolTypeSpec():
-            lastBoolStart = offset
-            lastBoolLength = _consecutive_bool_type_spec_num(value_types, i)
-            offset += _bool_sequence_length(lastBoolLength)
-            ignoreNext = lastBoolLength - 1
-            continue
+    def get_or_store(self, index: int, output: BaseType | None = None) -> Expr:
+        if index not in range(len(self.value_types)):
```
bbroder-algo, line 134: this seems like a slow way to check an index boundary

pyteal/ast/abi/tuple.py
```diff
@@ -117,97 +122,128 @@ def _encode_tuple(values: Sequence[BaseType]) -> Expr:
...
+        valueType = self.value_types[index]
+        if output is not None and output.type_spec() != valueType:
+            raise TypeError("Output type does not match value type")
 
-    start_index = Int(offset)
-    length = Int(valueType.byte_length_static())
+        if type(valueType) is BoolTypeSpec:
+            if ignoreNext > 0:
+                # value is in the middle of a bool sequence
+                bitOffsetInBoolSeq = lastBoolLength - ignoreNext
```
bbroder-algo, line 166: this made my head hurt but it looks right

pyteal/ast/abi/tuple.py
ahangsu: pushed a faster way to check index boundary: [fast index boundary check](https://github.com/algorand/pyteal/pull/624/commits/9b56038e24ae76f51d427c3e0297c53273e5eb09)

pyteal/ast/abi/util_test.py
```diff
@@ -1057,3 +1059,146 @@ def exists_in_unsafe_bidirectional(_ts: type):
...
+
+    get_against_encoding = _GetAgainstEncoding(
+        encoded,
+        testcase.type_spec,
+        start_index=testcase.start_index,
+        end_index=testcase.end_index,
+        length=testcase.length,
+    )
+    actual_store_expr = get_against_encoding.get_or_store(output_expr)
+    actual_encode_expr = get_against_encoding.get_or_store()
```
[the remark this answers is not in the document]
ahangsu: I think this is the only place `get_or_store` syntactically no args, but for all other places, `output` in the argument is of type `BaseType | None`.

pyteal/ast/abi/tuple.py
```diff
@@ -117,97 +122,128 @@ def _encode_tuple(values: Sequence[BaseType]) -> Expr:
...
-        nextDynamicValueOffset = offset + 2
+        offset = 0
         ignoreNext = 0
-        for i, typeAfter in enumerate(value_types[index + 1 :], start=index + 1):
+        lastBoolStart = 0
+        lastBoolLength = 0
+        for i, typeBefore in enumerate(self.value_types[:index]):
             if ignoreNext > 0:
                 ignoreNext -= 1
                 continue
```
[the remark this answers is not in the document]
bbroder-algo: yeah but it could be refactored to be a jump and not an iterated increment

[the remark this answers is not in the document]
ahangsu: commit the refactoring in commit [refactoring skip looping with next](https://github.com/algorand/pyteal/pull/624/commits/8237f389ab959860544281461017cbfd53bbe265)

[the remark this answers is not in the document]
ahangsu: I would rather not do this right now, for it makes code even harder to read, and more error prone.

Would be good to jump index by `ignoreNext`.

bbroder-algo (approved):
lgtm

PR #626, opened 2022-12-29
PR for Py-SDK version bump

ahangsu:
This should be blocked by https://github.com/algorand/py-algorand-sdk/issues/384, for pyteal is using py-sdk as a runtime dependency.

Merge in after https://github.com/algorand/py-algorand-sdk/issues/384 is resolved.

This PR upgrading runtime dependency should favor some type alert in `abi.Contract` in pyteal.


jasonpaulos (approved):
Looks good to me, thanks for making this

PR #629, opened 2023-01-03
Make encode method generalized for `ComputedType`

ahangsu:


pyteal/ast/abi/tuple.py
```diff
@@ -118,190 +122,130 @@ def _encode_tuple(values: Sequence[BaseType]) -> Expr:
...
-            bitOffsetInEncoded = offset * NUM_BITS_IN_BYTE
-        return SetBit(Bytes(b"\x00"), Int(0), GetBit(encoded, Int(bitOffsetInEncoded)))
-
-    if valueType.is_dynamic():
-        hasNextDynamicValue = False
-        nextDynamicValueOffset = offset + 2
+        offset = 0
         ignoreNext = 0
-        for i, typeAfter in enumerate(value_types[index + 1 :], start=index + 1):
+        lastBoolStart = 0
```
tzaffi, line 139: It's hard to tell -without a deep dive- whether this new version is functionally equivalent. _If_ I understand correctly (???), there is currently no unit test that covers this code so we can't be 100% confident of correctness. If a unit test is added, (but probably that should be in the parent PR), then this would allay my concerns.

As an aside, I'm already thinking of starting some nightly tests. It might be a good opportunity to add python's coverage tool to such tests: https://github.com/nedbat/coveragepy

tzaffi (commented):
A couple of minor suggestions.

pyteal/ast/abi/util.py
```diff
@@ -593,3 +594,51 @@ def type_spec_is_assignable_to(a: TypeSpec, b: TypeSpec) -> bool:
         return True
 
     return False
+
+
+def _get_encoding_or_store_from_encoded_bytes(
+    encoding_type: TypeSpec,
+    full_encoding: Expr,
```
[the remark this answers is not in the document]
ahangsu: got it, I will make the change as suggested, seems reasonable to me.

pyteal/ast/abi/tuple.py
```diff
@@ -118,190 +119,138 @@ def _encode_tuple(values: Sequence[BaseType]) -> Expr:
...
-            lastBoolStart = offset
-            lastBoolLength = _consecutive_bool_type_spec_num(value_types, i)
-            offset += _bool_sequence_length(lastBoolLength)
-            ignoreNext = lastBoolLength - 1
-            continue
-
-        if typeBefore.is_dynamic():
-            offset += 2
-            continue
+    def __call__(self, index: int, output: BaseType | None = None) -> Expr:
```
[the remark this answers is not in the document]
ahangsu: yes, that makes sense, somehow I regret making too succinct and fancy, yet hard to maintain.

I suggest renaming it to `get_or_store`, lmk if a better naming is proposed.

pyteal/ast/abi/tuple.py
ahangsu: My apologies, I am hacking through these APIs without a same set of unit test coverage, so I guarantee there will be a same set of unit test for functional equivalence.

The side note for coveragepy also makes sense to me, maybe not going into this PR, but as a separate one. The expected result should be somehow like go-algorand one, i.e., coverage +/- lines?

tzaffi: Regarding coverage, definitely not this PR. I've created issue #630 as a placeholder to discuss this.


I see that `__prototype_encoding_store_into()` is being tested for the case when the `output` parameter is present  from pre-existing tests that applied to `store_into()`. I'm not seeing where the case of _empty_ `output` parameter is being tested. Since this is very relevant to (or even exactly) the functionality being refactored, we can't have confidence -from testing at least- that the refactoring is correct. So I think before merging this child PR we should add more coverage to #624.

@barnjamin

ahangsu: The refactoring of code is straightforward, which is against the old version of `_index_tuple`, by changing return item according to the existence of `output: BaseType | None`.

I added the testcases for tuple elements and array elements anyways, and will add some testcases for `_GetAgainstEncoding`, but I don't anticipate I have bandwidth for coverage adding in #624, and I hope @barnjamin can help on that one.

tzaffi (approved):
I see now that the case of `output=None` for `__prototype_encoding_store_into()` is exercised, thank you!

Issue #611, opened 2022-12-06
CI: Check for sdist metadata warnings


michaeldiamant:
## Problem

Like https://github.com/algorand/py-algorand-sdk/issues/409, but for pyteal.

PR #631, opened 2023-01-11
Add build check for python setup.py sdist

michaeldiamant:
Addresses https://github.com/algorand/pyteal/issues/611 by adding a build check to confirm `python setup.py` passes without warnings or errors.

The PR ports the solution applied in py-algorand-sdk (https://github.com/algorand/py-algorand-sdk/pull/427/).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,11 @@ flake8:
 mypy:
 	mypy
 
-lint: black flake8 mypy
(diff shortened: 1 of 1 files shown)


algochoi (approved):
Looks good - works locally on my machine, and looks like it is run in github workflow as well :shipit:

PR #636, opened 2023-01-18
Refactoring Router taking CSP expression

### LINKED ISSUE #638: Improve `Router` clear state program generation
## Problem

Currently the ABI `Router` class supports methods and bare app calls for the clear state action. Any such bare app call and/or methods get put into the clear state program, following the same routing rules as the approval program.

This is a problem, since clear state programs operate under very different failure rules. A clear state program which wishes to have any effect on the state of the chain must never fail, since doing so reverts any effects made in that program. With the current routing rules, it's possible to fail a program by attempting to call an unsupported ABI method, or by providing any arguments when attempting to issue a bare app call.

## Solution

Based on internal discussion, we've agreed on the following solution:
* Drop support for clear state bare app calls and for ABI methods on clear state.
  * These argument keys should still exist, but an error should be thrown at runtime if someone attempts to use them. The error message should indicate the alternative approach described below.
* Introduce a new top-level `Router` argument, `clear_state`, that takes an `Expr`. If provided, this expression would become the entire clear state program; no additional code would be injected by PyTeal.

ahangsu:
An attempt made, so we can get around the erroring in csp.

Closes #638

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,24 @@
 # Unreleased
 
-# Added
+## Added
+
+## Changed
+
+## Fixed
+
+# v0.21.0
+
+## Added
 * Added frame pointer support for subroutine arguments, replacing the previous usage of scratch. ([#562](https://github.com/algorand/pyteal/pull/562))
 * Added frame pointer support for local ABI variables in subroutine. ([#606](https://github.com/algorand/pyteal/pull/606))
-* Added `frame_pointers` property in `OptimizeOptions` to optimize away scratch slots during subroutine calls. This defaults to frame pointer usage when not specified. ([#613](https://github.com/algorand/pyteal/pull/613))
+* Added `frame_pointers` property in `OptimizeOptions` to optimize away scratch slots during subroutine calls. This defaults to frame pointer usage when not specified for AVM version 8+. ([#613](https://github.com/algorand/pyteal/pull/613))
 
-# Fixed
+## Fixed
 * Allowing the `MethodCall` and `ExecuteMethodCall` to be passed `None` as app_id argument in the case of an app create transaction ([#592](https://github.com/algorand/pyteal/pull/592))
+* No longer accidentally include additional package folders in release ([#610](https://github.com/algorand/pyteal/pull/610), [#614](https://github.com/algorand/pyteal/pull/614))
+* Fixed mistakes in docs ([#612](https://github.com/algorand/pyteal/pull/612), [#625](https://github.com/algorand/pyteal/pull/625), [#627](https://github.com/algorand/pyteal/pull/627))
 
-# Changed
+## Changed
 * Introducing `AbstractVar` to abstract value access: store, load, and stack type. ([#584](https://github.com/algorand/pyteal/pull/584))
   * NOTE: a backwards incompatable change was imposed in this PR: previous ABI value's public member `stored_value` with type `ScratchVar`, is now changed to protected member `_stored_value` with type `AbstractVar`.
 * Starting with program version 9, when `scratch_slots` flag isn't provided to `OptimizeOptions`, default to optimizing. For versions 8 and earlier the default is and remains to _not_ optimize. ([#613](https://github.com/algorand/pyteal/pull/613))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ## Changed
 * Improved Router clear state program generation. ([#636](https://github.com/algorand/pyteal/pull/636))
+  * NOTE: a backwards incompatable change was imposed in this PR: previous Clear State Program (CSP) can be constructed in router by registering ABI methods or bare appcalls, now one has to use `clear_state` argument in `Router.__init__` to construct the CSP.
 
 ## Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Added
 
 ## Changed
+* Improved Router clear state program generation. ([#636](https://github.com/algorand/pyteal/pull/636))
 
 ## Fixed
 
@@ -25,7 +26,6 @@
 * Replaced the usage of `typing.NamedTuple` with `dataclass` for `class OpType` in the **ir** package in order to avoid [a regression coming in Python 3.11.1](https://github.com/python/cpython/issues/100098). ([#615](https://github.com/algorand/pyteal/pull/615))
 * Upgrade mypy to v0.991. ([#618](https://github.com/algorand/pyteal/pull/618))
 * Upgrade py-algorand-sdk to v2.0.0. ([#626](https://github.com/algorand/pyteal/pull/626))
-* Improved Router clear state program generation. ([#636](https://github.com/algorand/pyteal/pull/636))
 
 # 0.20.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 * Replaced the usage of `typing.NamedTuple` with `dataclass` for `class OpType` in the **ir** package in order to avoid [a regression coming in Python 3.11.1](https://github.com/python/cpython/issues/100098). ([#615](https://github.com/algorand/pyteal/pull/615))
 * Upgrade mypy to v0.991. ([#618](https://github.com/algorand/pyteal/pull/618))
 * Upgrade py-algorand-sdk to v2.0.0. ([#626](https://github.com/algorand/pyteal/pull/626))
+* Improved Router clear state program generation. ([#636](https://github.com/algorand/pyteal/pull/636))
 
 # 0.20.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ## Changed
 * Improved Router clear state program generation. ([#636](https://github.com/algorand/pyteal/pull/636))
-  * NOTE: a backwards incompatable change was imposed in this PR: previous Clear State Program (CSP) can be constructed in router by registering ABI methods or bare appcalls, now one has to use `clear_state` argument in `Router.__init__` to construct the CSP.
+  * NOTE: a backwards incompatable change was imposed in this PR: previous Clear State Program (CSP) can be constructed in router by registering ABI methods or bare app calls, now one has to use `clear_state` argument in `Router.__init__` to construct the CSP.
 
 ## Fixed
 
--- a/pyteal/ast/router.py
+++ b/pyteal/ast/router.py
@@ -504,7 +504,9 @@ class Router:
             name: the name of the smart contract, used in the JSON object.
             bare_calls: the bare app call registered for each on_completion.
             descr: a description of the smart contract, used in the JSON object.
-            clear_state: an expression describing the behavior of clear state program.
+            clear_state: an expression describing the behavior of clear state program. This
+                expression will be the entirety of the clear state program; no additional code is
+                inserted by the Router. If not provided, the clear state program will always reject.
         """
 
         self.name: str = name
--- a/pyteal/ast/router.py
+++ b/pyteal/ast/router.py
@@ -126,15 +126,16 @@ class MethodConfig:
 MethodConfig.__module__ = "pyteal"
 
 
+ActionType = Expr | SubroutineFnWrapper | ABIReturnSubroutine
+
+
 @dataclass(frozen=True)
 class OnCompleteAction:
     """
     OnComplete Action, registers bare calls to one single OnCompletion case.
     """
 
-    action: Optional[Expr | SubroutineFnWrapper | ABIReturnSubroutine] = field(
-        kw_only=True, default=None
-    )
+    action: Optional[ActionType] = field(kw_only=True, default=None)
     call_config: CallConfig = field(kw_only=True, default=CallConfig.NEVER)
 
     def __post_init__(self):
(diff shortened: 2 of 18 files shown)


pyteal/ast/router.py
```diff
@@ -96,6 +83,13 @@ class MethodConfig:
     update_application: CallConfig = field(kw_only=True, default=CallConfig.NEVER)
     delete_application: CallConfig = field(kw_only=True, default=CallConfig.NEVER)
 
+    def __post_init__(self):
+        if self.clear_state != CallConfig.NEVER:
+            raise TealInputError(
+                "Attempt to construct clear state program from MethodConfig, "
+                "use Router top level argument `clear_state` instead"
+            )
+
```
ahangsu, line 92: Question: this function will error if one sets `clear_state` to anything other than default. Are we good here, or any more steps needed to totally remove the keyword arg `clear_state`?

jasonpaulos: This seems sufficient to me

pyteal/ast/router.py
```diff
@@ -508,19 +489,26 @@ def __init__(
...
 
         self.name: str = name
         self.descr = descr
 
         self.approval_ast = ASTBuilder()
-        self.clear_state_ast = ASTBuilder()
+        self.clear_state: Expr = (
+            Reject()
+            if clear_state is None
+            else ASTBuilder.wrap_handler(False, clear_state)
```
jasonpaulos, line 510: nit: right now `wrap_handler` has error messages that talk about bare app calls. That will no longer be accurate for the clear state program, since it's not a normal bare app call.

pyteal/ast/router.py
```diff
@@ -63,19 +63,6 @@ def approval_condition_under_config(self) -> Expr | int:
             case _:
                 raise TealInternalError(f"unexpected CallConfig {self}")
 
-    def clear_state_condition_under_config(self) -> int:
```
jasonpaulos, line 66: I'd like to see tests that confirm an error is raised if you attempt to provide a clear state bare app call or method.

jasonpaulos:
Note that we will also need to update the docs, notably https://pyteal.readthedocs.io/en/stable/abi.html#registering-bare-app-calls

pyteal/ast/router.py
tzaffi: We might want to have a link to a URL that explains in more detail. But that can also be added in the followup pr/release.

pyteal/ast/router.py
```diff
@@ -128,8 +122,11 @@ def approval_cond(self) -> Expr | int:
                         )
             return Or(*cond_list)
 
-    def clear_state_cond(self) -> Expr | int:
-        return self.clear_state.clear_state_condition_under_config()
+
+MethodConfig.__module__ = "pyteal"
+
+
+ActionType = Expr | SubroutineFnWrapper | ABIReturnSubroutine
```
tzaffi, line 129: 👍

pyteal/compiler/compiler_test.py
```diff
@@ -2326,25 +2324,8 @@ def log_creation(*, output: pt.abi.String) -> pt.Expr:
         )
         assert meth.method_signature() == "log_creation()string"
 
-        @pt.ABIReturnSubroutine
-        def approve_if_odd(condition_encoding: pt.abi.Uint32) -> pt.Expr:
```
tzaffi, line 2330: This deleted router method is a great example to think through. 
In the short term, I support the current strict approach. But in the medium term, it's worth seeing if we can relax the restrictions a bit.
As we discussed, there could still be a use case when someone actually wants to _fail_ during a clear state cleanup action. So the `approve_if_odd` can be a stand-in for that. How would we want to enable such a call? Some ideas:
* Do we allow any arguments at all?
* If so, what kind of checks for the arguments should we generate? In particular, what happens if the 0'th argument - for the method selector - is incorrect?
* Should we allow an option that instead of issuing `err` issues `int 1; return` immediately instead?

These questions get kind of dicey, but I can imagine modifying `@router.method` (or maybe even introducing `@router.clear_method`) with some parameters that control this. EG:
* `for_clear: bool` (unneeded in the case of `@router.clear_method`) indicates whether this is intended for the clear program, and then enables the following parameters:
* `no_erroring = True` - in the case that the method selector (or any other parameter) is wrong, will issue an `int 1; return` instead of `err`. But this means that all the other `@router.clear_method` would need to have the same `no_erroring` value. So maybe the _routing err_ behavior belongs as a _router level option_.

tzaffi (commented):
In addition to a URL, I left a point of discussion. But overall, looking good.

pyteal/ast/router.py
ahangsu: I update the doc a little bit and referred to doc in [030eaf4](https://github.com/algorand/pyteal/pull/636/commits/030eaf4c3e3743587abdbd85a4b933bc8510b72a), keep looking into what should be done for csp refactoring.

pyteal/ast/router.py
ahangsu: made a push on [2575241](https://github.com/algorand/pyteal/pull/636/commits/2575241e66155778c6a41865e4793e25ce27dbea), so error message varies by provided args

pyteal/compiler/compiler_test.py
ahangsu: I think my take on the discussion is that, we might want to let router user to *explicitly* specify what kind of logic they want to use in clear state program (CSP), so my thoughts and opinions are as follows:

> Do we allow any arguments at all?

There is no way we are disallowing args, one can always use `Txn.application_args` in their composed `Expr`. We will just no longer generate code to handle the IO process in CSP.

> If so, what kind of checks for the arguments should we generate? In particular, what happens if the 0'th argument - for the method selector - is incorrect?

If we go down the path of letting users to specify whatever logic in CSP, then I don't see too good a way to generate checks for arguments.

For now, I haven't thought through how much flexibility and check can be done on our side, but rather, probably we want to hand the full control over CSP to router users?

pyteal/ast/router.py
```diff
@@ -637,21 +620,28 @@ def method(
             opt_in (optional): The allowed calls during :code:`OnComplete.OptIn`.
             close_out (optional): The allowed calls during :code:`OnComplete.CloseOut`.
             clear_state (optional): The allowed calls during :code:`OnComplete.ClearState`.
+                This argument has been deprecated, and will error on compile time if one wants to access it.
```
jasonpaulos, line 623: Can you mention that they should use the `clear_state` argument when creating a Router here as well?

pyteal/ast/router.py
```diff
@@ -637,21 +620,28 @@ def method(
...
         """
         # we use `is None` extensively for CallConfig to distinguish 2 following cases
         # - None
         # - CallConfig.Never
         # both cases evaluate to False in if statement.
+
+        if clear_state is not None:
+            raise TealInputError(
+                "Attempt to register ABI method for clear state program. "
+                "Use Router top level argument `clear_state` instead."
```
jasonpaulos, line 635: It looks like you added a link to the other error message but not this one?

pyteal/ast/router.py
```diff
@@ -302,11 +287,13 @@ def wrap_handler(
...
         """
         if not is_method_call:
+            wrap_to_name = "bare appcall" if wrap_to_name is None else wrap_to_name
+
             match handler:
                 case Expr():
                     if handler.type_of() != TealType.none:
                         raise TealInputError(
-                            f"bare appcall handler should be TealType.none not {handler.type_of()}."
+                            f"{wrap_to_name} handler should be TealType.none not {handler.type_of()}."
```
jasonpaulos, line 296: `wrap_to_name` is optional but you use it in the error string. Does that print "None"?

jasonpaulos:
I had some docs improvement ideas, so I opened #641

barnjamin (approved):

pyteal/ast/router.py
ahangsu: no, `wrap_to_name` is `None`, then converted to a default `"bare appcall"` on line https://github.com/algorand/pyteal/pull/636/files#diff-3844cf2bef1f9fc5ce0b421c65141cf76d172d885b5792ac0cd79dc360618769R290

pyteal/ast/router.py
ahangsu: good find, just add to the error message.

pyteal/ast/router.py
```diff
@@ -508,19 +496,28 @@ def __init__(
...
+                expression will be the entirety of the clear state program; no additional code is
+                inserted by the Router. If not provided, the clear state program will always reject.
         """
 
         self.name: str = name
         self.descr = descr
 
         self.approval_ast = ASTBuilder()
-        self.clear_state_ast = ASTBuilder()
+        self.clear_state: Expr = (
```
tzaffi, line 516: I believe this is backwards compatible in the sense that if we have a program written today with default behavior, it will generate an `int 0` for its clear. However, it is arguable that even in _this_ default case, we still want a _NEW_ behavior and produce `int 1` instead for the clear program.

And assuming we agree with this point, then unfortunately, the parameter will need to become mandatory because of the lack of backwards compatibility. So keeping it as kwarg-only param, we would need to have something like:

```python
assert clear_state, "... something similar to the other error messages above ..."
```

In either case, I encourage the creation of a tiny unit test (assuming it doesn't already exist) that will assert that the simplest compiling PyTeal program generates the approval and clear programs that we would like to generate.

tzaffi (commented):
Minor grammatical suggestions and a controversial suggestion to default to generating `int 1` for the clear program

pyteal/ast/router_test.py
```diff
@@ -361,6 +360,20 @@ def camel_to_snake(name: str) -> str:
...
+        pt.MethodConfig(clear_state=pt.CallConfig.CALL)
+
+
+def test_bare_call_config_clear_state_failure():
+    with pytest.raises(pt.TealInputError):
+        pt.BareCallActions(
+            clear_state=pt.OnCompleteAction(
+                action=pt.Seq(), call_config=pt.CallConfig.CALL
+            )
+        )
```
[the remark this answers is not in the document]
ahangsu: just add a testcase [router testcase](https://github.com/algorand/pyteal/pull/636/commits/814ab4973f919b20d2effdfd2a8d02e1a3ba249d)

pyteal/ast/router.py
```diff
@@ -96,6 +83,15 @@ class MethodConfig:
     update_application: CallConfig = field(kw_only=True, default=CallConfig.NEVER)
     delete_application: CallConfig = field(kw_only=True, default=CallConfig.NEVER)
 
+    def __post_init__(self):
+        if self.clear_state != CallConfig.NEVER:
+            raise TealInputError(
+                "Attempt to construct clear state program from MethodConfig, "
+                "use Router top level argument `clear_state` instead. "
+                "For more details please refer to "
+                "https://pyteal.readthedocs.io/en/latest/abi.html#registering-bare-app-calls"
```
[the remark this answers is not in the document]
ahangsu: 👍

pyteal/ast/router.py
tzaffi: Upon further reflection, I changed my mind.... `int 0` for the simplest possible clear program still makes sense because there is no state that is changing inside of the program and that needs to be preserved.

pyteal/ast/router.py
ahangsu: good catch, just add a line for referring using `clear_state` argument.

pyteal/ast/router.py
ahangsu: Thought for a bit, we still keep the backwards compatible way, right?

One question: even one call a CSP with `int 0`, then it failed, it still get its state cleared, right?

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Added
 
 ## Changed
+* Improved Router clear state program generation. ([#636](https://github.com/algorand/pyteal/pull/636))
```
jasonpaulos, line 6: Could you also explain that this contains a breaking change with respect to Router clear states?

jasonpaulos (commented):
It's nearly there for me, just waiting on these two comments

pyteal/ast/router_test.py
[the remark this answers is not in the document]
jasonpaulos: I can't find that commit, it must have gotten removed in a rebase. I still don't see error strings mentioned in the tests.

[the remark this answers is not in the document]
ahangsu: One of the latest commit https://github.com/algorand/pyteal/pull/636/commits/7003f85b2fe3cc08587a8ba00ae815365614febf should check the error message check for registering methods for csp

CHANGELOG.md
ahangsu: Please grammar police [d943b3f](https://github.com/algorand/pyteal/pull/636/commits/d943b3f0fd126c8a6d571efeab93758314f3bf2a)

jasonpaulos (approved):

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
...
+  * NOTE: a backwards incompatable change was imposed in this PR: previous Clear State Program (CSP) can be constructed in router by registering ABI methods or bare appcalls, now one has to use `clear_state` argument in `Router.__init__` to construct the CSP.
```
tzaffi, line 7:
```suggestion
  * NOTE: a backwards incompatable change was imposed in this PR: previous Clear State Program (CSP) can be constructed in router by registering ABI methods or bare app calls, now one has to use `clear_state` argument in `Router.__init__` to construct the CSP.
```

pyteal/ast/router.py
```diff
@@ -316,7 +303,7 @@ def wrap_handler(
                         )
                     if handler.subroutine.argument_count() != 0:
                         raise TealInputError(
-                            f"subroutine call should take 0 arg for bare-app call. "
+                            f"subroutine call should take 0 arg for {wrap_to_name}. "
```
tzaffi, line 306:
```suggestion
                            f"subroutine call should take 0 args for {wrap_to_name}. "
```

tzaffi (approved):
LGTM

PR #639, opened 2023-01-20
Non idempotence of router

tzaffi:
The PR in its current state attempts 2 changes.

The first -some minor refactoring of the router- is not very controversial.

~The second -resetting `ScratchSlot.nextSlotId` is causing flakey tests, and I will need to revert it unless I can find a more fundamental fix.  In particular, this assertion becomes flakey on my machine in with the 2nd change:~

https://github.com/algorand/pyteal/blob/a240ec01ccc0e20076be298a04f282a516e7267b/pyteal/compiler/optimizer/optimizer_test.py#L261

Seems like the current PR state is stable, but still requires careful consideration.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,7 +58,7 @@ test-unit-async:
 	pytest -n $(NUM_PROCS) --durations=10 -sv pyteal tests/unit -m "not serial"
 
 test-unit-sync:
-	pytest --durations=10 -sv pyteal tests/unit tests/integration -m serial
+	pytest --durations=10 -sv pyteal tests/unit -m serial
 
 test-unit: test-unit-async test-unit-sync
 
--- a/Makefile
+++ b/Makefile
@@ -52,10 +52,15 @@ lint: black flake8 mypy sdist-check
 # ---- Unit Tests (no algod) ---- #
 
 # TODO: add blackbox_test.py to multithreaded tests when following issue has been fixed https://github.com/algorand/pyteal/issues/199
+
 NUM_PROCS = auto
-test-unit:
-	pytest -n $(NUM_PROCS) --durations=10 -sv pyteal tests/unit --ignore tests/unit/blackbox_test.py --ignore tests/unit/user_guide_test.py
-	pytest -n 1 -sv tests/unit/blackbox_test.py tests/unit/user_guide_test.py
+test-unit-async:
+	pytest -n $(NUM_PROCS) --durations=10 -sv pyteal tests/unit -m "not serial"
+
+test-unit-sync:
+	pytest --durations=10 -sv pyteal tests/unit tests/integration -m serial
+
+test-unit: test-unit-async test-unit-sync
 
 lint-and-test: check-generate-init lint test-unit
 
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,6 +1,6 @@
 sphinx==5.1.1
 sphinx-rtd-theme==1.0.0
 # dependencies from setup.py
-py-algorand-sdk>=1.9.0,<2.0.0
+py-algorand-sdk>=2.0.0,<3.0.0
 semantic-version>=2.9.0,<3.0.0
 docstring-parser==0.14.1
--- a/pyteal/ast/router.py
+++ b/pyteal/ast/router.py
@@ -17,12 +17,14 @@ from pyteal.ir.ops import Mode
 from pyteal.ast import abi
 from pyteal.ast.subroutine import (
     OutputKwArgInfo,
+    Subroutine,
     SubroutineFnWrapper,
     ABIReturnSubroutine,
 )
 from pyteal.ast.assert_ import Assert
 from pyteal.ast.cond import Cond
 from pyteal.ast.expr import Expr
+from pyteal.ast.frame import FrameVar, Proto, ProtoStackLayout
 from pyteal.ast.app import OnComplete
 from pyteal.ast.int import Int, EnumInt
 from pyteal.ast.seq import Seq
@@ -71,7 +73,8 @@ class CallConfig(IntFlag):
                 return 1
             case CallConfig.CREATE | CallConfig.ALL:
                 raise TealInputError(
-                    "Only CallConfig.CALL or CallConfig.NEVER are valid for a clear state CallConfig, since clear state can never be invoked during creation"
+                    "Only CallConfig.CALL or CallConfig.NEVER are valid for a clear state CallConfig, "
+                    "since clear state can never be invoked during creation"
                 )
             case _:
                 raise TealInputError(f"unexpected CallConfig {self}")
@@ -130,6 +133,9 @@ class MethodConfig:
 
     def clear_state_cond(self) -> Expr | int:
         return self.clear_state.clear_state_condition_under_config()
+
+
+MethodConfig.__module__ = "pyteal"
 
 
 @dataclass(frozen=True)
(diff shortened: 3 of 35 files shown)


pyteal/ast/router_test.py
```diff
@@ -936,3 +938,32 @@ def handle(deposit: pt.abi.PaymentTransaction):  # noqa: F811
...
+
+    meth = router.add_method_handler(add)
+    assert meth.method_signature() == "add(uint64,uint64)uint64"
+
+    approval1, clear1, contract1 = router.compile_program(version=6)
+    approval2, clear2, contract2 = router.compile_program(version=6)
+
+    assert contract1.dictify() == contract2.dictify()
+    assert clear1 == clear2
+    assert approval1 == approval2
```
tzaffi, line 969: line 969 was failing for two reasons:
1. Unintended statefulness in the router that kept increasing compilation size. Every time `Router.build()` was called, it re-appended the routing conditions.
2. class variable `ScratchSlot.nextSlotId` keeps being incremented to during the life of program and doesn't reset after each compilation. This means that expressions that are recreated during compilation (such as the routing of bare-app calls) will keep getting assigned higher slot id's. The compiler may or not assign them to the same _finalized_ slot id's because of this artifact. By reseting `ScratchSlot.nextSlotId` hopefully this reduces the non-determinism. However, great care should be taken to assess whether the change is safe.

tzaffi:
> The PR in its current state attempts 2 changes.
> 
> The first -some minor refactoring of the router- is not very controversial.
> 
> The second -resetting `ScratchSlot.nextSlotId` is causing flakey tests, and I will need to revert it unless I can find a more fundamental fix. In particular, this assertion becomes flakey on my machine in with the 2nd change:
> 
> https://github.com/algorand/pyteal/blob/a240ec01ccc0e20076be298a04f282a516e7267b/pyteal/compiler/optimizer/optimizer_test.py#L261

This fix doesn't work as is because you end up overwriting previous assignments. I need to think more about it.

pyteal/ast/router.py
```diff
@@ -203,6 +204,10 @@ class BareCallActions:
         kw_only=True, default=OnCompleteAction.never()
     )
 
+    def __post_init__(self):
```
tzaffi, line 207: bring up the validation so can tell that bare calls are messed up before starting to build AST

pyteal/ast/router.py
```diff
@@ -309,10 +314,18 @@ def to_cond_node(self, use_frame_pt: bool = False) -> CondNode:
...
-    conditions_n_branches: list[CondNode] = field(default_factory=list)
-    methods_with_conds: list[CondWithMethod] = field(default_factory=list)
+    def __init__(self):
+        self.conditions_n_branches: list[CondNode] = []
+        self.methods_with_conds: list[CondWithMethod] = []
+        self._bare_cnbs: list[CondNode] = []
+
+    def _clean(self) -> None:
+        self._bare_cnbs = []
+        # self.conditions_n_branches = []
```
tzaffi, line 325: experimental crud... this is a draft PR after all

pyteal/ast/router.py
```diff
@@ -686,23 +699,11 @@ def __init__(
         self.method_sig_to_selector: dict[str, bytes] = dict()
         self.method_selector_to_sig: dict[bytes, str] = dict()
 
-        if bare_calls and not bare_calls.is_empty():
-            bare_call_approval = bare_calls.approval_construction()
-            if bare_call_approval:
```
tzaffi, line 691: moved to build phasse

pyteal/ast/router.py
```diff
@@ -877,6 +878,24 @@ def build_program(
             * clear_state_program: an AST for clear-state program
             * contract: a Python SDK Contract object to allow clients to make off-chain calls
         """
+        if self.bare_calls and not self.bare_calls.is_empty():
```
tzaffi, line 881: Was also considering of privatizing `build_program()`. What do you think?

pyteal/ast/router_test.py
```diff
@@ -936,3 +939,43 @@ def handle(deposit: pt.abi.PaymentTransaction):  # noqa: F811
...
+    # formerly nextSlotId: 265 --> 268:
+    approval3, clear3, contract3 = router.compile_program(version=6)
+
+    assert contract2.dictify() == contract3.dictify()
+    assert clear3 == clear2
+    assert approval3 == approval3
+
+    assert contract2.dictify() == contract1.dictify()
+    assert clear2 == clear1
+    assert (
```
tzaffi, line 978: This was the actual assert that was failing (after I fixed the simple bug)

Makefile
```diff
@@ -52,10 +52,15 @@ lint: black flake8 mypy sdist-check
 # ---- Unit Tests (no algod) ---- #
 
 # TODO: add blackbox_test.py to multithreaded tests when following issue has been fixed https://github.com/algorand/pyteal/issues/199
```
tzaffi, line 54: Is this tackleable now?

docs/requirements.txt
```diff
@@ -1,6 +1,6 @@
 sphinx==5.1.1
 sphinx-rtd-theme==1.0.0
 # dependencies from setup.py
-py-algorand-sdk>=1.9.0,<2.0.0
+py-algorand-sdk>=2.0.0,<3.0.0
```
tzaffi, line 4: ghost

pyteal/ast/router_test.py
```diff
@@ -825,14 +827,15 @@ def test_build_program_clear_state_empty():
...
-        )
         with pytest.raises(
             pt.TealInputError,
             match=r"Only CallConfig.CALL or CallConfig.NEVER are valid for a clear state CallConfig, since clear state can never be invoked during creation$",
         ):
-            pt.Router("test", bareCalls)
+            bareCalls = pt.BareCallActions(
+                clear_state=pt.OnCompleteAction(action=pt.Approve(), call_config=config)
+            )
+            _ = bareCalls  # temp linter hack
```
tzaffi, line 837: temp experimental change

pyteal/ast/router_test.py
```diff
@@ -936,3 +939,43 @@ def handle(deposit: pt.abi.PaymentTransaction):  # noqa: F811
     ap2, cs2, c2 = r2.compile_program(version=pt.compiler.MAX_PROGRAM_VERSION)
 
     assert (ap1, cs1, c1) == (ap2, cs2, c2)
+
+
+def test_router_build_idempotence():
```
tzaffi, line 944:
```suggestion
def test_router_compile_program_idempotence():
```

release-process.md
```diff
@@ -0,0 +1,81 @@
+# PyTeal Release Process
```
tzaffi, line 1: ghost

setup.py
```diff
@@ -7,7 +7,7 @@
 
 setuptools.setup(
     name="pyteal",
-    version="0.20.1",
+    version="0.21.0",
```
tzaffi, line 10: ghost

docs/requirements.txt
ahangsu: merge fixed

setup.py
ahangsu: merge fixed

pyteal/ast/router.py
ahangsu: yes, I up the vote for privatizing `build_program`, for I hope *for now* folks are using only generated code from router, rather than the composed Expr result...

pyteal/ast/router.py
ahangsu: what is the purpose of this `_clean` function?

pyteal/ast/router.py
ahangsu: and if we go down the path of rewinding the scratch slot id, we should not expect people to continue compiling in the same session of compiling such router program, or they will end up using same scratch slots, which will cause trouble...

tzaffi: Yes, let's chat about it.

tzaffi (commented):
somehow I made my comments "pending"

pyteal/compiler/compiler.py
```diff
@@ -2,17 +2,7 @@
...
-    Expr,
-    Return,
-    Seq,
-    SubroutineDefinition,
-    SubroutineDeclaration,
-)
-from pyteal.ir import Mode, Op, TealComponent, TealOp, TealBlock, TealSimpleBlock
-from pyteal.errors import TealInputError, TealInternalError
-
+from pyteal.ast import Expr, Return, Seq, SubroutineDefinition, SubroutineDeclaration
```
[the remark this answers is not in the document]
ahangsu: 👍

pyteal/ast/router.py
tzaffi: Hmmm... this could be wrong...

pyteal/ast/router_test.py
tzaffi: TODO: change this to use router's top level `clear_state`

pyteal/ast/router.py
```diff
@@ -913,22 +927,27 @@ def compile_program(
...
-        csp_compiled = compileTeal(
-            csp,
-            Mode.Application,
-            version=version,
-            assembleConstants=assemble_constants,
-            optimize=optimize,
-        )
-        return ap_compiled, csp_compiled, contract
+        starting_slot_id = ScratchSlot.nextSlotId
+        try:
```
ahangsu, line 931: nit: I am not too happy about using `try...finally` here. Is it possible to somehow find a hidden boolean indicator to say the current router is "contaminated" or not, then raise error here?

ahangsu: filing a pr here #649

pyteal/ast/router.py
tzaffi: Resolving as discussed

Makefile
tzaffi: Am not tackling this as part of the PR.

pyteal/ast/router.py
```diff
@@ -632,7 +638,7 @@ def add_method_to_ast(
         self.methods_with_conds.append(CondWithMethod(method_signature, cond, handler))
 
     def program_construction(self, use_frame_pt: bool = False) -> Expr:
-        self.conditions_n_branches += [
+        self.conditions_n_branches = [cnb for cnb in self._bare_cnbs] + [
```
ahangsu, line 641: nit: why not just
```suggestion
        self.conditions_n_branches = self._bare_cnbs + [
```

ahangsu (approved):

PR #643, opened 2023-01-21
Var type

barnjamin:
Talked to @ahangsu about this, this is just one idea but doesn't always play nice


pyteal/ast/var.py
```diff
@@ -0,0 +1,119 @@
...
+        self.type = val.type_of()
+        return self.slot.store(val)
+
+    def has_return(self):
+        return False
+
+    def type_of(self):
+        return self.type
+
+    def __add__(self, other: Expr) -> Expr:
```
tzaffi, line 108: I guess we could load other operators as well

* `*` - multiplication for `uint64`, repetition for `self: bytes, other: uint64`

Not sure about the other operators, but it's worth thinking about.

PR #649, opened 2023-01-24
Refactoring non-idempotent fix with contextmanager a bit

ahangsu:


pyteal/ast/router.py
```diff
@@ -900,6 +901,13 @@ def _build_program(
             self.contract_construct(),
         )
 
+    @contextmanager
+    def __revert_side_effect(self):
```
tzaffi, line 905:
```suggestion
    def _cleaning_context(self):
```

nit: proposing slightly more unified naming

tzaffi (approved):
This is much better, thank you! (but a small nit suggesting name change)

pyteal/ast/router.py
ahangsu: sure, that makes sense

tzaffi: Did you inted `_revert_side_effect` rather than `_cleaning_context`. Either is fine (even if I suggested the latter)

tzaffi (approved):

PR #651, opened 2023-01-26
Minor refactoring on non idempotent router

ahangsu:


pyteal/ast/router.py
```diff
@@ -297,12 +297,11 @@ def to_cond_node(self, use_frame_pt: bool = False) -> CondNode:
 
 class ASTBuilder:
     def __init__(self):
-        self.conditions_n_branches: list[CondNode] = []
```
ahangsu, line 300: `self.conditions_n_branches` are used (updated) only in `program_construction`, so we only keep things that leads to computation result, rather than intermediate artifact.

pyteal/ast/router.py
```diff
@@ -882,12 +881,12 @@ def _build_program(
         if self.bare_calls and not self.bare_calls.is_empty():
             bare_call_approval = self.bare_calls.approval_construction()
             if bare_call_approval:
-                self.approval_ast._bare_cnbs.append(
+                self.approval_ast.bare_calls = [
```
ahangsu, line 884: `bare_calls` is a `Maybe` analogy in haskell, either empty or one element here.

pyteal/ast/router.py
```diff
@@ -297,12 +297,11 @@ def to_cond_node(self, use_frame_pt: bool = False) -> CondNode:
 
 class ASTBuilder:
     def __init__(self):
-        self.conditions_n_branches: list[CondNode] = []
         self.methods_with_conds: list[CondWithMethod] = []
-        self._bare_cnbs: list[CondNode] = []
+        self.bare_calls: list[CondNode] = []
```
ahangsu, line 301: I turned `_bare_cnbs` into `bare_calls` for `ASTBuilder` is not exported, and we should be fine with a not hidden member in a not exported class...

tzaffi (approved):

PR #650, opened 2023-01-26
Teal to PyTeal Source Mapper

tzaffi:
# Source Mapper for PyTeal

## Notes for PR Reviewers

The most important changes to gain understanding for the workings of the source mapper are in:
* `pyteal/ast/router.py`
* `pyteal/compiler/compiler.py`
* `pyteal/compiler/sourcemap.py`
* `pyteal/stack_frame.py`

~I won't make~ I'm trying to minimize major changes to those files until I receive review comments. However, there may be small changes that result from cleaning up other files or merging in changes to upstream branches. I'll make sure to add a new comment and tag the reviewers every time such a change occurs.

### What do I need to do in order to make source mapping available in PyTeal ?

You'll need to turn it on in a `pyteal.ini` file that will live at the top level of your python repo.
Moreover, the`pyteal.ini` should look like:

```ini
[pyteal]

[pyteal-source-mapper]
enabled = True
debug = False
```

### What if as a consumer of PyTeal, I _don't_ supply a `pyteal.ini` ?

Everything should continue to work. **HOWEVER**, some error messages will be printed out. For example, when I experimented running AlgoKit I got the following printouts, before I added a `pyteal.ini`:

```sh
❯  ... -- -m smart_contracts
Turning off frame capture and disabling sourcemaps. 
Could not read section (pyteal-source-mapper, enabled) of config "pyteal.ini": No section: 'pyteal-source-mapper'
Disabling `debug` status for sourcemaps. 
Could not read section (pyteal-source-mapper, debug) of config "pyteal.ini": No section: 'pyteal-source-mapper'
```

### What if I want to map program counters (PC's) as well?
You'll need to ensure that a dev-mode-enabled algod is running on port 4001 (for example, this is the Sandbox's default port). 

### How do I get source mapping info using the `Router` ?

`class Router` has a new `compile()` method which returns a `RouterResults` object. (The API of the original `compile_program()` method remains the same). For example, usage might look like:
```python
router = Router(name="GreatDapp")

@router.method
def add(a: abi.Uint64, b: abi.Uint64, *, output: abi.Uint64) -> Expr:
    return output.set(a.get() + b.get())

results: RouterResults = router.compile(
    version=8,
    with_sourcemaps=True,
    pcs_in_sourcemap=True,
    annotate_teal=True,
    annotate_teal_headers=True,
    annotate_teal_concise=True,
)

# because `with_sourcemaps == pcs_in_sourcemap == annotate_teal == True`
# `results` is "full" and the following assertions will hold

[...]

PR #656, opened 2023-02-01
Adding `asdict()` to Router's `BareCallActions`

tzaffi:
# Summary

In anticipation to some changes being brought on by #650, and wanting to maintain compatibility with Beaker, I'm proposing to explicitly provide the underlying dictionary.

## Beaker Companion PR
https://github.com/algorand-devrel/beaker/pull/190

### DIFF
--- a/pyteal/ast/router.py
+++ b/pyteal/ast/router.py
@@ -197,6 +197,16 @@ class BareCallActions:
                 "https://pyteal.readthedocs.io/en/latest/abi.html#registering-bare-app-calls"
             )
 
+    def asdict(self) -> dict[str, OnCompleteAction]:
+        return {
+            "clear_state": self.clear_state,
+            "close_out": self.close_out,
+            "delete_application": self.delete_application,
+            "no_op": self.no_op,
+            "opt_in": self.opt_in,
+            "update_application": self.update_application,
+        }
+
     def is_empty(self) -> bool:
         for action_field in fields(self):
             action: OnCompleteAction = getattr(self, action_field.name)
--- a/pyteal/ast/router_test.py
+++ b/pyteal/ast/router_test.py
@@ -378,6 +378,35 @@ def test_bare_call_config_clear_state_failure():
     assert "Attempt to construct clear state program from bare app call" in str(tie)
 
 
+def test_bare_call_actions_asdict():
+    no_action = pt.OnCompleteAction()
+    del_action = pt.OnCompleteAction(action=pt.Int(1), call_config=pt.CallConfig.ALL)
+    close_action = pt.OnCompleteAction(action=pt.Int(2), call_config=pt.CallConfig.CALL)
+
+    bca = pt.BareCallActions(
+        delete_application=del_action,
+        close_out=close_action,
+    )
+
+    bcad = bca.asdict()
+    assert set(bcad.keys()) == {
+        "clear_state",
+        "close_out",
+        "delete_application",
+        "no_op",
+        "opt_in",
+        "update_application",
+    }
+    assert bcad == {
+        "clear_state": no_action,
+        "close_out": close_action,
+        "delete_application": del_action,
+        "no_op": no_action,
+        "opt_in": no_action,
+        "update_application": no_action,
+    }
+
+
 def test_router_register_method_clear_state_failure():
     router = pt.Router("doomedToFail")
 
--- a/pyteal/ast/router_test.py
+++ b/pyteal/ast/router_test.py
@@ -397,14 +397,15 @@ def test_bare_call_actions_asdict():
         "opt_in",
         "update_application",
     }
-    assert bcad == {
-        "clear_state": no_action,
-        "close_out": close_action,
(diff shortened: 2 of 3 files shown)


pyteal/ast/router.py
```diff
@@ -197,6 +197,16 @@ def __post_init__(self):
                 "https://pyteal.readthedocs.io/en/latest/abi.html#registering-bare-app-calls"
             )
 
+    def asdict(self) -> dict[str, OnCompleteAction]:
```
ahangsu, line 200: nit: this is a getter method, so we can consider making it `@property`?

pyteal/ast/router.py
```diff
@@ -197,6 +197,16 @@ def __post_init__(self):
...
+        return {
+            "clear_state": self.clear_state,
```
ahangsu, line 202: nit: for introduction of #638, bare-appcalls no longer take `clear_state` field, so do we want to keep this field in dict?

ahangsu (commented):
generally lgtm, just 2 questions on previous introductions of CSP change

barnjamin (approved):

ahangsu:
also nit reminder: add a line in `CHANGELOG`

pyteal/ast/router.py
tzaffi: The pattern of `asdict()` being a method is widespread in the Python world, so I'm inclined to stay with this approach. However, your suggestion is a reasonable one and we do use `@property`'s in several other places, for more custom properties such as `TxnObject.assets`. It might be worthwhile synthesizing when we should/shouldn't use properties and add it to our [code guidelines](https://github.com/algorand/pyteal/blob/005d8a618b652adca90f413dc0ffae303615e981/CONTRIBUTING.md#L19).

pyteal/ast/router.py
tzaffi: @barnjamin - in your Beaker use case, do you still need the `clear_state` coming from this dict?
@ahangsu - does it actually get populated?

pyteal/ast/router.py
ahangsu: ok, I am just suggesting a potential move, but I don't want to catch too much attention on code guidelines, so I am resolving here.

pyteal/ast/router.py
ahangsu: no it does not get populated. all of them get removed in #636

tzaffi: I'm starting to worry that the delayed computation I was pushing for in pyteal may not play well with Beaker. We should make sure that our current `master` doesn't break it: @barnjamin

tzaffi: Actually, more likely the issue would be with https://github.com/algorand/pyteal/tree/feature/fp-router... I'm juggling too many branches in my head.

ahangsu: I don't understand a few terms and sentences above:
> does it actually get populated?

Do you mean related CSP AST get populated, or you mean dataclass `MethodConfig.clear_state`?
If you mean latter one, it is kept simply because of backwards compatibility, but it is not used in any code generation.

> delayed computation

What is this?

pyteal/ast/router_test.py
```diff
@@ -378,6 +378,36 @@ def test_bare_call_config_clear_state_failure():
...
+    bcad = bca.asdict()
+    assert set(bcad.keys()) == {
+        "clear_state",
+        "close_out",
+        "delete_application",
+        "no_op",
+        "opt_in",
+        "update_application",
+    }
+    with pt.TealComponent.Context.ignoreExprEquality():
```
jasonpaulos, line 400: I believe we'd want to test without this, since the expressions should be the same

jasonpaulos (approved):

pyteal/ast/router.py
tzaffi: I'm keeping this discussion open, even as I merge this PR. I'm diving a little deeper into the Beaker weeds to see if anything I said above could lead to an actual problem.

tzaffi: After investigating a bit, I feel more confidence that my worries are unfounded. The conflict that #650 was causing with the former usage of `BareCallActions.__as_dict__` is occuring on an[ independent usage of this class](https://github.com/algorand-devrel/beaker/blob/5befaf887d5f10dcc84475447d0a7194966af5c7/beaker/decorators.py#L554) that's not in the router. Router usage is limited to [exactly when the compilation occurs](https://github.com/algorand-devrel/beaker/blob/5befaf887d5f10dcc84475447d0a7194966af5c7/beaker/application.py#L285-L310). Therefore, Beaker views the Router as a black-box and any refactoring we do of the Router's compilation ought not have any side effects in Beaker.

I'm closing this discussion. However, @barnjamin , feel free to re-open if you'd like to follow up in some way.

ahangsu: ok, thanks for looking into it! This is also the source of my concern and seems nothing's breaking!

PR #661, opened 2023-02-07
Added TealSeqError for better error reporting.

BarryJRowe:
Syntax errors on `Seq` arrays do not clearly indicate where the syntax error is when the return type of a non-final expression is not `TealType.none` and will only point to the parent `Seq` statement. This change will reference the offending statement in the error message.

This code:
```
Seq([
    InnerTxnBuilder.Begin(),
    InnerTxnBuilder.SetFields({
        TxnField.type_enum: TxnType.AssetTransfer,
        TxnField.xfer_asset: Txn.assets[0],
        TxnField.asset_amount: Int(0) ,
        TxnField.asset_receiver: Global.current_application_address()
    }),
    InnerTxnBuilder.Submit(),
    #create box:
    App.box_create(Itob(Txn.assets[0]),Int(16)),
    App.globalPut(Bytes("AssetsRatioTotal"),  Btoi(Txn.application_args[1])),
])
```
Will now produce this error:
```
  ...
  File "/home/user/my_contract.py", line 210, in generate_approval_program_v1
    Seq([
  File "/home/user/ENV3/lib/python3.10/site-packages/pyteal/ast/seq.py", line 60, in __init__
    raise e
  File "/home/user/ENV3/lib/python3.10/site-packages/pyteal/ast/seq.py", line 55, in __init__
    require_type(expr, TealType.none)
  File "/home/user/ENV3/lib/python3.10/site-packages/pyteal/types.py", line 38, in require_type
    raise TealTypeError(actual, expected)
pyteal.TealSeqError: (box_create (Itob (Txna Assets 0)) (Int 16)) must have a return type of TealType.none. Only the last entry of a Seq array can have a return value.
```

### DIFF
--- a/pyteal/ast/seq.py
+++ b/pyteal/ast/seq.py
@@ -1,7 +1,7 @@
 from typing import List, TYPE_CHECKING, overload
 
 from pyteal.types import TealType, require_type
-from pyteal.errors import TealInputError
+from pyteal.errors import TealInputError, TealTypeError, TealSeqError
 from pyteal.ir import TealSimpleBlock
 from pyteal.ast.expr import Expr
 
@@ -51,7 +51,17 @@ class Seq(Expr):
             if not isinstance(expr, Expr):
                 raise TealInputError("{} is not a pyteal expression.".format(expr))
             if i + 1 < len(exprs):
-                require_type(expr, TealType.none)
+                try:
+                    require_type(expr, TealType.none)
+                except TealTypeError:
+                    message = "{} must have a return type of TealType.none. ".format(
+                        expr
+                    )
+                    message += (
+                        "Only the last entry of a Seq array can have a return value."
+                    )
+                    seq_error = TealSeqError(message)
+                    raise seq_error
 
         self.args = exprs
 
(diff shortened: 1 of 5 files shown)


pyteal/ast/seq.py
```diff
@@ -51,7 +51,17 @@ def __init__(self, *exprs):
...
+                try:
+                    require_type(expr, TealType.none)
+                except TealTypeError as e:
+                    e.__class__ = TealSeqError
+                    e.message = "{} must have a return type of TealType.none. ".format(
+                        expr
+                    )
+                    e.message += (
+                        "Only the last entry of a Seq array can have a return value."
+                    )
```
jasonpaulos, line 63: This PR seems like a good change but I dislike this method of modifying the error object. Can you either:
* Create a new `TealSeqError` here and raise it
* Use a new function instead of `require_type` which raises the right error in the first place

pyteal/errors.py
```diff
@@ -15,6 +15,17 @@ def __str__(self):
 TealInternalError.__module__ = "pyteal"
 
 
+class TealSeqError(Exception):
```
jasonpaulos, line 18: It would be nice if this could inherit from `TealTypeError`, since it's a more specific type error

ahangsu (approved):
generally lgtm

jasonpaulos (approved):

PR #663, opened 2023-02-09
Tuple Index Into Minor Fix

ahangsu:


tzaffi:
I propose an additional test for `abi.Tuple2[abi.DynamicBytes, abi.StaticBytes[Literal[3]]`  near [this case](https://github.com/algorand/pyteal/blob/b58c5033676fe6ee4df6ed53025bdbfeff52ead9/tests/integration/abi_roundtrip_test.py#L79)

@bbroder-algo

pyteal/ast/abi/tuple.py
```diff
@@ -198,9 +198,6 @@ def _index_tuple(
         if offset == 0:
```
bbroder-algo, line 198: combine?

ahangsu: yes, doable, need to rephrase comments a bit.

tests/integration/teal/roundtrip/app_roundtrip_(byte[4],(bool,bool),uint64,address)[]_7_v6.teal
```diff
@@ -25,7 +25,7 @@ int 5
 extract_uint64
 store 2
 load 19
-extract 13 0
+extract 13 32
```
Comment (tzaffi) [inline tests/integration/teal/roundtrip/app_roundtrip_(byte[4],(bool,bool),uint64,address)[]_7_v6.teal:28]: So was the integration test giving a false negative?

ahangsu:
> I propose an additional test for abi.Tuple2[abi.DynamicBytes, abi.StaticBytes[Literal[3]] near [this case](https://github.com/algorand/pyteal/blob/b58c5033676fe6ee4df6ed53025bdbfeff52ead9/tests/integration/abi_roundtrip_test.py#L79)

yes, working on it

tests/integration/teal/roundtrip/app_roundtrip_(byte[4],(bool,bool),uint64,address)[]_7_v6.teal
Comment (ahangsu) [reply tests/integration/teal/roundtrip/app_roundtrip_(byte[4],(bool,bool),uint64,address)[]_7_v6.teal:28]: I don't think integration test (in general) is false negative, but the testcase is really a corner case we didn't account for.

Comment (tzaffi) [reply tests/integration/teal/roundtrip/app_roundtrip_(byte[4],(bool,bool),uint64,address)[]_7_v6.teal:28]: I think this is extracting the acddress which is at the end of the encoding, end therefore `extract 13 0` $\++$ `extract 13 32`. Does that make sense given your changes?

pyteal/ast/abi/tuple.py
```diff
@@ -194,15 +194,10 @@ def _index_tuple(
...
 
-    if index + 1 == len(value_types):
-        if offset == 0:
+    if offset == 0:
+        if index + 1 == len(value_types):
             # This is the first and only value in the tuple, so decode all of encoded
             return output.decode(encoded)
-        # This is the last value in the tuple, so decode the substring from start_index to the end of
-        # encoded
-        return output.decode(encoded, start_index=start_index)
```
jasonpaulos, line 203: I think this case is still correct for tuples of all static types. Do you think we should keep it for that case?

tests/integration/teal/roundtrip/app_roundtrip_(byte[4],(bool,bool),uint64,address)[]_7_v6.teal
Comment (ahangsu) [reply tests/integration/teal/roundtrip/app_roundtrip_(byte[4],(bool,bool),uint64,address)[]_7_v6.teal:28]: Okay, this two cases are equal only under a precondition, the last element in tuple (address in this case) has no leading dynamic type(s).

If we both agree on this precondition, then yes, it makes sense given my change.

pyteal/ast/abi/tuple_test.py
```diff
@@ -371,6 +371,15 @@ class IndexTest(NamedTuple):
                 encoded, start_index=pt.ExtractUint16(encoded, pt.Int(0))
             ),
         ),
+        IndexTest(
```
tzaffi, line 374: I tried this test case against `master` (and also the graviton roundtrip test that I proposed) and both tests passed. I'd like to get a better understanding of why that is. For the graviton roundtrip test, I'm guess the test is passing because its decoding the entire tuple as a posed to a portion of it.

pyteal/ast/abi/tuple.py
ahangsu: actually, by this discussion thread: https://github.com/algorand/pyteal/pull/663#discussion_r1102015352, I am pushing back a bit.

pyteal/ast/abi/tuple_test.py
ahangsu: good point on reverse testing on `master`, I am also trying to replicate things on `master` and will get back to you what I have.

ahangsu: ok, my first find: this is INDEED an insufficient test: for we are taking from byte, rather than a static array (with static type). So the compilation will use `getbyte`, which makes no difference with or without length.

I propose we should introduce another type spec (like static bytes) for this testcase.

ahangsu: made a change in https://github.com/algorand/pyteal/pull/663/commits/7bc748d4c7c38adbc139443726c65637aa2449ad, and it should be failing on master

tzaffi (approved):
LGTM

pyteal/ast/abi/tuple.py
```diff
@@ -198,9 +198,9 @@ def _index_tuple(
         if offset == 0:
             # This is the first and only value in the tuple, so decode all of encoded
             return output.decode(encoded)
-        # This is the last value in the tuple, so decode the substring from start_index to the end of
-        # encoded
-        return output.decode(encoded, start_index=start_index)
+        elif all(not x.is_dynamic() for x in value_types):
```
jasonpaulos, line 201: nit: `elif` not required because the previous if will always return.
```suggestion
        if all(not x.is_dynamic() for x in value_types):
```

pyteal/ast/abi/tuple_test.py
```diff
@@ -371,6 +372,15 @@ class IndexTest(NamedTuple):
...
+            types=[dynamic_array_t1, static_bytes_t],
+            typeIndex=1,
+            expected=lambda output: output.decode(
+                encoded,
+                start_index=pt.Int(2),
+                length=pt.Int(2),
+            ),
+        ),
         IndexTest(
```
jasonpaulos, line 384: Could you also add a test case to add coverage for indexing the last element of static tuples? I think something like this would work (but not tested):

```python
        IndexTest(
            types=[static_bytes_t, static_bytes_t],
            typeIndex=1,
            expected=lambda output: output.decode(
                encoded,
                start_index=pt.Int(2),
            ),
        ),
```

jasonpaulos (commented):
Nice improvement! I just have some small requests

pyteal/ast/abi/tuple_test.py
ahangsu: yes, the testcase your proposed is correct, see [8fb2e33](https://github.com/algorand/pyteal/pull/663/commits/8fb2e33f8448ddd58668855a2a71df0982a7678b)

pyteal/ast/abi/tuple.py
ahangsu: [8fb2e33](https://github.com/algorand/pyteal/pull/663/commits/8fb2e33f8448ddd58668855a2a71df0982a7678b)

jasonpaulos (approved):

bbroder-algo (commented):
approved

PR #666, opened 2023-02-13
Experimental Nightly Build

tzaffi:
This is the first of a sequence of PR's which I'm using to hone in on creating nightly tests.

At the end of the experiment, I'll create a PR to revert to the current release tag (though I'll assess any other commits to ensure I don't revert anyone elses work).

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,41 @@
+name: "Nightly Build"
+on:
+  schedule:
+    - cron: "10 * * * *"
+
+jobs:
+  check-date:
+    runs-on: ubuntu-20.04
+    name: Check latest commit
+    outputs:
+      should_run: ${{ steps.should_run.outputs.should_run }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: print latest_commit
+        run: echo ${{ github.sha }}
+
+      - id: should_run
+        continue-on-error: true
+        name: check latest commit is less than a day
+        if: ${{ github.event_name == 'schedule' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly.yml
```diff
@@ -0,0 +1,41 @@
+name: "Nightly Build"
+on:
+  schedule:
+    - cron: "10 * * * *"
```
tzaffi, line 4: As configured, this will run at 10 past every hour (that is not the long term plan, just to make testing easier)

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,41 @@
...
+
+jobs:
+  check-date:
```
tzaffi, line 7: This job is supposed to make it possible to stop running the job if there was no change to the branch in the last 24 hours.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,41 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: print latest_commit
+        run: echo ${{ github.sha }}
+
+      - id: should_run
+        continue-on-error: true
+        name: check latest commit is less than a day
+        if: ${{ github.event_name == 'schedule' }}
+        run: test -z $(git rev-list  --after="24 hours"  ${{ github.sha }}) && echo "::set-output name=should_run::false"
```
bbroder-algo, line 21: nifty

tzaffi: I have no idea if it actually works. Stack Overflow thumbs are the only evidence.

bbroder-algo (approved):

bbroder-algo:
looks good, I verified the cron spec

PR #676, opened 2023-02-18
Better router mapping

tzaffi:
## PLEASE DISREGARD THIS PR


pyteal/stack_frame.py
```diff
@@ -365,9 +407,16 @@ def _walk_asts(
                     walker_args = [expr.value]
                     if expr.index_expression:
                         walker_args.append(expr.index_expression)
+                case SubroutineCall():
+                    subdef = expr.subroutine.get_declaration_by_option(False)
+                    if subdef:
+                        walker_args.append(subdef)
+                    subdef = expr.subroutine.get_declaration_by_option(True)
+                    if subdef:
+                        walker_args.append(subdef)
```
tzaffi, line 416: @ahangsu - introducing these lines of code into the sourcemapper, actually changed the resulting teal. I'm trying to understand why.

pyteal/compiler/flatten.py
```diff
@@ -123,7 +144,13 @@ def flattenSubroutines(
             if isinstance(stmt, TealLabel):
                 stmt.getLabelRef().addPrefix(labelPrefix)
 
-        combinedOps.append(TealLabel(None, LabelReference(label), comment))
+        dexpr: Expr | None = None
+        try:
+            dexpr = subroutine.declarations.get_declaration()
```
tzaffi, line 149: deprecated

pyteal/stack_frame.py
ahangsu: seems fixed? For we introduced a fix #684 and the problem should go away

pyteal/compiler/flatten.py
ahangsu: I believe this is also relieved in #650

pyteal/stack_frame.py
tzaffi: Absolutely. Please disregard this stale comment. Also, I'm about to close this draft in favor of #689

Issue #677, opened 2023-02-23
Source Mapper Improvements: Better Performance and Its Consequences


tzaffi:
## Problem

The source mapper is very slow and a memory hog. As a result, two non-ideal artifacts have been added:

1. `pyteal.ini` which defaults to `pyteal-source-mapper.enabled = False`
2. Nightly tests in C.I. which carve out the slowest portion of the tests and complicate local testing

## Solution

@jasonpaulos has an [experimental approach](https://github.com/jasonpaulos/pyteal/tree/source-map-performance) which promises to speed things up to the point that these artifacts can be eliminated, or mitigated.

I propose:
* nightly tests be eliminated once it is established that all tests can be run in a reasonable amount of time
* Default the sourcemapper to `enabled = True` in the `pyteal.ini` but don't get rid of it

### Why keep the `pyteal.ini` ?
* It may still be useful to allow dev's to turn the feature off completely.
* We may desire to feature gate (cf. #601 ) via `pyteal.ini`. However, a more modern approach uses `pyproject.toml` instead.

## Dependencies

None

## Urgency

Medium - defaulting to `pyteal-source-mapper.enabled = True` will make source mapping a lot more accessible, and getting rid of nightly tests will reduce developer friction.

Issue #601, opened 2022-11-14
Prototype:  Feature gate support


michaeldiamant:
## Problem

Taking inspiration from https://blog.rust-lang.org/2014/10/30/Stability.html, we'd like to prototype a feature gating mechanism like https://rustc-dev-guide.rust-lang.org/feature-gates.html.  A feature gating system is one way for PyTeal to support faster velocity with fewer support guarantees while not sacrificing quality. 

The story's intent is to gauge implementation + maintenance challenges associated with supporting a feature gating system in PyTeal.  At the story's conclusion, we'd like to make a yes/no decision on supporting the approach as part of instituting a broader stability policy.

The feature gating mechanism does _not_ need to mirror Rust exactly.  Criteria we want to support include:
* Define a way to demarcate modules, classes, methods, and variables as experimental.  Access to experimental capabilities is controlled by PyTeal release version.  Only non-final (e.g. nightly builds, branch builds) versions can access experimental capabilities. 
* Optionally, providing an explicit opt-in mechanism to use experimental features.

It's understood that we'll negotiate the effort while underway.  We may realize some objectives are untenable.  Ultimately, if the approach isn't tenable, we want to gauge where to go next.

## Solution

Compiling links that may be helpful:
* https://stackoverflow.com/questions/8951787/defining-python-decorators-for-a-complete-module - The approach described here _might_ benefit our own implementation.
* https://github.com/carta/flipper-client - Example Python feature flag library.  It _might_ help us brainstorm our own approaches.

tzaffi:
I ran into a related consideration in #552 . The PR is proposing to enable a new feature -source mapping- while having it unavailable -by default. To turn the feature on, someone who clones the repo, or a depending library (in my case [Beaker](https://github.com/algorand-devrel/beaker/pull/117/files)) needs to explicitly set a config variable _**on**_ in the [introduced pyteal.ini](https://github.com/algorand/pyteal/blob/a59e4b678eace2a9c80ec83d417b00bfe3020fed/pyteal.ini):

<img width="196" alt="image" src="https://user-images.githubusercontent.com/291133/201776228-0f74ea8c-86a8-443e-9065-cbc6c6637c41.png">

I'm not sure the approach I took is the perfect solution. There were some definite downsides:
* I needed to assert that the feature was turned on, before attempting to use it
* I needed to introduce tests that asserted that the feature was off by default
* For tests of the actual feature, I needed to monkeypatch the config to fool the test into believing that the feature was on
[...]

PR #686, opened 2023-03-09
Feature Gates Starting with Sourcemap

tzaffi:
Addressing the pyteal.ini portion of https://github.com/algorand/pyteal/issues/677 as well as proposing a way to address https://github.com/algorand/pyteal/issues/601

Please see `examples/applications/sourcemap.py` for usage example.

## Regarding Feature Gates Issue [601](https://github.com/algorand/pyteal/issues/601)

`feature_gates.FeatureGates` provides a mechanism for defining global booleans that are loaded before the rest of PyTeal is. An experimental code path could directly use these globals via `FeatureGates.theFeature_isOn()` or I can also imagine a decorator such as `@FeatureGate.decorate_theFeature_isOn` which actually runs an `assert FeatureGates.theFeature_isOn()` as part of the wrapped method. To avail the feature in a particular build, one could call `FeatureGate.set_theFeature_isOn(True)` and to officially release (without removing the decorators) is as simple as changing `FeatureGates.DEFAULTS.theFeature_isOn`

>Optionally, providing an explicit opt-in mechanism to use experimental features.

That's available via -for example- `FeatureGate.set_theFeature_isOn(True)`


examples/application/sourcemap.py
```diff
@@ -0,0 +1,65 @@
+"""
+This example shows how to use the sourcemap feature to generate a source map for a PyTeal program.
+
+For this example, no `pyteal.ini` file is assumed and to enable the feature
+one must import `FeatureGate` from `feature_gates` and set `FeatureGate.sourcemap_enabled` to `True`.
+This import must occur before any object from PyTeal is imported, as PyTeal will
+default to it's own sourcemap configuration if not set before hand.
```
tzaffi, line 7:
```suggestion
default to its own sourcemap configuration if not set before hand.
```

examples/application/vote_deploy.py
```diff
@@ -16,6 +16,7 @@
 algod_address = "http://localhost:4001"
```
tzaffi, line 16: Changes due to bumping minor version of `black`

feature_gates/__init__.py
```diff
@@ -0,0 +1,136 @@
+from dataclasses import dataclass
+from configparser import ConfigParser, NoOptionError, NoSectionError
+from typing import Final
+from pathlib import Path
+
+
+@dataclass
```
tzaffi, line 7: If folks find this objectionable, this can be seperated out into a seperate class file, rather than doing everything in the `__init__.py`

feature_gates/__init__.py
```diff
@@ -0,0 +1,136 @@
...
+                            f"Failed to load {feature_prop} from expected {path}. Defaulting to {gate=}."
+                        )
+                cls._set(feature_prop, gate, overwrite)
+
+        for feature_prop in cls._features:
+            # for validation purposes only:
+            cls.get(feature_prop)
+
+
+for feature in FeatureGates._features:
```
tzaffi, line 135: We implicitly create the following methods:

```
@classmethod
def FeatureGates.sourcemap_enabled(cls):

@classmethod
def FeatureGates.sourcemap_debug(cls):

@classmethod
def FeatureGates.set_sourcemap_enabled(cls):

@classmethod
def FeatureGates.set_sourcemap_debug(cls):
```

pyteal/__init__.py
```diff
@@ -1,69 +1,74 @@
+# this needs to be imported before any other pyteal module
+from feature_gates import FeatureGates
+
+FeatureGates.load()
```
tzaffi, line 4: re-sorted the imports since I was touching this file

pyteal/ast/txn.py
```diff
@@ -553,7 +553,8 @@ def config_asset_total(self) -> TxnExpr:
 
         Only set when :any:`type_enum()` is :any:`TxnType.AssetConfig`.
 
-        For more information, see https://developer.algorand.org/docs/reference/transactions/#total"""
+        For more information, see https://developer.algorand.org/docs/reference/transactions/#total
```
tzaffi, line 556: black 23 is being silly - oh well

pyteal/stack_frame.py
```diff
@@ -11,6 +12,14 @@
 from executing import Source  # type: ignore
 
 
+def _sourcmapping_is_off() -> bool:
+    return not FeatureGates.sourcemap_enabled()  # type: ignore[attr-defined]
```
tzaffi, line 16: `type: ignore` because of the implicitly defined method

feature_gates/__init__.py
```diff
@@ -0,0 +1,136 @@
...
+        if current is None or overwrite:
+            setattr(cls._gates, feature, gate)
+
+    @classmethod
+    def set(cls, feature: str, gate: bool) -> None:
+        """Set the feature gate for the given feature to the given value."""
+        cls._set(feature, gate, overwrite=True)
+
+    @classmethod
+    def load(cls, *, overwrite: bool = False, config_path: str = ""):
```
bbroder-algo, line 86: I like the feature gates.  I hate the filesystem access.

tzaffi: That means that the only way to change the default behavior of a feature gate is to explicitly call it in every indendent execution -for example- via `FeatureGate.set_sourcemap_enabled(False)`.

You wouldn't be able to just write the config file and have sourcemapping work universally in your project. 

Are you ok with that?

setup.py
```diff
@@ -16,6 +16,7 @@
     url="https://github.com/algorand/pyteal",
     packages=setuptools.find_packages(
         include=(
+            "feature_gates",
```
tzaffi, line 19: needed to export the new package

feature_gates/__init__.py
bbroder-algo: I am ok with that.  Is @algoanne ok with that

tzaffi:
closing in favor of #687

PR #687, opened 2023-03-14
Feature Gates without a Config File

tzaffi:
Addressing the pyteal.ini portion of https://github.com/algorand/pyteal/issues/677 by **removing it** as well as proposing a way to address https://github.com/algorand/pyteal/issues/601

Please see `examples/applications/sourcemap.py` for usage example.

## Regarding Feature Gates Issue [601](https://github.com/algorand/pyteal/issues/601)

`feature_gates.FeatureGates` provides a mechanism for defining global booleans that are loaded before the rest of PyTeal is. An experimental code path could directly use these globals via `FeatureGates.theFeature_isOn()` or I can also imagine a decorator such as `@FeatureGate.decorate_theFeature_isOn` which actually runs an `assert FeatureGates.theFeature_isOn()` as part of the wrapped method. To avail the feature in a particular build, one could call `FeatureGate.set_theFeature_isOn(True)` and to officially release (without removing the decorators) is as simple as changing the default defined in `FeatureGates._gates.theFeature_isOn`

>Optionally, providing an explicit opt-in mechanism to use experimental features.

That's available via -for example- `FeatureGate.set_theFeature_isOn(True)`

PR #688, opened 2023-03-15
Best Guess Sourcemapping when Partially Enabled

tzaffi:
As of #687, to enable source mapping one must `import FeatureGates from feature_gates` and then call `FeaturGates.set_sourcemap_enabled(True)`. One must do this _before_ any **PyTeal** imports as the `NatalStackFrame` embedded in `Expr`'s needs to be aware of the feature gate at construction time. Furthermore, as some expressions are statically derived at **PyTeal**'s importation, these end up lacking any stack frame information.

This PR aims to ameliorate this situation by providing stand-in `PyTealFrame`'s in the case that this has happened. In such cases, a **warning** is printed out alerting of the fact that `FeatureGate` must be imported and enabled before the PyTeal imports, but everything
else proceeds as before.

For example, I've temporarily modified [the AlgoBank example](https://github.com/algorand/pyteal/blob/3a7bfa2a1aa91b2713b5f96275b847c38b4313bd/examples/application/abi/algobank.py#L143) to illustrate what occurs in such a case. In particular, we will see the following error messages:

```sh
❯ python algobank.py
-------------------
WARNING: Error retrieving the best frame for source mapping.
This may occur because FeatureGates was imported and `FeatureGates.set_sourcemap_enabled(True)` called _AFTER_ pyteal was imported.
error: expected to have some frames but currently self._frames=[]

-------------------
WARNING: Error retrieving the best frame for source mapping.
This may occur because FeatureGates was imported and `FeatureGates.set_sourcemap_enabled(True)` called _AFTER_ pyteal was imported.
error: expected to have some frames but currently self._frames=[]

... more similar errors ...

-----------------
WARNING: Source mapping is unknown for the following:
TealLine(21: int NoOp // PyTeal: router.compile(version=6, optimize=OptimizeOptions(scratch_slots=True), with_sourcemaps=True, annotate_teal=True))

-----------------
WARNING: Source mapping is unknown for the following:
TealLine(42: int NoOp // PyTeal: router.compile(version=6, optimize=OptimizeOptions(scratch_slots=True), with_sourcemaps=True, annotate_teal=True))

-----------------
WARNING: Source mapping is unknown for the following:
TealLine(63: int NoOp // PyTeal: router.compile(version=6, optimize=OptimizeOptions(scratch_slots=True), with_sourcemaps=True, annotate_teal=True))

-----------------
WARNING: Source mapping is unknown for the following:
TealLine(70: int OptIn // PyTeal: router.compile(version=6, optimize=OptimizeOptions(scratch_slots=True), with_sourcemaps=True, annotate_teal=True))

... more similar errors ...
```

### DIFF
--- a/pyteal/compiler/sourcemap.py
+++ b/pyteal/compiler/sourcemap.py
@@ -461,6 +461,7 @@ class TealMapItem(PyTealFrame):
         teal_line: str,
         teal_component: "pt.TealComponent",
         pcs: list[int] | None = None,
+        is_sentinel: bool = False,
     ):
         super().__init__(
             frame_info=pt_frame.frame_info,
@@ -474,6 +475,7 @@ class TealMapItem(PyTealFrame):
         self.teal_line: Final[str] = teal_line
         self.teal_component: Final[pt.TealComponent] = teal_component
         self.pcs: Final[list[int] | None] = pcs if pcs else None
+        self.is_sentinel: Final[bool] = is_sentinel
 
     def pcs_repr(self, prefix: str = "") -> str:
         if not self.pcs:
@@ -482,7 +484,7 @@ class TealMapItem(PyTealFrame):
 
     def __repr__(self) -> str:
         P = " // "
-        return f"TealLine({self.teal_lineno}: {self.teal_line}{self.pcs_repr(prefix=P)} // PyTeal: {self._hybrid_w_offset()[0]}"
+        return f"TealLine({self.teal_lineno}: {self.teal_line}{self.pcs_repr(prefix=P)} // PyTeal: {self._hybrid_w_offset()[0]})"
 
     def teal_column(self) -> int:
         """Always returns 0 as the 0-index STARTING column offset"""
@@ -700,7 +702,7 @@ class _PyTealSourceMapper:
         self.teal_filename: str | None = teal_filename
         self.verbose: bool = verbose
 
-        self._best_frames: list[PyTealFrame] = []
+        self._best_frames: list[PyTealFrame | None] = []
         self._cached_r3sourcemap: R3SourceMap | None = None
 
         self._cached_tmis: list[TealMapItem] = []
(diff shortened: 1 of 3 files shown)


bbroder-algo:
this is good an earlier idea I had was to just check `sys.modules` for pyteal in feature_gates and error out if it's there, but this is more complicated and better-er

bbroder-algo (approved):

pyteal/compiler/sourcemap.py
```diff
@@ -857,6 +866,16 @@ def _validate_build(self):
...
                 )
 
+        for tmi in self._cached_tmis:
+            if not tmi.is_sentinel:
+                continue
+            print(
+                f"""-----------------
+WARNING: Source mapping is unknown for the following:
+{tmi!r}
+"""
```
ahangsu, line 876: if I understood correctly, this warning will happen also in normal compilation with sourcemap (and correctly set feature gate), right?

ahangsu (commented):
just one question

pyteal/compiler/sourcemap.py
tzaffi: I would be very surprised if this happens when feature gates are imported _prior _to pyteal and set correctly. If this occurs it would be a bug.

As a sanity check I just ran locally and got the following:
```
❯ cd examples/application
❯ python sourcemap.py
SUCCESS!!! Please check out teal/sourcemap.teal and teal/sourcemap_annotated.teal
```

The logic here is:

1. Insert a **sentinel** `TealMapItem` when no _best frame_ is found for a `TealComponent`
2. In `_validate_build()` look for such a sentinel. Only in the case that it is found print the error message.

tzaffi: Ran another sanity check, this time on [the AlgoBank example](https://github.com/algorand/pyteal/blob/3a7bfa2a1aa91b2713b5f96275b847c38b4313bd/examples/application/abi/algobank.py#L143) -but brough the `... import FeatureGate ...` stuff to the top.

ahangsu: oh my bad, I was confused! The logic got me into the wrong point is that, I wrongly assume the case that `self._best_frame` would contain `None` before this PR, but *only under wrongly set FeatureGate* would cause such issue, and thus we introduce the sentinel_frame to trigger the warning!

Now I understood... resolving

ahangsu (approved):

Issue #681, opened 2023-02-23
Source Mapper Improvements: Small Fixes


tzaffi:
## Problem

There were a few issues that were not resolved at the time of mergine #650 . 

## Solution

- [ ] Small improvements
  - [ ] Refactor `NatalStackFrame` to provide a single frame instead of a list
  - [ ] Coverage triage
  - [ ] Prune unused methods in the `StackFrame - PyTealFrame - TealMapItem` hierarchy
  - [ ] Pull in updated version of `mjpieters`'s **SourceMap**  to get faster R3SourceMap (cf: https://github.com/tzaffi/pyteal/blob/f2a945a8eeb396e9c544b62cd3d1a6383bfc9bff/pyteal/compiler/sourcemap.py#L75)
  - [ ] Final status checks (ensure all changes are actually necessary)
    - [ ] which of the usages of `._sframes_container` are actually necessary. To see all usages: [f96ef86](https://github.com/algorand/pyteal/pull/650/commits/f96ef8657124466884974501137764a3f7c0a266)
  - [ ] Less clunky workflow and artifact access in `_PyTealSourceMapper` (cf: https://github.com/tzaffi/pyteal/blob/f2a945a8eeb396e9c544b62cd3d1a6383bfc9bff/pyteal/compiler/sourcemap.py#L730)
  - [ ] Clarify the legend for `_PyTealMapper.annotate()` (cf: https://github.com/tzaffi/pyteal/blob/f2a945a8eeb396e9c544b62cd3d1a6383bfc9bff/pyteal/compiler/sourcemap.py#L844)

## Dependencies

None

## Urgency

Medium - while the urgency of any improvement individually is Low, addressing at least a few of these issues in the short term will improve the fetature's usefuleness.

PR #689, opened 2023-03-16
Better Router Mapping

tzaffi:
Along with several tasks hilighted below, the PR improves the hit rate on router generated code by associating most of the previously unmatched lines to specific methods or bare calls. For example, you can see the improvement by looking at some of the diffs in this PR:

image

### Other tasks from #681 which were completed
- [x] Prune unused methods in the `StackFrame - PyTealFrame - TealMapItem` hierarchy (#689)
- [x] Final status checks (ensure all changes are actually necessary) (#689)
  - [x] which of the usages of `._sframes_container` are actually necessary. To see all usages: [f96ef86](https://github.com/algorand/pyteal/pull/650/commits/f96ef8657124466884974501137764a3f7c0a266)
- [x] audit the following test files and either remove, unskip, or provide clear commentary as to why the skipped test needs to stay in (#689 unified using `STABLE_SLOT_GENERATION: bool` constants currently `False`):
  - [x] tests/integration/sourcemap_monkey_integ_test.py
  - [x] tests/unit/sourcemap_constructs311_test.py
  - [x] tests/unit/sourcemap_constructs_allpy_test.py
  - [x] tests/unit/sourcemap_monkey_unit_test.py
  - [x] tests/unit/sourcemap_test.py
- [x] At the end of #676 pay close attention to `_PyTealSourceMap._infer()` and see which lines are still being hit. Probably, there will be some situations that are no longer needed. (#689 tightened this up)

## Related PR's and Issues
* Continuing from #676 
* Implementing around 50% of the taks in #681


pyteal/ast/router.py
```diff
@@ -256,13 +256,12 @@ def approval_construction(self) -> Optional[Expr]:
...
-            conditions_n_branches.append(
-                CondNode(
-                    Txn.on_completion() == oc,
-                    cond_body,
-                )
-            )
-        return Cond(*[[n.condition, n.branch] for n in conditions_n_branches])
+            cn = CondNode(Txn.on_completion() == oc, cond_body)
+            cn.reframe_asts(oca.stack_frames)
+            conditions_n_branches.append(cn)
```
bbroder-algo, line 261: way better

pyteal/ast/router.py
```diff
@@ -288,6 +287,9 @@ class CondNode:
     condition: Expr
     branch: Expr
 
+    def reframe_asts(self, stack_frames: NatalStackFrame) -> None:
```
bbroder-algo, line 290: maybe a line about what a reframe is

pyteal/ast/router.py
```diff
@@ -531,6 +540,13 @@ def wrap_handler(
                 - for abi-method it returns the txn args correctly decomposed into ABI variables,
                   passed in ABIReturnSubroutine and logged, then approve.
         """
+
+        def scavenge(frames_holder) -> None:
```
bbroder-algo, line 544: why "scavenge"?

pyteal/stack_frame.py
```diff
@@ -133,20 +133,16 @@ def _is_pyteal(self) -> bool:
...
-        return self._frame_info_is_pyteal_import(self.frame_info)
-
     @classmethod
     def _frame_info_is_pyteal_import(cls, f: FrameInfo) -> bool:
         cc = f.code_context
         if not cc:
             return False
 
-        code = "".join(cc).split()
+        code = re.split(r"\s|\.", " ".join(cc))
```
bbroder-algo, line 142: this splits up an array of strings by spaces and periods

pyteal/stack_frame.py
```diff
@@ -229,7 +233,7 @@ def _debugging() -> bool:
     def __init__(
         self,
     ):
-        self._compiler_gen: bool = False
+        self._pyteal_gen: bool = False
```
bbroder-algo, line 236: better

pyteal/stack_frame.py
tzaffi: will add a comment

pyteal/stack_frame.py
```diff
@@ -318,46 +331,96 @@ def _best_frame_as_pyteal_frame(self) -> "PyTealFrame | None":
...
+                case Int(), MethodSignature(), ScratchStackStore(), TxnaExpr():
                     pass
+                case _:
+                    supported_type = False
+
+            should_stop = True
+            if supported_type or force_root_apply:
+                should_stop = func(expr)
+            if exit_when_stop_signal and should_stop:
+                return
```
ahangsu, line 415: nit: should we return here or continue in the loop here?

pyteal/stack_frame.py
```diff
@@ -318,46 +331,96 @@ def _best_frame_as_pyteal_frame(self) -> "PyTealFrame | None":
...
     @classmethod
-    def _walk_asts(cls, func: Callable[["Expr"], None], *exprs: "Expr") -> None:  # type: ignore
+    def _walk_asts(
+        cls,
+        func: Callable[["Expr"], bool],  # type: ignore[name-defined]
+        *exprs: "Expr",  # type: ignore[name-defined]
+        force_root_apply: bool = True,
+        exit_on_user_defined: bool = False,
+        exit_when_stop_signal: bool = False,
+    ) -> None:
```
ahangsu, line 347: nit: I guess I understood the `func` is a method that traverse AST while giving out some auxiliary information on if keep traversing.

Maybe a sentence or two on commenting the purpose of the following configs would be helpful, in case my memory on the code fades away:
- `func`
- `force_root_apply`
- `exit_on_user_defined`
- `exit_when_stop_signal`

pyteal/stack_frame.py
```diff
@@ -318,46 +331,96 @@ def _best_frame_as_pyteal_frame(self) -> "PyTealFrame | None":
...
                 case SubroutineDeclaration():
-                    cls._walk_asts(func, e.body)
-                case _:
-                    # TODO: implement more cases, but no need to error as this isn't used for functionality's sake.
+                    walker_args = [expr.body]
+                case ScratchStore():
+                    walker_args = [expr.value]
+                    if expr.index_expression:
+                        walker_args.append(expr.index_expression)
+                case SubroutineCall():
```
ahangsu, line 397: qq: can you handle re-entrance of subroutine call caused by recursion?

pyteal/stack_frame.py
```diff
@@ -422,7 +489,6 @@ class PT_GENERATED:
     BRANCH_LABEL = "PyTeal generated branching label"
     TYPE_ENUM_TXN = "PyTeal generated transaction Type Enum"
     TYPE_ENUM_ONCOMPLETE = "PyTeal generated OnComplete Type Enum"
-    FLAGGED_BY_DEV = "Developer has flagged expression as compiler generated"
```
ahangsu, line 425: qq: this `PT_GENERATED` reason deletion, and following deletion in `compiler_generated_reason`, are caused by removal of `_smap_friendly_approve` in router, right?

For it is the only place that mark `_compiler_gen` as `True`, which is the only cause of this reason?

pyteal/ast/subroutine.py
ahangsu, line 1: Let me start from somewhere easier... So I assume the changes in `subroutine.py` is for marking all variable preparation stage of subroutine under the natal stack frame of the subroutine declaration, right?

ahangsu (commented):
Walking across the weeds... I think I grasp a feeling of how this PR goes... leaving some comments for now

pyteal/ast/router.py
tzaffi: f94fa7a8f5ee7f5356eafb9f462c192c3a605605

pyteal/ast/router.py
tzaffi: f94fa7a8f5ee7f5356eafb9f462c192c3a605605

pyteal/stack_frame.py
tzaffi: f94fa7a8f5ee7f5356eafb9f462c192c3a605605

pyteal/stack_frame.py
tzaffi: Yeah, I no longer was actually using the code, so deleted it.

pyteal/stack_frame.py
tzaffi: We want to exit the traversal under this condition. In the "should_stop" case (which is `True` for the main case of `_reframe_asts()` ) we are basically using `func == set_frames`'s output to determine whether to continue or not. An important exit condition is when `e.stack_frames == stack_frames` which alerts us that we've already seen this `Expr` as we've previously reset it's `stack_frames` and therefore avoid an infinite loop.

I'll add this explanation as a doc-string

pyteal/stack_frame.py
tzaffi: As explained in a discussion above, we know we can avoid infinite loops as we exit when the `stack_frames` param is observed to be identical to `self.stack_frames`. There is also 2 unit test that ensure this:
* https://github.com/algorand/pyteal/blob/6a49be7b4c74abe6fa08493bdb8e5585dabf0bca/tests/unit/sourcemap_constructs311_test.py#L1901
* https://github.com/algorand/pyteal/blob/6a49be7b4c74abe6fa08493bdb8e5585dabf0bca/tests/unit/sourcemap_constructs311_test.py#L1955

pyteal/stack_frame.py
ahangsu: I see, that makes sense to me, and seems rather simpler than what I thought. I guess this captures any recursion case that will eventually reach back to the function call itself.

pyteal/stack_frame.py
ahangsu: This thread should be resolved by the discussion thread below.

pyteal/stack_frame.py
tzaffi: Thinking more about it, I'm no longer 100% sure. I see I have 2 seperate logics. I will investigate the effects of just continuing instead of returning

pyteal/ast/subroutine.py
tzaffi: Correct

pyteal/ast/abi/method_return.py
```diff
@@ -30,7 +29,7 @@ def __teal__(self, options: "CompileOptions") -> Tuple[TealBlock, TealSimpleBloc
         start, end = Log(Concat(Bytes(RETURN_HASH_PREFIX), self.arg.encode())).__teal__(
             options
         )
-        NatalStackFrame.reframe_ops_in_blocks(self._sframes_container or self, start)
+        NatalStackFrame.reframe_ops_in_blocks(self, start)
```
ahangsu, line 32: qq: could you explain this line, why `_sframes_container` is removed in `MethodReturn`?

I suppose the side effect is that, the `root_expr` of `start` (and all following blocks of `start`) is tied to `self`, which is the instance of `MethodReturn`.

pyteal/ast/router.py
```diff
@@ -583,13 +617,19 @@ def wrap_handler(
...
-            return ASTBuilder.__de_abify_subroutine_vanilla(handler)
-        else:
-            return ASTBuilder.__de_abify_subroutine_frame_pointers(handler)
+
+        ret_expr, subdef = (
+            ASTBuilder.__de_abify_subroutine_frame_pointers(handler)
+            if use_frame_pt
+            else ASTBuilder.__de_abify_subroutine_vanilla(handler)
+        )
+        scavenge(subdef)
```
ahangsu, line 626: I see, so most of the changes in router is for `scavange` to build sourcemap

pyteal/ast/abi/method_return.py
tzaffi: At an earlier stage of this work, I used a more "cautious" approach that instead of overwriting "useless" `stack_frames` belonging to an `Expr` with a useful instance -as is happening now- I tried to pass the information around in associated `TealBlocks` and also keep the original `expr` and `stack_frames`.   I found this indirect approach to be harder to manage then simply overwriting. I still use this indirect approach in conditional expressions, and `SubRoutine`s but I found that it was basically unworkable for the `Router`.

pyteal/stack_frame.py
tzaffi: Ok. Thanks for your original suggestion. I've unified and trimmed all the exit logic thanks to the nudge. With the logic simplified, it was much easier to reason about, and now it seems clear that it's preferable to `continue` rather than `return`. That's because there is no obvious ordering for the AST that implies we should eagerly exit that traversal without visiting subsequent siblings. I think I was being overly conservative originally, trying to only explore the AST when absolutely necessary, and preferring to get a "miss" versus an over-aggressive and wrong mapping. However, this was before I implemented the skip strategies for not re-writing frames and not overwriting user-defined frames.

cf: 622bb848ebffa6b8e81dbecde58338e90ff29793

pyteal/stack_frame.py
tzaffi: 622bb848ebffa6b8e81dbecde58338e90ff29793

pyteal/ast/router.py
```diff
@@ -525,12 +540,25 @@ def wrap_handler(
             is_method_call: a boolean value that specify if the handler is an ABI method.
             handler: an `ABIReturnSubroutine`, or `SubroutineFnWrapper` (for `Subroutine` case), or an `Expr`.
             use_frame_pt: a boolean value that specify if router is compiled to frame pointer based code.
+            handler_stack_frames_container: an optional list that is used to return import NatalStackFrame's
+                for source mapping purposes.
```
tzaffi, line 544:
```suggestion
            handler_stack_frames_container: an optional list that is filled with NatalStackFrame's
                used in source mapping.
```

pyteal/stack_frame.py
ahangsu: looks good, thanks for the move!

ahangsu (approved):

pyteal/stack_frame.py
```diff
@@ -133,20 +133,30 @@ def _is_pyteal(self) -> bool:
...
-    def _is_pyteal_import(self) -> bool:
-        return self._frame_info_is_pyteal_import(self.frame_info)
-
     @classmethod
     def _frame_info_is_pyteal_import(cls, f: FrameInfo) -> bool:
+        """
+        This method is used to determine if a FrameInfo is associated to a pyteal import.
+        It does so by splitting its joined code context on whitespace and periods.
+        For example, the following joined code context
+            from pyteal.ast.abi import Uint64 as Positive
```
tzaffi, line 142:
```suggestion
            'from pyteal.ast.abi import Uint64 as Positive'
```

pyteal/stack_frame.py
```diff
@@ -133,20 +133,30 @@ def _is_pyteal(self) -> bool:
...
+        For example, the following joined code context
+            'from pyteal.ast.abi import Uint64 as Positive'
+        becomes
+            ['from', 'pyteal', 'ast', 'abi', 'import', 'Uint64', 'as', 'Positive']
+        and we can verify that this is a pyteal import by checking that
+        'pyteal' and 'import' are both in the list.
+        This works for other imports such as
+            import pyteal.ast.abi as my_abi
+            import pyteal as pt
+            from pyteal import *
```
tzaffi, line 150:
```suggestion
            'import pyteal.ast.abi as my_abi'
            'import pyteal as pt'
            'from pyteal import *'
```

.flake8
```diff
@@ -29,8 +29,7 @@ per-file-ignores =
     pyteal/ir/tealconditionalblock.py:              F821
     pyteal/ir/tealsimpleblock.py:                   F821
     pyteal/stack_frame.py:                          F821
-    tests/teal/rps_helpers/program.py:              F401, F403, F405
-    tests/teal/rps.py:                              F403, F405, I252
+    tests/teal/rps.py:                              F401, F403, F403, F405, I252
```
tzaffi, line 32:
```suggestion
    tests/teal/rps.py:                              F401, F403, F405, I252
```

PR #690, opened 2023-03-16
Validation Already Happens in Compiler - Remove from Router

tzaffi:
## Reasoning

The following call-trace shows validation is already happenend in `compiler.py` by the time that the code running validation in the router runs again. Given the instability of compilation (#199 ), and to avoid unnecessary CPU cycles - we should remove the second validation.

## Call Trace that Proves Validation
* [Router._build_impl()](https://github.com/algorand/pyteal/pull/690/files#diff-3844cf2bef1f9fc5ce0b421c65141cf76d172d885b5792ac0cd79dc360618769L1236)
  * [Compilation.get_compilation()](https://github.com/algorand/pyteal/blob/master/pyteal/compiler/compiler.py#L505)


bbroder-algo (approved):
i feel validated.

PR #703, opened 2023-04-18
Relax timeout for `test_many_ifs`, and redirect algod image

ahangsu:
Timeout for `test_many_ifs` seems blocking #702


jasonpaulos (approved):
Seems like a good improvement

PR #715, opened 2024-01-02
Release v0.26.0

jasonpaulos:
Update changelog and `setup.py` to prepare for version 0.26.0 release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## Added
 
+## Fixed
+
+## Changed
+
+# v0.26.0
+
+## Added
+
 * Support for AVM v10 programs. ([#714](https://github.com/algorand/pyteal/pull/714))
   * New box operations:
     * `App.box_resize`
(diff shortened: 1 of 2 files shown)


jannotti (approved):
Weren't there two extra fields that got added because we forgot them from v9?

jasonpaulos:
@jannotti let me know if I have this right: v9 didn't introduce any new fields, but v10 introduced `global AssetCreateMinBalance` and `global AssetOptInMinBalance`, which were developed some time ago (and of course the more recent `global GenesisHash`

jannotti:
Sorry, it was v6.  These were in your v10 PR, so I figured they should be mentioned here.

:any:`Global.opcode_budget()`               :code:`TealType.uint64` 6                    The remaining cost that can be spent by opcodes in this program
:any:`Global.caller_app_id()`               :code:`TealType.uint64` 6                    The ID of the application that called the current application, or zero. Application mode only
:any:`Global.caller_app_address()`          :code:`TealType.bytes`  6                    32 byte address of the application that called the current application, or the zero address. Application mode only.

jasonpaulos:
Oh those were just missing from the docs. They've still been functional

PR #722, opened 2025-01-27
Added new fields for global, acct_params_get, and block

jannotti:
Added new fields for global, acct_params_get, and block along with new opcodes, `voter_params_get`, `online_stake`, and `mimc`.


pyteal/ast/block.py
```diff
@@ -14,8 +14,16 @@
...
-    block_seed =      (0, "BlkSeed",      TealType.bytes,  7)  # noqa: E222
+    block_seed = (0, "BlkSeed",      TealType.bytes,  7)  # noqa: E222
     block_timestamp = (1, "BlkTimestamp", TealType.uint64, 7)  # noqa: E222
+    block_proposer = (2, "BlkProposer", TealType.bytes, 11)  # noqa: E222
+    block_fees_collected = (3, "BlkFeesCollected", TealType.uint64, 11)  # noqa: E222
+    block_bonus = (4, "BlkBonus", TealType.uint64, 11)  # noqa: E222
+    block_branch = (5, "BlkBranch", TealType.bytes, 11)  # noqa: E222
+    block_fee_sink = (6, "BlkFeeSink", TealType.bytes, 11)  # noqa: E222
+    block_protocol = (7, "BlkProtocol", TealType.bytes, 11)  # noqa: E222
+    block_txn_counter = (8, "BlkTxnCounter", TealType.uint64, 11)  # noqa: E222
```
gmalouf, line 25: I guess figured support most of the simple type fields?

pyteal/ast/block.py
```diff
@@ -14,8 +14,16 @@
 class BlockField(Enum):
```
gmalouf, line 14: I guess do not include round here since available as global already right?

pyteal/ast/block_test.py
```diff
@@ -58,3 +60,37 @@ def test_block_timestamp():
...
+        )
+
+    actual, _ = expr.__teal__(avm11Options)
+    actual.addIncoming()
+    actual = pt.TealBlock.NormalizeBlocks(actual)
+
+    assert actual == expected
+
+    with pytest.raises(pt.TealInputError):
+        expr.__teal__(avm10Options)
```
gmalouf, line 96: Why raise the options from previous version (v10) here?

pyteal/ast/block.py
jannotti: That's about right, though again, that's a choice we made at AVM level. There's no `BlkRound` constant for looking at the round in a block.  It's exactly because there's a `global`, since that's the current round, while this is accessing an old block. But the code has to look that round up by specifying the round, so there's little point in then accessing it.

pyteal/ast/block_test.py
jannotti: It's saying that if you call `expr._teal_` and specify that you want it compiled for v10, then an exception will be raised (since the field didn't exist in v10).

algorandskiy (approved):

gmalouf (approved):

